PR #4, opened 2015-10-09
added option to use customviews instead of provided ones.

Bartmax:


src/OpenIddict/OpenIddictOptions.cs
```diff
@@ -79,5 +79,11 @@ public class OpenIddictOptions {
         /// Setting this option to false in production is strongly encouraged to mitigate man-in-the-middle attacks.
         /// </summary>
         public bool AllowInsecureHttp { get; set; }
+
+        /// <summary>
+        /// Set to True to allow you to use your own views/styles/scripts in your server.
+        /// When using custom views you MUST provide cshtml views for: Authorize (consent), Logout, and SignIn
+        /// </summary>
+        public bool UseCustomViews { get; set; }
```
kevinchalet, line 87: Not sure how I feel about a single option that control both the assets and the views.

src/OpenIddict/OpenIddictOptions.cs
```diff
@@ -79,5 +79,11 @@ public class OpenIddictOptions {
         /// Setting this option to false in production is strongly encouraged to mitigate man-in-the-middle attacks.
         /// </summary>
         public bool AllowInsecureHttp { get; set; }
+
+        /// <summary>
+        /// Set to True to allow you to use your own views/styles/scripts in your server.
```
kevinchalet: Please use `<c>true</c>` instead :+1:

src/OpenIddict/OpenIddictOptions.cs
```diff
@@ -79,5 +79,11 @@ public class OpenIddictOptions {
...
+        /// When using custom views you MUST provide cshtml views for: Authorize (consent), Logout, and SignIn
```
kevinchalet: Don't forget the ending `.` :trollface:

kevinchalet:
Merged, thanks! https://github.com/openiddict/core/commit/fb103fad11093b20b8211b53e1dc23c8a12962ba :tada:

PR #8, opened 2015-10-16
Decouple the EntityFramework adapter from the core project

kevinchalet:
(please note that it's still a WIP :smile:)

/cc @Bartmax @ilmax


samples/Mvc.Server/Startup.cs
```diff
@@ -34,7 +34,10 @@ public class Startup {
 
             services.AddIdentity<ApplicationUser, IdentityRole>()
                 .AddEntityFrameworkStores<ApplicationDbContext>()
-                .AddDefaultTokenProviders();
+                .AddDefaultTokenProviders()
+
+                .AddOpenIddict<ApplicationUser, Application>()
```
Bartmax: why not `AddOpenIddict<Application>` ApplicationUser is already part of the AddIdentity part, can we use that ?

src/OpenIddict.Core/IOpenIddictStore.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Identity;
+
+namespace OpenIddict {
+    public interface IOpenIddictStore<TUser, TApplication> : IUserStore<TUser> where TUser : class where TApplication : class {
```
Bartmax, line 6: Do we need to inherit from IUserStore ? we are not using TUser for anything on the store.

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -13,31 +13,13 @@
...
-            Context = context;
-        }
-
-        /// <summary>
-        /// Gets the OpenIddict context.
-        /// </summary>
-        protected TContext Context { get; }
-
+    // Note: this controller is generic and doesn't need to be marked as internal to prevent MVC from discovering it.
+    public class OpenIddictController<TUser, TApplication> : Controller where TUser : class where TApplication : class {
```
Bartmax, line 22: instead of `Controller<T>` wouldn't be better if we inject in the constructor the `OpenIddictManager` and `UserManager` ?

src/OpenIddict.Core/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,205 @@
...
+                        routes.MapRoute("{C7DB102A}", instance.LogoutEndpointPath.Value.Substring(1), new {
+                            controller = typeof(OpenIddictController<,>).Name,
+                            action = nameof(OpenIddictController<dynamic, dynamic>.Logout)
+                        });
+                    }
+                });
+            }, services => {
+                services.AddAuthentication();
+                services.AddCaching();
+                services.AddCors();
```
Bartmax: I think AddMvc() already adds Cors and Cache, there's no need to Add them again (?)

src/OpenIddict.Core/OpenIddictManager.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Identity;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.OptionsModel;
+
+namespace OpenIddict {
+    public class OpenIddictManager<TUser, TApplication> : UserManager<TUser> where TUser : class where TApplication : class {
```
Bartmax, line 11: What do you think about leaving OpenIddictManager as one manager and UserManager as another and avoiding inheritance here ?

src/OpenIddict.Core/OpenIddictProvider.cs
```diff
@@ -12,19 +12,11 @@
...
 using Microsoft.Extensions.Internal;
-using OpenIddict.Models;
 
 namespace OpenIddict {
-    public class OpenIddictProvider<TContext, TUser, TRole, TKey> : OpenIdConnectServerProvider
-        where TContext : OpenIddictContext<TUser, TRole, TKey>
-        where TUser : IdentityUser<TKey>
-        where TRole : IdentityRole<TKey>
-        where TKey : IEquatable<TKey> {
+    public class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServerProvider where TUser : class where TApplication : class {
```
Bartmax, line 19: same suggestion as in the controller, constructor inject dependent managers instead of Provider<T>.

src/OpenIddict.Core/OpenIddictProvider.cs
```diff
@@ -136,8 +126,8 @@ public class OpenIddictProvider<TContext, TUser, TRole, TKey> : OpenIdConnectSer
             }
 
             // Confidential applications MUST authenticate.
-            else if (application.Type == ApplicationType.Confidential &&
-                !string.Equals(context.ClientSecret, application.Secret, StringComparison.Ordinal)) {
+            else if (string.Equals(type, "confidential", StringComparison.OrdinalIgnoreCase) &&
```
Bartmax: make magic string "confidential" as a const or enum

src/OpenIddict.Core/Views/Shared/Authorize.cshtml
```diff
@@ -1,14 +1,13 @@
 ﻿@using AspNet.Security.OpenIdConnect.Extensions
 @using Microsoft.IdentityModel.Protocols.OpenIdConnect
-@using OpenIddict.Models
 
-@model Tuple<OpenIdConnectMessage, Application>
+@model Tuple<OpenIdConnectMessage, string>
```
Bartmax, line 4: since we are embedding everything I think we should create a specific ViewModel instead of passing in a Tuple.

src/OpenIddict.EF/OpenIddictContext.cs
```diff
@@ -10,12 +10,13 @@
...
         where TUser : IdentityUser<TKey>
+        where TApplication : Application
         where TRole : IdentityRole<TKey>
         where TKey : IEquatable<TKey> {
-        public DbSet<Application> Applications { get; set; }
+        public DbSet<TApplication> Applications { get; set; }
     }
 
-    public class OpenIddictContext<TUser> : OpenIddictContext<TUser, IdentityRole, string> where TUser : IdentityUser { }
+    public class OpenIddictContext<TUser> : OpenIddictContext<TUser, Application, IdentityRole, string> where TUser : IdentityUser { }
```
Bartmax, line 21: Non-Generic constructor with IdentityUser as TUser ?

src/OpenIddict.EF/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.AspNet.Identity;
+using Microsoft.AspNet.Identity.EntityFramework;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Internal;
+using OpenIddict;
+using OpenIddict.Models;
+
+namespace Microsoft.AspNet.Builder {
+    public static class OpenIddictExtensions {
+        public static OpenIddictBuilder AddEntityFrameworkStores<TUser, TApplication, TRole, TKey>([NotNull] this OpenIddictBuilder builder)
```
Bartmax: We only use 1 store, rename `AddEntityFrameworkStores` to `AddEntityFrameworkStore`.

src/OpenIddict.EF/OpenIddictStore.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Identity.EntityFramework;
+using Microsoft.Data.Entity;
+using OpenIddict.Models;
+
+namespace OpenIddict {
+    public class OpenIddictStore<TUser, TApplication, TRole, TKey> : UserStore<TUser, TRole, OpenIddictContext<TUser, TApplication, TRole, TKey>, TKey>, IOpenIddictStore<TUser, TApplication>
```
Bartmax, line 9: I don't like that we inherit from UserStore here. I would prefer to have 2 separate stores, the user and the OpenIddict one. Isn't the UserStore a dependecy?

samples/Mvc.Server/Startup.cs
kevinchalet: Sure. Updated :+1:

src/OpenIddict.Core/IOpenIddictStore.cs
kevinchalet, line 6: `IOpenIddictStore` inherits from `IUserStore` for conveniency :smile: 

In the current bits, we're not using `TUser` yet, but we'll ultimately need it to retrieve the authorizations corresponding to a given user.

src/OpenIddict.Core/OpenIddictController.cs
kevinchalet, line 22: Sure we can use constructor injection here, but we still need the two generic parameters, since we need them for `OpenIddict<TApp, TUser>` and `UserManager<TUser>`.

src/OpenIddict.Core/OpenIddictExtensions.cs
kevinchalet: Yup for `AddCors()`. I'll triple check for `AddCaching()` and fix it :+1:

src/OpenIddict.Core/OpenIddictManager.cs
kevinchalet, line 11: IMHO, inheritance is quite handy, as you don't have to import 2 (or more) managers in your code to play with extremely-related stuff.

src/OpenIddict.Core/OpenIddictProvider.cs
kevinchalet, line 19: We can't. The provider is necessarily a singleton and the managers are scoped.

src/OpenIddict.Core/OpenIddictProvider.cs
kevinchalet: Sure.

src/OpenIddict.Core/Views/Shared/Authorize.cshtml
kevinchalet, line 4: I'm definitely not a huge fan of view models, but I'll consider it.

src/OpenIddict.EF/OpenIddictContext.cs
kevinchalet, line 21: FYI, it's rather unusual in ASP.NET Identity, where the User is always a custom type.

Bartmax, line 21: mmmm ok, I have some applications that use IdentityUser as is, but I see your point (also passing IdentityUser is trivial in case of not using a custom type. let's keep it this way and if we see someone ask for it we can always add it later.

src/OpenIddict.Core/OpenIddictProvider.cs
Bartmax, line 19: :+1:

src/OpenIddict.EF/OpenIddictContext.cs
kevinchalet, line 21: I'm fine adding a new constructor, it's not too painful :+1:

src/OpenIddict.Core/Views/Shared/Authorize.cshtml
Bartmax, line 4: Maybe a `OpenIddictConnectResult` is better suited than a ViewModel, like IdentityResult and SignInResult but I don't know where `request` is coming from at controller level so I leave this for you to decide.

src/OpenIddict.Core/OpenIddictExtensions.cs
kevinchalet: `AddCaching()` is not automatically called by MVC. I removed CORS and Options.

src/OpenIddict.Core/Views/Shared/Authorize.cshtml
kevinchalet, line 4: I'll consider that in a future PR :+1:

src/OpenIddict.EF/OpenIddictContext.cs
kevinchalet, line 21: Done.

src/OpenIddict.Core/OpenIddictProvider.cs
kevinchalet: Done.

src/OpenIddict.Core/OpenIddictController.cs
kevinchalet, line 22: Fixed to use constructor injection.

src/OpenIddict.Core/OpenIddictProvider.cs
```diff
@@ -126,8 +113,10 @@ public class OpenIddictProvider<TContext, TUser, TRole, TKey> : OpenIdConnectSer
                 return;
             }
 
-            // Reject tokens requests containing a client_secret if the client application is not confidential.
-            if (application.Type == ApplicationType.Public && !string.IsNullOrEmpty(context.ClientSecret)) {
+            // Reject tokens requests containing a client_secret
+            // if the client application is not confidential.
+            var type = await manager.GetApplicationTypeAsync(application);
```
ilmax, line 118: nit: applicationType instead of type

Bartmax:
:shipit:

kevinchalet:
Merged. Thanks for the review guys :+1:

PR #19, opened 2015-10-22
Implement userinfo support

kevinchalet:
/cc @ilmax @Bartmax

### DIFF
--- a/src/OpenIddict.Core/OpenIddictController.cs
+++ b/src/OpenIddict.Core/OpenIddictController.cs
@@ -115,14 +115,28 @@ namespace OpenIddict {
             var identity = new ClaimsIdentity(Options.AuthenticationScheme);
             identity.AddClaim(ClaimTypes.NameIdentifier, await Manager.GetUserIdAsync(user));
 
-            // Only add the name claim if the "profile" scope was present in the token request.
-            if (request.GetScopes().Contains("profile", StringComparer.OrdinalIgnoreCase)) {
-                identity.AddClaim(ClaimTypes.Name, await Manager.GetUserNameAsync(user), destination: "id_token token");
+            // Resolve the username and the email address associated with the user.
+            var username = await Manager.GetUserNameAsync(user);
+            var email = await Manager.GetEmailAsync(user);
+
+            // Only add the name claim if the "profile" scope was present in the authorization request.
+            if (request.ContainsScope(OpenIdConnectConstants.Scopes.Profile)) {
+                // Return an error if the username corresponds to the registered
+                // email address and if the "email" scope has not been requested.
+                if (!request.ContainsScope(OpenIdConnectConstants.Scopes.Email) &&
+                     string.Equals(username, email, StringComparison.OrdinalIgnoreCase)) {
+                    return View("Error", new OpenIdConnectMessage {
+                        Error = OpenIdConnectConstants.Errors.InvalidRequest,
+                        ErrorDescription = "The 'email' scope is required."
+                    });
+                }
+
+                identity.AddClaim(ClaimTypes.Name, username, destination: "id_token token");
             }
 
             // Only add the email address if the "email" scope was present in the token request.
-            if (request.GetScopes().Contains("email", StringComparer.OrdinalIgnoreCase)) {
-                identity.AddClaim(ClaimTypes.Email, await Manager.GetEmailAsync(user), destination: "id_token token");
+            if (request.ContainsScope(OpenIdConnectConstants.Scopes.Email)) {
+                identity.AddClaim(ClaimTypes.Email, email, destination: "id_token token");
             }
 
             // Note: AspNet.Security.OpenIdConnect.Server automatically ensures an application
--- a/src/OpenIddict.Core/OpenIddictProvider.cs
+++ b/src/OpenIddict.Core/OpenIddictProvider.cs
@@ -216,13 +216,16 @@ namespace OpenIddict {
             var identity = new ClaimsIdentity(context.Options.AuthenticationScheme);
             identity.AddClaim(ClaimTypes.NameIdentifier, await manager.GetUserIdAsync(user));
 
-            // Only add the name claim if the "profile" scope was present in the token request.
-            if (context.Request.GetScopes().Contains("profile", StringComparer.OrdinalIgnoreCase)) {
+            // Only add the name claim if the "profile" scope was present in the authorization request.
+            // Note: filtering the username is not needed at this stage as OpenIddictController.Accept
+            // and OpenIddictProvider.GrantResourceOwnerCredentials are expected to reject requests that
+            // don't include the "email" scope if the username corresponds to the registed email address.
+            if (context.Request.ContainsScope(OpenIdConnectConstants.Scopes.Profile)) {
                 identity.AddClaim(ClaimTypes.Name, await manager.GetUserNameAsync(user), destination: "id_token token");
             }
 
-            // Only add the email address if the "email" scope was present in the token request.
-            if (context.Request.GetScopes().Contains("email", StringComparer.OrdinalIgnoreCase)) {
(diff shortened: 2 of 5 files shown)


src/OpenIddict.Core/OpenIddictProvider.cs
```diff
@@ -234,6 +240,55 @@ public class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServerProvid
...
+                context.Claims[JwtRegisteredClaimNames.FamilyName] = await manager.FindClaimAsync(user, ClaimTypes.Surname);
+                context.Claims[JwtRegisteredClaimNames.GivenName] = await manager.FindClaimAsync(user, ClaimTypes.GivenName);
+                context.Claims[JwtRegisteredClaimNames.Birthdate] = await manager.FindClaimAsync(user, ClaimTypes.DateOfBirth);
+            }
+
+            if (principal.HasClaim(OpenIdConnectConstants.Claims.Scope, OpenIdConnectConstants.Scopes.Email)) {
+                context.Claims[JwtRegisteredClaimNames.Email] = email;
+                context.Claims[OpenIddictConstants.Claims.EmailVerified] = await manager.IsEmailConfirmedAsync(user) ? "true" : "false";
+            };
+
```
kevinchalet, line 274: I'll probably add native support for `address` and `phone` before merging this PR: http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -115,14 +115,20 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
...
+            // and if the username doesn't correspond to the registered email address.
+            if (request.ContainsScope(OpenIdConnectConstants.Scopes.Profile) &&
+                !string.Equals(username, email, StringComparison.OrdinalIgnoreCase)) {
+                identity.AddClaim(ClaimTypes.Name, username, destination: "id_token token");
             }
 
             // Only add the email address if the "email" scope was present in the token request.
-            if (request.GetScopes().Contains("email", StringComparer.OrdinalIgnoreCase)) {
-                identity.AddClaim(ClaimTypes.Email, await Manager.GetEmailAsync(user), destination: "id_token token");
+            if (request.ContainsScope(OpenIdConnectConstants.Scopes.Email)) {
```
ilmax, line 138: :+1:

kevinchalet, line 138: FYI, this extension is deliberately case insensitive to avoid scope values that only differ by the name (despite the fact the OAuth2 specs say `scope` is supposed to be case sensitive :smile: )

src/OpenIddict.Core/OpenIddictProvider.cs
```diff
@@ -234,6 +240,55 @@ public class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServerProvid
             context.HandleResponse();
         }
 
+        public override async Task ProfileEndpoint([NotNull] ProfileEndpointContext context) {
+            var manager = context.HttpContext.RequestServices.GetRequiredService<OpenIddictManager<TUser, TApplication>>();
+
+            var principal = context.AuthenticationTicket?.Principal;
```
ilmax, line 243: nit: what's the benefit of `?.` if we assert this should not be null and call `GetUserId()` extension method on the instance?

kevinchalet, line 243: Actually, ASOS is supposed to guard against null tickets. So I guess I'm probably paranoid here :smile:

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -115,14 +115,20 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
...
-                identity.AddClaim(ClaimTypes.Name, await Manager.GetUserNameAsync(user), destination: "id_token token");
+            // Resolve the username and the email address corresponding to the user.
+            var username = await Manager.GetUserNameAsync(user);
+            var email = await Manager.GetEmailAsync(user);
+
+            // Only add the name claim if the "profile" scope was present in the token request
+            // and if the username doesn't correspond to the registered email address.
+            if (request.ContainsScope(OpenIdConnectConstants.Scopes.Profile) &&
+                !string.Equals(username, email, StringComparison.OrdinalIgnoreCase)) {
+                identity.AddClaim(ClaimTypes.Name, username, destination: "id_token token");
```
ilmax, line 134: Curious: what's the rationale behind this change?

kevinchalet, line 134: In the default ASP.NET Identity 3 template, the username is the email address. To avoid disclosing it when the `email` scope is not requested, we ensure the username is not the same as the email address before adding it.

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -115,14 +115,21 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
...
+            // Resolve the username and the email address corresponding to the user.
+            var username = await Manager.GetUserNameAsync(user);
+            var email = await Manager.GetEmailAsync(user);
+
+            // Only add the name claim if the "profile" scope was present in the token request
+            // and if the username doesn't correspond to the registered email address.
+            if (request.ContainsScope(OpenIdConnectConstants.Scopes.Profile) &&
+               (request.ContainsScope(OpenIdConnectConstants.Scopes.Email) ||
+                !string.Equals(username, email, StringComparison.OrdinalIgnoreCase))) {
+                identity.AddClaim(ClaimTypes.Name, username, destination: "id_token token");
```
ilmax, line 134: Curious: what's the rationale behind this?

kevinchalet, line 134: https://github.com/openiddict/core/pull/19#discussion_r42819996

src/OpenIddict.Core/OpenIddictController.cs
ilmax, line 134: Does this mean in some circumstances we can end up with a token without a name claim right?

kevinchalet, line 134: Yep, exactly. This is not something I'm really happy with, TBH :cry:

ilmax, line 134: Yes, this could potentially break clients

kevinchalet, line 134: What do you suggest instead?

Bartmax, line 134: why do we care? if it's the username ... not sure if OpenIddict should do anything about it. Looks like too magic and unexpected.

If we really wanto to, we can strip everything after @.
If we need uniqueness, we can hash the whole email and set it as part as the username like name@domain.com into name#{hash}.

I still think we shouldn't do anything special.

kevinchalet, line 134: Because allowing the client application to indirectly retrieve the email address without the `email` scope being granted by the user is... evil :smile: 

But maybe we should simply ensure the `email` scope has been granted and fail if we determine that the username is the email address? It would be less surprising.

ilmax, line 134: Both options looks good to me, may a failure can help the user to understand what's going on

Bartmax, line 134: another option if we prefer to avoid failures is to submit a Logger.Warning indicating that email scope wasn't asked and username contains email, return a standard name like "not available", "undisclosed", "private", etc. and make clear documentation about it.

kevinchalet, line 134: I've opted for the error path: if the username and the email address are identical, we require the `email` scope. If it's missing, an error is returned and the authorization/token request is rejected.

kevinchalet:
Merged. Thanks for the review! :tada:

PR #20, opened 2015-10-22
Do not merge - Scopes handling

ilmax:
This is the first iteration of automatic scope handling, the code is still a WIP.
/cc @PinpointTownes @Bartmax

### DIFF
--- a/src/OpenIddict.Core/OpenIddictController.cs
+++ b/src/OpenIddict.Core/OpenIddictController.cs
@@ -5,6 +5,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading;
@@ -87,7 +88,37 @@ namespace OpenIddict {
                 });
             }
 
-            return View("Authorize", Tuple.Create(request, await Manager.GetDisplayNameAsync(application)));
+            // Get all scopess associated with the current application and ensure all scopes specified in the authorization request 
+            // belongs to the current application or fail with an error, note that we automatically strip out standard **openid** scopes.
+            var requestedScopes = request.GetCustomScopes();
+
+            // Get all application scopes
+            var applicationScopes = await Manager.GetScopesByApplicationAsync(application);
+
+            // Get all request scopes
+            var requestScopes = await Manager.GetAuthorizationRequesteScopesAsync(requestedScopes);
+            
+            // TODO Add a better/smarter check
+            if (!requestScopes.Intersect(applicationScopes).SequenceEqual(requestScopes)){
+                return View("Error", new OpenIdConnectMessage{
+                    Error = OpenIdConnectConstants.Errors.InvalidRequest,
+                    ErrorDescription = "One or more of the request scopes does not belong to the calling client application"
+                });
+            }
+
+            var scopeModels = new List<ScopeModel>();
+            foreach (var scope in requestScopes)
+            {
+                // TODO this is higly inefficient in case of db lookups, so we should avoid await inside the foreach!
+                scopeModels.Add(new ScopeModel
+                {
+                    ScopeId = await Manager.GetScopeIdAsync(scope),
+                    DisplayName = await Manager.GetScopeDisplayNameAsync(scope),
+                    Description = await Manager.GetScopeDescriptionAsync(scope),
+                });
+            }
+
+            return View("Authorize", Tuple.Create(request, await Manager.GetDisplayNameAsync(application), scopeModels));
         }
 
         [Authorize, HttpPost, ValidateAntiForgeryToken]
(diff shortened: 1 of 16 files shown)


src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -86,7 +88,37 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
                 });
             }
 
-            return View("Authorize", Tuple.Create(request, await Manager.GetDisplayNameAsync(application)));
+            // Get all scopes associated with the current application and ensure all scopes specified in the authorization request 
+            // belongs to the current application or fail with an error, note that we automatically strip out standard **openid** scopes.
+            var requestedScopes = request.GetCustomScopes();
+
+            // Get all application scopes
+            var applicationScopes = await Manager.GetScopesByApplicationAsync(application);
```
ilmax, line 94: I didn't like this part at all

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -86,7 +88,37 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
...
+
+            var scopeModels = new List<ScopeModel>();
+            foreach (var scope in requestScopes)
+            {
+                // TODO this is higly inefficient in case of db lookups, so we should avoid await inside the foreach!
+                scopeModels.Add(new ScopeModel
+                {
+                    ScopeId = await Manager.GetScopeIdAsync(scope),
+                    DisplayName = await Manager.GetScopeDisplayNameAsync(scope),
+                    Description = await Manager.GetScopeDescriptionAsync(scope),
```
ilmax: this neither, all those nasty await inside a foreach are evil!

kevinchalet: Personally, I wouldn't care too much about these `await`: your default `OpenIddictStore.Get*Async` methods return completed tasks via `Task.FromResult` so the performance impact should be really minimal (no thread switch).

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -86,7 +88,37 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
...
-            return View("Authorize", Tuple.Create(request, await Manager.GetDisplayNameAsync(application)));
+            // Get all scopes associated with the current application and ensure all scopes specified in the authorization request 
+            // belongs to the current application or fail with an error, note that we automatically strip out standard **openid** scopes.
+            var requestedScopes = request.GetCustomScopes();
+
+            // Get all application scopes
+            var applicationScopes = await Manager.GetScopesByApplicationAsync(application);
+
+            // Get all request scopes
+            var requestScopes = await Manager.GetAuthorizationRequesteScopesAsync(requestedScopes);
```
kevinchalet: Curious: why do we need this method?

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -228,4 +266,50 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
...
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            return message.GetScopes().Except(GetOpenIdStandardScopes());
+        }
+
+        public static IEnumerable<string> GetOpenIdStandardScopes()
+        {
+            yield return OpenIdConnectConstants.Scopes.OpenId;
+            yield return OpenIdConnectConstants.Scopes.Email;
```
kevinchalet: Does that mean we assume that `email`, `profile`, `offline_access` (and `phone`/`address`) scopes are necessarily approved by the user?

src/OpenIddict.Core/OpenIddictController.cs
ilmax: Yes, just 3 state machines (only for the foreach) but in case the lookup happens from somewhere else (e.g. using a custom `IOpenIddictStore`) this could have a much larger performance impact.

src/OpenIddict.Core/OpenIddictController.cs
ilmax: Fair point! What do you suggest?

kevinchalet: I think we should only ignore the `openid` scope :+1:

ilmax: Do we need to provide a standard description/display name to display the standard openid connect scopes `email`, `profile`, `offline_access` (and `phone`/`address`) in the shiny consent screen? :)

src/OpenIddict.Core/OpenIddictController.cs
ilmax: To get scopes from db/somewhere for current request. In this point we have AR scopes (list of strings) and Application scopes (lists of ??) since `where TScope : class` so we need to demand all scope related work to the Store.

src/OpenIddict.Core/OpenIddictController.cs
```diff
@@ -147,6 +182,11 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
             identity.Actor.AddClaim(ClaimTypes.NameIdentifier, request.ClientId);
             identity.Actor.AddClaim(ClaimTypes.Name, await Manager.GetDisplayNameAsync(application), destination: "id_token token");
 
+            // Add selected scopes back to ASOS
+            if (Request.Form.ContainsKey("scope")) {
+                request.Scope = Request.Form["scope"];
```
kevinchalet, line 187: FYI, I just pushed a commit that should help you here: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Server/Controllers/AuthorizationController.cs#L132-L144

ilmax, line 187: Thank you! will update shortly, maybe tomorrow night

ilmax:
Will rework this

PR #24, opened 2015-10-24
Rename OpenIddictBuilder to OpenIddictServices and introduce a new OpenIddictBuilder to make options easier to configure

kevinchalet:
/cc @ilmax @Bartmax


Bartmax:
:shipit:

PR #25, opened 2015-10-26
Decouple MVC from the core project and introduce OpenIddict modules support

kevinchalet:
/cc @ilmax @Bartmax


kevinchalet:
Here's a screenshot of the "new" UI we get for free with the `CombinedFileSystemProvider` :smile: 

image

src/OpenIddict.Mvc/OpenIddictController.cs
```diff
@@ -38,7 +38,7 @@ public class OpenIddictController<TUser, TApplication> : Controller where TUser
         protected virtual OpenIddictOptions Options { get; }
 
         [HttpGet, HttpPost]
-        public async Task<IActionResult> Authorize() {
+        public virtual async Task<IActionResult> Authorize() {
```
ilmax, line 41: What's the point on making it virtual?

src/OpenIddict.Core/OpenIddictModule.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace OpenIddict {
+    /// <summary>
+    /// Defines an OpenIddict module.
+    /// </summary>
+    public class OpenIddictModule {
+        /// <summary>
+        /// Gets or sets the position of the module in the ASP.NET pipeline.
+        /// </summary>
+        public int Position { get; set; }
```
ilmax, line 12: This property is used for registering a module in the asp.net 5 pipeline, what's happening if two modules are registered using same position? since asp.net pipeline is based on registration order does we need to take care of this case?

kevinchalet, line 12: Nothing particular: we just assume the order between the modules sharing the same position doesn't matter and that they can be registered in any order between them.

src/OpenIddict.Mvc/OpenIddictController.cs
kevinchalet, line 41: Being able to subclass the default OpenIddict controller and change its logic. I'm fine sealing it if you don't like this change :+1:

ilmax, line 41: I don't have strong opinion, I expect that using custom views (which this PR addresses) is the predominant scenario, making controller action virtual seems like a dupe, having said that I'm not totally against this change.

ilmax:
The new UI looks great! :+1:

src/OpenIddict.Mvc/OpenIddictController.cs
Bartmax, line 41: virtual adds the option to subclass and override what you need and also call base. I like the virtual thing.

PR #31, opened 2015-11-13
Rename ApplicationID property to Id for better compatibility with EF

damccull:
When subclassing an EF model, the ID field of the parent should be "Id" in order to avoid situations where the model has a different name than the parent (all subclasses). EF expects the ID field to be named "Id" or "ClassNameId". Having the Application model's ID field named as anything other than "Id" prevents best compability with EF.


kevinchalet:
Merged: https://github.com/openiddict/core/commit/592cca06f07cd0e69186b96b8ca7810d27bd71e8

Thanks for your (first!) contribution :tada:

PR #36, opened 2015-12-09
Better CSP Header Handling

damccull:
Moving from NWebsec.Owin to NWebsec.Middleware which supports .net core and is compatible with ASP.net5.

There is an issue with the current version of NWebsec.Middleware: it is compiled against RC1, and so will not work with the RC2 builds of OpenIddict. Referencing the NWebsec source and changing all the RC1 references to RC2, however, allows this code to work. Maybe @klings can make an RC2 version some day.

This move is being made in conjunction with adding the ability for users to supply their own CSP headers while still using OpenIddict. I have an idea as to how to implement this, but would like some advice because I also see a possible problem or downside.

Currently the CSP headers are set in OpenIddict.Mvc project in the file OpenIddictExtensions. I would like to modify the UseMvc method in that file to accept an optional IFluentCspOptions object. If the object is supplied, OpenIddict should use it instead of the default. If not, it should use the default already in the code.

The downside I see to this (and it may just be that I don't understand how it works) is that users of OpenIddict will also have to import the same (or compatible) version of NWebsec that OpenIddict uses in order to be able to instantiate and pass an IFluentCspOptions object. Is this going to cause grief for users such that we should consider a different approach or am I overthinking this?


damccull:
Splitting the nwebsec and CSP handling into two issues and two PRs.

klings:
You're running on the dev packages from MSFT? I'll have to fix a few things, and add a few features before I can RC NWebsec. I'll be doing that against the RC1, as things aren't moving around there right now and that's what people would be using in production. Unfortunately, I don't have a lot of time available for NWebsec at the moment, so I'll have to direct my effort towards getting feature complete and won't even attempt to keep up with the dev packages. You guys are really bleeding edge. :)

BTW, I just released a new gamma, I had inadvertently introduced an issue with `data:` URIs when porting to vnext.

PR #38, opened 2015-12-10
Improved security headers

damccull:
This PR moves the nwebsec code to a new project called `OpenIddict.Security` and creates a new extension method `UseSecurityHeaders` which is enabled by default in the standard `UseOpenIddict` method. The users should see no changes at all by default.

The code also adds an optional parameter to `UseSecurityHeaders` of type `IFluentCspHeaders` that, when present, will allow the user to override the built in headers. This technique requires that the user reference `NWebsec.Middleware` in his or her own app. This is a first draft and I think we can come up with a better way to handle this.

What do you think @PinpointTownes?

### DIFF
--- a/OpenIddict.sln
+++ b/OpenIddict.sln
@@ -27,6 +27,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "external", "external", "{DE
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "NWebsec", "external\NWebsec\NWebsec.xproj", "{38C8E88F-1D01-466F-B47D-6D67F13C1594}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "OpenIddict.Security", "src\OpenIddict.Security\OpenIddict.Security.xproj", "{3744B1BC-3498-4958-B020-B2688A78B989}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -69,6 +71,10 @@ Global
 		{38C8E88F-1D01-466F-B47D-6D67F13C1594}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{38C8E88F-1D01-466F-B47D-6D67F13C1594}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{38C8E88F-1D01-466F-B47D-6D67F13C1594}.Release|Any CPU.Build.0 = Release|Any CPU
+		{3744B1BC-3498-4958-B020-B2688A78B989}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{3744B1BC-3498-4958-B020-B2688A78B989}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{3744B1BC-3498-4958-B020-B2688A78B989}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{3744B1BC-3498-4958-B020-B2688A78B989}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -83,5 +89,6 @@ Global
 		{E60CF8CA-6313-4359-BE43-AFCBB927EA30} = {D544447C-D701-46BB-9A5B-C76C612A596B}
 		{7AE46E2F-E93B-4FF9-B941-6CD7A3E1BF84} = {D544447C-D701-46BB-9A5B-C76C612A596B}
 		{38C8E88F-1D01-466F-B47D-6D67F13C1594} = {DE26CC68-28BA-44BB-B28E-43B949C6C606}
+		{3744B1BC-3498-4958-B020-B2688A78B989} = {D544447C-D701-46BB-9A5B-C76C612A596B}
 	EndGlobalSection
 EndGlobal
--- a/samples/Mvc.Server/Startup.cs
+++ b/samples/Mvc.Server/Startup.cs
@@ -10,8 +10,10 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Mvc.Server.Models;
 using Mvc.Server.Services;
+using NWebsec.Middleware;
 using OpenIddict;
 using OpenIddict.Models;
+
 
 namespace Mvc.Server {
     public class Startup {
@@ -92,7 +94,17 @@ namespace Mvc.Server {
 
             // Note: OpenIddict must be added after
             // ASP.NET Identity and the external providers.
-            app.UseOpenIddict();
+            app.UseOpenIddict(o => {
+                o.UseNWebsec(directives => {
+                    directives.DefaultSources(directive => directive.Self())
+                        .ImageSources(directive => directive.Self().CustomSources("*"))
+                        .ScriptSources(directive => directive
+                            .Self()
+                            .UnsafeInline()
+                            .CustomSources("https://my.custom.url"))
+                        .StyleSources(directive => directive.Self().UnsafeInline());
+                });
+            });
 
             app.UseMvcWithDefaultRoute();
 
--- a/src/OpenIddict.Core/OpenIddictBuilder.cs
+++ b/src/OpenIddict.Core/OpenIddictBuilder.cs
@@ -3,7 +3,7 @@ using OpenIddict;
 
 namespace Microsoft.AspNet.Builder {
     /// <summary>
-    /// Holds various properties allowing to configure OpenIddct.
+    /// Holds various properties allowing to configure OpenIddict.
     /// </summary>
     public class OpenIddictBuilder {
         /// <summary>
--- a/src/OpenIddict.Core/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Core/OpenIddictExtensions.cs
@@ -49,9 +49,10 @@ namespace Microsoft.AspNet.Builder {
             [NotNull] this OpenIddictBuilder builder,
             [NotNull] string name, int position,
             [NotNull] Action<IApplicationBuilder> registration) {
-            // By default, prevent duplicate registrations.
-            if (builder.Modules.Any(module => string.Equals(module.Name, name))) {
-                return builder;
+            // By default, replace duplicate registrations.
+            var replacedModule = builder.Modules.Where(module => string.Equals(module.Name, name)).ToList();
+            foreach (var module in replacedModule) {
+                builder.Modules.Remove(module);
             }
 
             builder.Modules.Add(new OpenIddictModule {
@@ -81,12 +82,6 @@ namespace Microsoft.AspNet.Builder {
 
             configuration(builder);
 
-            builder.AddModule("CORS", -10, map => map.UseCors(options => {
-                options.AllowAnyHeader();
-                options.AllowAnyMethod();
-                options.AllowAnyOrigin();
-                options.AllowCredentials();
-            }));
 
             // Add OpenIdConnectServerMiddleware to the ASP.NET 5 pipeline.
             builder.AddModule("ASOS", 0, map => map.UseOpenIdConnectServer(builder.Options));
--- a/src/OpenIddict.Security/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Security/OpenIddictExtensions.cs
@@ -1,0 +1,58 @@
+﻿using Microsoft.Extensions.Internal;
+using NWebsec.Middleware;
+using System;
+
+namespace Microsoft.AspNet.Builder {
+    public static class OpenIddictExtensions {
+        public static OpenIddictBuilder UseNWebsec([NotNull] this OpenIddictBuilder builder, Action<IFluentCspOptions> cspOptions) {
+
+            //Add security headers to the app.
+            return builder.AddModule("NWebsec", -20, app => {
+                // Insert a new middleware responsible of setting the Content-Security-Policy header.
+                // See https://nwebsec.codeplex.com/wikipage?title=Configuring%20Content%20Security%20Policy&referringTitle=NWebsec
+                app.UseCsp(options => cspOptions(options));
+
+                SetXHeaders(app);
+            });
+        }
+
+        public static OpenIddictBuilder UseNWebsec([NotNull] this OpenIddictBuilder builder) {
+            //Add security headers to the app.
+            return builder.AddModule("NWebsec", -20, app => {
+                // Insert a new middleware responsible of setting the Content-Security-Policy header.
+                // See https://nwebsec.codeplex.com/wikipage?title=Configuring%20Content%20Security%20Policy&referringTitle=NWebsec
+                app.UseCsp(options => options.DefaultSources(directive => directive.Self())
+                    .ImageSources(directive => directive.Self().CustomSources("*"))
+                    .ScriptSources(directive => directive.Self().UnsafeInline())
+                    .StyleSources(directive => directive.Self().UnsafeInline()));
+                SetXHeaders(app);
+            });
+        }
+
+        public static OpenIddictBuilder UseCORS([NotNull] this OpenIddictBuilder builder) {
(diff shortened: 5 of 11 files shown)


damccull:
@klings, answering you here instead of https://github.com/openiddict/core/pull/36#issuecomment-163444256 since this is the proper PR now.

Yeah, for some reason or other @PinpointTownes likes to work on code that has to be changed every day :D so I do too since I like his library. For now, we have created an internal copy of your library and built it in instead of referencing. It's more work to maintain at the moment, but much less than writing our own library, and I think you probably won't change yours every day anyhow.

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.AspNet.Builder {
     /// <summary>
-    /// Holds various properties allowing to configure OpenIddct.
+    /// Holds various properties allowing to configure OpenIddict.
```
kevinchalet, line 6: lol, nice catch! :clap:

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -15,5 +15,10 @@ public class OpenIddictBuilder {
         /// Gets or sets the options used by OpenIddict.
         /// </summary>
         public OpenIddictOptions Options { get; set; } = new OpenIddictOptions();
+
+        /// <summary>
+        /// Gets or sets a dictionary of properties. Store anything you need in here.
+        /// </summary>
+        public Dictionary<string, object> Properties { get; set; }
```
kevinchalet: Please always initialize collection properties and remove the setter :clap:

src/OpenIddict.Security/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Reflection;
```
kevinchalet: Please kill this file with fire! :fire:

src/OpenIddict.Security/project.json
```diff
@@ -0,0 +1,22 @@
+﻿{
+    "version": "1.0.0-*",
+
+    "description": "Security headers module for OpenIddict.",
+
+    "dependencies": {
+        "OpenIddict.Core": "1.0.0-*",
+        "NWebsec.Middleware": {
+            "type": "build",
+            "version": "1.0.0-gamma-custom-*"
```
kevinchalet: Please note I renamed the package and changed the version in a really recent commit.

src/OpenIddict/OpenIddictExtensions.cs
```diff
@@ -38,6 +38,7 @@ public static IdentityBuilder AddOpenIddict<TApplication>([NotNull] this Identit
                 // and the MVC modules are enabled.
                 builder.UseAssets();
                 builder.UseMvc();
+                builder.UseSecurityHeaders();
```
kevinchalet: nit: modules order doesn't matter, but I'd put this one at the top :smile:

src/OpenIddict.Security/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Microsoft.Extensions.Internal;
+using NWebsec.Middleware;
+using System;
+
+namespace Microsoft.AspNet.Builder {
+    public static class OpenIddictExtensions {
+        public static OpenIddictBuilder UseSecurityHeaders([NotNull] this OpenIddictBuilder builder, Action<IFluentCspOptions> cspOptions = null) {
+            
+            //Add security headers to the app.
+            builder.AddModule(-20, app => {
```
kevinchalet: Please specify a module name here (or it won't compile at all).

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet: Looks like this property is not used?

kevinchalet:
I wonder if we shouldn't directly integrate NWebsec in the core project :confused:

src/OpenIddict.Core/OpenIddictBuilder.cs
damccull: Not yet. Will remove if it doesn't ever get used.

src/OpenIddict.Security/Properties/AssemblyInfo.cs
damccull: System. Reflection? I thought you added it. I rebased on your changes and had a conflict. Np.

kevinchalet: No, I mean the whole file :smile:

src/OpenIddict/OpenIddictExtensions.cs
damccull: OK. I tend to add to the end by default because I don't always know what setup needs done before my code.

src/OpenIddict.Security/Properties/AssemblyInfo.cs
damccull: Haha OK. I added cause you had one elsewhere.

damccull:
Do you mean you're thinking maybe we should just remove nwebsec from the core project and let the user add it if they want? That would make the project more specific to what it does rather than adding in a third party component that users may choose to use on their own anyways. We can always recommend security headers in the documentation.

damccull:
I'm not sure whether we should move it to OpenIddict.Core or not. Doing that would force its use if we integrate it directly, and some users may choose to implement their own solution, adding OpenIddict modules individually instead of just using UseOpenIddict.

Additionally, having it in a separate module and calling it as part of the default setup activates it by default anyways.

Finally, should we ever choose to move away from NWebsec to another library (or custom code), replacing the module will be easier than trying to disintegrate it from .Core.

My vote is no, but up to you.

ghost:
I agree with damccull

damccull:
OK. Here's a second draft for approval. This code:
- Moves security header functionality into its own project
- Adds the ability for users to add their own headers to the defaults
- Retains the ability to just use the defaults (by default)
- Adds a method UseOpenIddictNoSecurityHeaders() (originally a workaround) which will do everything BUT the security headers. Just in case someone wants it.

@PinpointTownes What you think? Works properly in the demo projects and in the project I'm actually using this all for.

OpenIddict.sln
```diff
@@ -27,6 +27,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "external", "external", "{DE
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "NWebsec", "external\NWebsec\NWebsec.xproj", "{38C8E88F-1D01-466F-B47D-6D67F13C1594}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "OpenIddict.Security", "src\OpenIddict.Security\OpenIddict.Security.xproj", "{3744B1BC-3498-4958-B020-B2688A78B989}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "NWebsec.Annotations", "external\NWebsec.Annotations\NWebsec.Annotations.xproj", "{00BD4606-DE8D-4C4F-92FF-3A5B2B1E83A2}"
```
kevinchalet: Please revert these changes (except `OpenIddict.Security` of course :smile:)

external/NWebsec.Annotations/NWebsec.Annotations.xproj
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
kevinchalet: Kill that. Hard.

external/NWebsec.Core/NWebsec.Core.xproj
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
kevinchalet: That too.

external/NWebsec.Middleware/NWebsec.Middleware.xproj
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
kevinchalet: And also that :smile:

src/OpenIddict/OpenIddictExtensions.cs
```diff
@@ -38,7 +38,29 @@ public static IdentityBuilder AddOpenIddict<TApplication>([NotNull] this Identit
...
 
+        /// <summary>
+        /// This method will enable all the default OpenIddict options EXCEPT for the NWebsec security headers
+        /// such as the Content Security Policy headers. This means you must implement all headers yourself if
+        /// you want them.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <param name="configuration"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseOpenIddictNoSecurityHeaders(
```
kevinchalet: I don't think we should offer such a method. If you _really_ want to prevent the security headers from being applied, using `UseOpenIddictCore` to roll your own configuration is the way to go.

src/OpenIddict.Security/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Microsoft.Extensions.Internal;
+using NWebsec.Middleware;
+using System;
+
+namespace Microsoft.AspNet.Builder {
+    public static class OpenIddictExtensions {
+        public static OpenIddictBuilder UseSecurityHeaders([NotNull] this OpenIddictBuilder builder, Action<IFluentCspOptions> cspOptions = null) {
```
kevinchalet: Should we call this method `UseNWebsec`?

external/NWebsec.Annotations/NWebsec.Annotations.xproj
damccull: OK but what's wrong with xml encoding tags?

src/OpenIddict.Security/OpenIddictExtensions.cs
damccull: Well that depends on how specific you want to be. Calling it UseNWebsec ties it very closely to that library from a user's perspective and prevents us changing the implementation to another library. However, considering that is already tied to NWebsec via the IFluentCspOptions parameter, that could be a valid name too. If we had our own abstraction for setting headers I'd argue against a rename. In this case, what do you want?

src/OpenIddict/OpenIddictExtensions.cs
damccull: Will nuke.

external/NWebsec.Annotations/NWebsec.Annotations.xproj
kevinchalet: I mean the entire file, not just the XML declaration :smile:

src/OpenIddict.Security/OpenIddictExtensions.cs
kevinchalet: As you mentioned, it's closely tied to NWebsec, so `UseNWebsec` is a good name :smile:

damccull:
OK @PinpointTownes, what you think now?

samples/Mvc.Server/Startup.cs
```diff
@@ -10,8 +9,10 @@
 using Microsoft.Extensions.Logging;
 using Mvc.Server.Models;
 using Mvc.Server.Services;
+using NWebsec.Middleware;
 using OpenIddict;
 using OpenIddict.Models;
+using System.Linq;
```
kevinchalet: Please put the `System` usings at the top :+1:

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -1,9 +1,9 @@
-﻿using System.Collections.Generic;
-using OpenIddict;
+﻿using OpenIddict;
```
kevinchalet: Same remark.

src/OpenIddict.Core/OpenIddictExtensions.cs
```diff
@@ -49,9 +49,10 @@ public static class OpenIddictExtensions {
             [NotNull] this OpenIddictBuilder builder,
             [NotNull] string name, int position,
             [NotNull] Action<IApplicationBuilder> registration) {
-            // By default, prevent duplicate registrations.
-            if (builder.Modules.Any(module => string.Equals(module.Name, name))) {
-                return builder;
+            // By default, replace duplicate registrations.
+            var replacedModule = builder.Modules.Where(module => string.Equals(module.Name, name)).ToList();
```
kevinchalet, line 53: nit: no need for `ToList` here :smile:

src/OpenIddict.Security/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Microsoft.AspNet.Builder {
+    public static class OpenIddictExtensions {
+        public static OpenIddictBuilder UseNWebsec([NotNull] this OpenIddictBuilder builder, Action<IFluentCspOptions> cspOptions) {
+
+            //Add security headers to the app.
+            return builder.AddModule("NWebsec", -20, app => {
+                // Insert a new middleware responsible of setting the Content-Security-Policy header.
+                // See https://nwebsec.codeplex.com/wikipage?title=Configuring%20Content%20Security%20Policy&referringTitle=NWebsec
+                app.UseCsp(options => {
```
kevinchalet: Consider using just here `app.UseCsp(cspOptions)` and moving the default directives to the other overload.

src/OpenIddict.Security/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                app.UseXfo(options => options.Deny());
+
+                // Insert a new middleware responsible of setting the X-Xss-Protection header.
+                // See https://nwebsec.codeplex.com/wikipage?title=Configuring%20security%20headers&referringTitle=NWebsec
+                app.UseXXssProtection(options => options.EnabledWithBlockMode());
+            });
+        }
+        
+        public static OpenIddictBuilder UseNWebsec([NotNull] this OpenIddictBuilder builder) {
+            return builder.UseNWebsec(null);
```
kevinchalet: Like that:

``` csharp
return builder.UseNWebsec(options => options.DefaultSources(directive => directive.Self())
                                            .ImageSources(directive => directive.Self().CustomSources("*"))
                                            .ScriptSources(directive => directive.Self().UnsafeInline())
                                            .StyleSources(directive => directive.Self().UnsafeInline());
```

src/OpenIddict/OpenIddictExtensions.cs
```diff
@@ -4,10 +4,10 @@
  * the license and the contributors participating to this project.
  */
 
-using System;
 using Microsoft.AspNet.Identity;
 using Microsoft.Extensions.Internal;
 using OpenIddict.Models;
+using System;
```
kevinchalet: nit: same remark.

src/OpenIddict.Security/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public static OpenIddictBuilder UseCORS([NotNull] this OpenIddictBuilder builder) {
```
kevinchalet, line 32: Should that be `UseCors` to match the official name?

src/OpenIddict.Core/OpenIddictExtensions.cs
damccull, line 53: Actually you DO need a `ToList()`. If not, the foreach freaks out saying that the collection was modified and it can't iterate. Linq queries don't actually enumerate until the last second. Inside a foreach doesn't always work right. The `ToList()` forces it to enumerate before we use the foreach on the resultant collection, ensuring that the results can't change while the foreach is iterating. I tried without it :)

kevinchalet, line 53: Ah yeah, nice catch. I had not realized we were altering the collection in the foreach block :smile:

damccull:
Ok @PinpointTownes press the BIG RED BUTTON.

kevinchalet:
Merged, thanks @damccull! :clap: :clap:

https://github.com/openiddict/core/commit/dad1b307b88e20dd9dd3a01c485ddcfbfeff189c

PR #42, opened 2016-01-05
Use latest rc2 build for CryptoHelper

henkmollema:
Since you are targeting RC2 bits of ASP.NET, you should use the RC2 build of CryptoHelper too.


kevinchalet:
Merged, thanks! :clap: 

https://github.com/openiddict/core/commit/e40d3122aadad4bc013e12cc2e6b1a04c01bf6fc

PR #83, opened 2016-04-03
Reacting to MVC API changes

joshcomley:
https://github.com/aspnet/Mvc/pull/4391/files


joshcomley:
Line 142, `Register the assembly provider in the isolated container.`, I've just commented it out as I don't think it's needed any more because `IAssemblyProvider` is replaced with a static, non-interfaced class `DefaultAssemblyPartDiscoveryProvider`, and the controller discovery is done completely differently anyway.

kevinchalet:
Merged: https://github.com/openiddict/openiddict-core/commit/5d0bfc9b4a251d37536871195fe875bfccd737f4.

Thanks Josh! :clap:

Issue #18, opened 2015-10-21
Add built-in logging


kevinchalet:
We should update `OpenIddictManager` and/or `OpenIddictProvider` to log important events, like ASP.NET Identity does: https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/UserManager.cs#L586

ilmax:
Self assigned

ilmax:
Can this be closed or do you plan other changes?

kevinchalet:
Looks like we have a solid logging base now. I think we can indeed close this ticket :+1: 

Thanks :clap:

PR #122, opened 2016-05-23
[WIP] log all the things

ilmax:
Initial work around logging, put up for early feedback, addresses #18

### DIFF
--- a/src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
+++ b/src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
@@ -92,7 +92,7 @@ namespace OpenIddict {
                     var email = await services.Users.GetEmailAsync(user);
 
                     if (!string.IsNullOrEmpty(email) && string.Equals(username, email, StringComparison.OrdinalIgnoreCase)) {
-                        logger.LogWarning("The username correspond to the email address and we carefully avoid leaking the user email if the 'email' scope is not requested.");
+                        logger.LogWarning("The username '{UserName}' correspond to the email address and we carefully avoid leaking the user email if the 'email' scope is not requested.", username);
                         context.Reject(
                             error: OpenIdConnectConstants.Errors.InvalidRequest,
                             description: "The 'email' scope is required.");
@@ -156,11 +156,11 @@ namespace OpenIddict {
         }
 
         public override async Task HandleAuthorizationRequest([NotNull] HandleAuthorizationRequestContext context) {
-            // Only handle prompt=none requests at this stage.
             var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<OpenIddictProvider<TUser, TApplication>>>();
 
+            // Only handle prompt=none requests at this stage.
             if (!string.Equals(context.Request.Prompt, "none", StringComparison.Ordinal)) {
-                logger.LogDebug($"The current prompt is '{context.Request.Prompt}' so skipping HandleAuthorizationRequest.");
+                logger.LogDebug("The current prompt is '{Prompt}', skipping the current request.", context.Request.Prompt);
                 return;
             }
 
@@ -169,7 +169,10 @@ namespace OpenIddict {
             // Note: principal is guaranteed to be non-null since ValidateAuthorizationRequest
             // rejects prompt=none requests missing or having an invalid id_token_hint.
             var principal = await context.HttpContext.Authentication.AuthenticateAsync(context.Options.AuthenticationScheme);
-            Debug.Assert(principal != null);
+            if (principal == null) {
+                logger.LogDebug("The current principal is null, throwing exception.");
+                throw new InvalidOperationException("The current principal is null");
+            }
 
             // Note: user may be null if the user was removed after
             // the initial check made by ValidateAuthorizationRequest.
@@ -177,7 +180,7 @@ namespace OpenIddict {
             // continue to the next middleware in the pipeline.
             var user = await services.Users.GetUserAsync(principal);
             if (user == null) {
-                logger.LogWarning("Unable to retrieve user from the current principal.");
+                logger.LogWarning("Unable to retrieve user from the current principal, ensure the user was not removed from the database.");
                 return;
             }
 
@@ -185,13 +188,10 @@ namespace OpenIddict {
             // and OpenIddictProvider.GrantResourceOwnerCredentials are expected to reject requests that
             // don't include the "email" scope if the username corresponds to the registed email address.
             var identity = await services.Applications.CreateIdentityAsync(user, context.Request.GetScopes());
-
             if (identity == null) {
-                logger.LogDebug("There was an error during identity creation for user {User}, the current identity is null.", user);
-                // TODO should we fail here witha gracefoul exception other than ArgumentNullException?
-                // http://referencesource.microsoft.com/#mscorlib/system/security/claims/ClaimsPrincipal.cs,194
+                logger.LogDebug("There was an error during identity creation for user '{NameIdentifier}', the current identity is null, throwing exception.", principal.FindFirst(ClaimTypes.NameIdentifier)?.Value);
+                throw new InvalidOperationException("There was an error during identity creation.");
             }
-            Debug.Assert(identity != null);
 
             // Create a new authentication ticket holding the user identity.
             var ticket = new AuthenticationTicket(
@@ -202,12 +202,10 @@ namespace OpenIddict {
             ticket.SetResources(context.Request.GetResources());
             ticket.SetScopes(context.Request.GetScopes());
 
-            logger.LogDebug("AuthenticationTicket generated succesfully.");
+            logger.LogDebug("An authentication ticket was successfully generated for user '{NameIdentifier}'.", principal.FindFirst(ClaimTypes.NameIdentifier)?.Value);
 
             // Call SignInAsync to create and return a new OpenID Connect response containing the serialized code/tokens.
             await context.HttpContext.Authentication.SignInAsync(ticket.AuthenticationScheme, ticket.Principal, ticket.Properties);
-
-            logger.LogDebug("SignInAsync called.");
 
             // Mark the response as handled
             // to skip the rest of the pipeline.
--- a/src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
+++ b/src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
@@ -17,8 +17,7 @@ using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
-namespace OpenIddict
-{
+namespace OpenIddict {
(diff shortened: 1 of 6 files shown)


src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -16,17 +15,21 @@
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
-namespace OpenIddict {
+namespace OpenIddict
```
ilmax: will revert this :)

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -16,17 +15,21 @@
...
         public override async Task ValidateAuthorizationRequest([NotNull] ValidateAuthorizationRequestContext context) {
             var services = context.HttpContext.RequestServices.GetRequiredService<OpenIddictServices<TUser, TApplication>>();
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<OpenIddictProvider<TUser, TApplication>>>();
 
             // Note: redirect_uri is not required for pure OAuth2 requests
             // but this provider uses a stricter policy making it mandatory,
             // as required by the OpenID Connect core specification.
             // See http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest.
             if (string.IsNullOrEmpty(context.RedirectUri)) {
+                logger.LogWarning("The required redirect_uri parameter was missing.");
```
kevinchalet: FYI, ASOS beta5 now logs the error code and the description you specify when calling `context.Reject(...)` so it may be redundant. Consider removing the log messages that are equivalent to the error description.

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -37,6 +40,7 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
             // Retrieve the application details corresponding to the requested client_id.
             var application = await services.Applications.FindApplicationByIdAsync(context.ClientId);
             if (application == null) {
+                logger.LogWarning($"Application not found in the database with client_id '{context.ClientId}'.");
```
kevinchalet: You should consider using semantic logging instead:

``` csharp
logger.LogWarning("Application not found in the database with client_id '{ClientId}.'", context.ClientId);
```

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -118,9 +128,10 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
                 }
 
                 // Extract the principal contained in the id_token_hint parameter.
-                // If no principal can be extracted, an error is returned to the client application.
+                // If no principal can be extracted, an error is returned to the client aplication.
```
kevinchalet: Pretty sure there are 2 'p' in 'application' :smile:

ilmax: hahahha that was unintentional :)

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -140,6 +152,7 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
                 }
             }
 
+            logger.LogInformation("ValidateAuthorizationRequestContext is validated succesfully.");
```
kevinchalet: Definitely something we should add in ASOS (https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/52#issuecomment-210710325).

In the meantime, it's probably fine to have it here. That said, maybe something like "The authorization request was successfully validated." might be better :smile:

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
ilmax: ok, will add a comment message to keep trace that log is demanded to ASOS

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -180,12 +202,17 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
             ticket.SetResources(context.Request.GetResources());
             ticket.SetScopes(context.Request.GetScopes());
 
+            logger.LogDebug("AuthenticationTicket generated succesfully.");
+
             // Call SignInAsync to create and return a new OpenID Connect response containing the serialized code/tokens.
             await context.HttpContext.Authentication.SignInAsync(ticket.AuthenticationScheme, ticket.Principal, ticket.Properties);
 
+            logger.LogDebug("SignInAsync called.");
```
ilmax: does this make sense?

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -162,13 +177,20 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
             // continue to the next middleware in the pipeline.
             var user = await services.Users.GetUserAsync(principal);
             if (user == null) {
+                logger.LogWarning("Unable to retrieve user from the current principal.");
```
kevinchalet: Sounds like an error to me. Maybe we should also mention the most common case of this error: when the user was removed from the DB.

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -162,13 +177,20 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
...
             }
 
             // Note: filtering the username is not needed at this stage as OpenIddictController.Accept
             // and OpenIddictProvider.GrantResourceOwnerCredentials are expected to reject requests that
             // don't include the "email" scope if the username corresponds to the registed email address.
             var identity = await services.Applications.CreateIdentityAsync(user, context.Request.GetScopes());
+
+            if (identity == null) {
+                logger.LogDebug("There was an error during identity creation for user {User}, the current identity is null.", user);
+                // TODO should we fail here witha gracefoul exception other than ArgumentNullException?
```
kevinchalet: We can (or even should, as one can override `CreateIdentityAsync` to potentially return `null`). I think we should just throw an `InvalidOperationException` at this stage.

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
```diff
@@ -180,12 +202,17 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
             ticket.SetResources(context.Request.GetResources());
             ticket.SetScopes(context.Request.GetScopes());
 
+            logger.LogDebug("AuthenticationTicket generated succesfully.");
```
kevinchalet: What about something like `logger.LogDebug("An authentication ticket was successfully generated for user {NameIdentifier}.", principal.FindFirst(ClaimTypes.NameIdentifier)?.Value);`?

ilmax: I like it, will do

src/OpenIddict.Core/OpenIddictProvider.Exchange.cs
```diff
@@ -91,14 +99,21 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
...
         }
 
         public override async Task GrantClientCredentials([NotNull] GrantClientCredentialsContext context) {
             var services = context.HttpContext.RequestServices.GetRequiredService<OpenIddictServices<TUser, TApplication>>();
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<OpenIddictProvider<TUser, TApplication>>>();
 
             // Retrieve the application details corresponding to the requested client_id.
             var application = await services.Applications.FindApplicationByIdAsync(context.ClientId);
+            if (application == null) {
+                logger.LogDebug("There was an error finding application for client_id {ClientId}, the current application is null.", context.ClientId);
```
kevinchalet: Another `InvalidOperationException`?

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
ilmax: Does LogDebug plays nicely with nulls? because of `FindFirst(...)?.Value`

src/OpenIddict.Core/OpenIddictProvider.Exchange.cs
ilmax: will do

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
kevinchalet: Yep, at least, it's supposed to handle them without crashing :smile:

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
kevinchalet: Yes, but that would be great if we could avoid introducing meaningless messages like the ASP.NET team did for the OIDC client middleware :smile: 

What about something like `logger.LogInformation("The user session was successfully refreshed and a new access token was returned to the client application.");`? Maybe with some more details...

src/OpenIddict.Core/OpenIddictProvider.Exchange.cs
```diff
@@ -64,6 +69,7 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
 
             // Reject tokens requests containing a client_secret if the client application is not confidential.
             if (await services.Applications.IsPublicApplicationAsync(application) && !string.IsNullOrEmpty(context.ClientSecret)) {
+                logger.LogWarning("Public application {ClientId} are not allowed to send a client secret.", context.ClientId);
```
kevinchalet: Singular/plural issue spotted :smile:

ilmax: Good catch!

src/OpenIddict.Core/OpenIddictProvider.Authentication.cs
ilmax: I'm thinking about removing it because we already logged `logger.LogInformation("The authorization request was successfully handled.");`
What about you?

kevinchalet:
FYI, I'll fix the conflicts and merge it tonight or tomorrow :clap:

ilmax:
nice!

kevinchalet:
Merged, thanks! https://github.com/openiddict/openiddict-core/commit/758e1266af5f1d2f1a764833760fb212f1fc7104

FYI, I removed a few log messages that were generic enough to be included directly in ASOS. Since we'll do that for beta6, it was not worth adding them at this point.

PR #124, opened 2016-05-25
Update the authorization view to hide the resources list when the "resource" parameter is empty

kevinchalet:


ilmax:
LGTM

kevinchalet:
@ilmax thanks for your review :clap:

PR #127, opened 2016-06-01
Refactor OpenIddict/OpenIddict.Core, introduce new entities and add built-in refresh token revocation

kevinchalet:
Main changes:
- `services.AddOpenIddict()`/`services.AddOpenIddictCore()` have been merged into `services.AddOpenIddict()`: all the `services.AddOpenIddict()` overloads **only register the core services** in the DI container, but the extensions taking a `TContext` generic argument can be used to easily register the default Entity Framework stores and the default entities.
- `app.UseOpenIddict()`/`app.UseOpenIddictCore()` have been merged into `app.UseOpenIddict()`, which is now parameterless. The OpenIddict options/modules are now configured using `OpenIddictBuilder` (from `ConfigureServices`), which is consistent with what Identity does.
- `app.UseOpenIddict()` no longer registers the assets/MVC/security modules by default. The resource owner password grant is much more popular than I had originally expected. It's actually so popular that it's probably the most common flow used with OpenIddict. Making the assets/MVC/security modules opt-in features reflects that.
- `OpenIddict.Models` has been removed. The default models are now included in the Entity Framework package (renamed to `OpenIddict.EntityFramework`), similarly to what Identity does.
- `IOpenIddictStore` and `OpenIddictManager` have been split into separate classes (e.g `OpenIddictAuthorizationManager`/`OpenIddictTokenManager`).
- The application user entity now has to inherit from the new `OpenIddictUser` entity when using the EF stores. A generic constraint on the `services.AddOpenIddict()` taking a `TContext` parameter enforces that.
- Automatic token revocation has been implemented as part of this (giant) PR. The actual behaviour differs depending on whether sliding expiration is enabled or not:
  - When sliding expiration is enabled (default value), the refresh token can be used only once: it is automatically revoked when a new one is issued.
  - When sliding expiration is disabled, the refresh token is not automatically revoked but can be revoked manually using `OpenIddictTokenManager`.

/cc @ilmax


src/OpenIddict.Core/Infrastructure/OpenIddictServices.cs
```diff
@@ -12,20 +12,25 @@
...
             Services = services;
         }
 
         /// <summary>
-        /// Gets the <see cref="OpenIddictManager{TUser, TApplication}"/>.
+        /// Gets the <see cref="OpenIddictApplicationManager{TApplication}"/>.
         /// </summary>
-        public virtual OpenIddictManager<TUser, TApplication> Applications {
-            get { return Services.GetRequiredService<OpenIddictManager<TUser, TApplication>>(); }
+        public virtual OpenIddictApplicationManager<TApplication> Applications {
```
ilmax: nit: all those properties can use the C# 6 sysntax of expression bodied members

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -1,19 +1,404 @@
...
+
+            return this;
+        }
+
+        /// <summary>
+        /// Replaces the default application manager by a custom manager.
+        /// </summary>
+        /// <typeparam name="TManager">The type of the custom manager.</typeparam>
+        /// <returns>The <see cref="OpenIddictBuilder"/>.</returns>
+        public virtual OpenIddictBuilder ReplaceApplicationManager<TManager>() {
```
ilmax: nit: What about using "Withers" like method names? e.g WithApplicationManager<TManager>()?

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -1,19 +1,404 @@
...
+            var contract = typeof(OpenIddictApplicationManager<>).MakeGenericType(ApplicationType);
+            if (!contract.IsAssignableFrom(typeof(TManager))) {
+                throw new InvalidOperationException("Custom managers must be derived from OpenIddictApplicationManager.");
+            }
```
ilmax, line 92: This piece of code to ensure the user provided type conforms to the expectation is duplicated, can be refactored into a private method with a couple of type parameters

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -1,19 +1,404 @@
...
+            if (string.IsNullOrEmpty(name)) {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (registration == null) {
+                throw new ArgumentNullException(nameof(registration));
+            }
+
+            return Configure(options => {
+                if (options.Modules.Any(module => module.Name == name)) {
```
ilmax, line 231: consider log a warning to make debugging this kind of errors simpler

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -1,19 +1,404 @@
...
+            return Configure(options => options.RefreshTokenLifetime = lifetime);
+        }
 
         /// <summary>
-        /// Gets or sets the options used by OpenIddict.
+        /// Sets JWT as the default token format for access tokens.
         /// </summary>
-        public OpenIddictOptions Options { get; set; } = new OpenIddictOptions();
+        /// <returns>The <see cref="OpenIddictBuilder"/>.</returns>
+        public virtual OpenIddictBuilder UseJsonWebTokens() {
```
ilmax, line 427: nit: not sure if `UseJsonWebTokens` is better than `UseJwtTokens`.  :)

src/OpenIddict.Core/OpenIddictExtensions.cs
```diff
@@ -5,290 +5,103 @@
...
+                Registration = builder => builder.UseOpenIdConnectServer(options)
+            });
 
             // Register the OpenIddict modules in the ASP.NET Core pipeline.
-            foreach (var module in builder.Modules.OrderBy(module => module.Position)) {
-                if (module.Registration == null) {
-                    throw new InvalidOperationException("The registration delegate cannot be null.");
+            foreach (var module in modules.OrderBy(module => module.Position)) {
+                if (module?.Registration == null) {
+                    throw new InvalidOperationException("An invalid OpenIddict module was registered.");
```
ilmax, line 94: Does it make sense to include the module name in the exception as well?

src/OpenIddict.Core/OpenIddictModule.cs
```diff
@@ -1,24 +1,32 @@
...
+using System.Diagnostics;
 using Microsoft.AspNetCore.Builder;
 
 namespace OpenIddict {
     /// <summary>
-    /// Defines an OpenIddict module.
+    /// Represents an OpenIddict module.
     /// </summary>
+    [DebuggerDisplay("{Name,nq}")]
     public class OpenIddictModule {
```
ilmax, line 17: nit: this class can be made immutable since it's not supposed to change after `UseOpenIddict` has  been called

src/OpenIddict.Core/Stores/IOpenIddictTokenStore.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <returns>A <see cref="Task"/> that can be used to monitor the asynchronous operation.</returns>
+        Task<TToken> FindByIdAsync(string identifier, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Revokes a token.
+        /// </summary>
+        /// <param name="token">The token to revoke.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
+        /// <returns>A <see cref="Task"/> that can be used to monitor the asynchronous operation.</returns>
+        Task RevokeAsync(TToken token, CancellationToken cancellationToken);
```
ilmax, line 39: :tada: :)

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
```diff
@@ -0,0 +1,140 @@
...
+        where TContext : DbContext
+        where TKey : IEquatable<TKey> {
+        public OpenIddictApplicationStore(TContext context) {
+            Context = context;
+        }
+
+        /// <summary>
+        /// Gets the database context associated with the current store.
+        /// </summary>
+        public virtual TContext Context { get; }
```
ilmax: Are there any specific reasons for make this property public? otherwise can we turn this into a protected property?

Same applies to the `Applications` property as well and in general to all *Store classes

src/OpenIddict.EntityFramework/Stores/OpenIddictTokenStore.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// </summary>
+        /// <param name="identifier">The unique identifier associated with the token to create.</param>
+        /// <param name="type">The token type.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
+        /// <returns>A <see cref="Task"/> that can be used to monitor the asynchronous operation.</returns>
+        public virtual Task CreateAsync(string identifier, string type, CancellationToken cancellationToken) {
+            var converter = TypeDescriptor.GetConverter(typeof(TKey));
+
+            // If the string key cannot be converted to TKey, return null
+            // to indicate that the requested token doesn't exist.
```
ilmax: Shouldn't this comment say that the token cannot be created/stored instead? We are storing a token, not querying for one :)

src/OpenIddict.Core/Stores/IOpenIddictApplicationStore.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Threading.Tasks;
+
+namespace OpenIddict {
+    /// <summary>
+    /// Provides methods allowing to manage the applications stored in a database.
+    /// </summary>
+    /// <typeparam name="TApplication">The type of the Application entity.</typeparam>
+    public interface IOpenIddictApplicationStore<TApplication> where TApplication : class {
+        /// <summary>
+        /// Creates a new application.
```
ilmax, line 17: Stores a new application?

src/OpenIddict.Mvc/OpenIddictController.cs
```diff
@@ -15,23 +15,20 @@
...
-        protected virtual OpenIddictServices<TUser, TApplication> Services { get; }
-
+    public class OpenIddictController<TUser, TApplication, TAuthorization, TToken> : Controller
+        where TUser : class
+        where TApplication : class
+        where TAuthorization : class
+        where TToken : class {
         [HttpGet, HttpPost]
-        public virtual async Task<IActionResult> Authorize() {
+        public virtual async Task<IActionResult> Authorize([FromServices] OpenIddictApplicationManager<TApplication> applications) {
```
ilmax, line 27: nit: can `applications` be renamed to `applicationManager`?

src/OpenIddict.Mvc/OpenIddictController.cs
```diff
@@ -15,23 +15,20 @@
...
             // Note: when a fatal error occurs during the request processing, an OpenID Connect response
             // is prematurely forged and added to the ASP.NET context by OpenIdConnectServerHandler.
```
ilmax: nit: Missing core suffix to ASP.NET in comment

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Exchange.cs
```diff
@@ -125,7 +131,7 @@ public partial class OpenIddictProvider<TUser, TApplication> : OpenIdConnectServ
         }
 
         public override async Task GrantRefreshToken([NotNull] GrantRefreshTokenContext context) {
-            var services = context.HttpContext.RequestServices.GetRequiredService<OpenIddictServices<TUser, TApplication>>();
+            var services = context.HttpContext.RequestServices.GetRequiredService<OpenIddictServices<TUser, TApplication, TAuthorization, TScope, TToken>>();
             var options = context.HttpContext.RequestServices.GetRequiredService<IOptions<IdentityOptions>>();
```
ilmax: nit: `options` variable is not used in this method and can be safely removed

src/OpenIddict.EntityFramework/Models/OpenIddictScope.cs
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+using System;
+
+namespace OpenIddict {
+    /// <summary>
+    /// Represents an OpenIddict scope.
+    /// </summary>
+    public class OpenIddictScope : OpenIddictScope<string> {
+        public OpenIddictScope() : base() { }
```
ilmax: nit: this ctor is redundant

src/OpenIddict.EntityFramework/Models/OpenIddictUser.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
+
+namespace OpenIddict {
+    /// <summary>
+    /// Represents an OpenIddict user.
+    /// </summary>
+    public class OpenIddictUser : OpenIddictUser<OpenIddictAuthorization, OpenIddictToken> {
+        public OpenIddictUser() : base() { }
```
ilmax: nit: this ctor is redundant

src/OpenIddict.Mvc/OpenIddictController.cs
```diff
@@ -15,23 +15,20 @@
...
-        public OpenIddictController([NotNull] OpenIddictServices<TUser, TApplication> services) {
-            Services = services;
-        }
-
-        /// <summary>
-        /// Gets the OpenIddict services used by the controller.
-        /// </summary>
-        protected virtual OpenIddictServices<TUser, TApplication> Services { get; }
-
+    public class OpenIddictController<TUser, TApplication, TAuthorization, TToken> : Controller
```
ilmax, line 24: Why this class has moved from using `OpenIddictServices` to e.g. `OpenIddictApplicationManager` directly?

ilmax:
Few nits, but LGTM.

Nice job! :tada:

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Exchange.cs
kevinchalet: Nice catch, will remove :+1:

src/OpenIddict.Core/Infrastructure/OpenIddictServices.cs
kevinchalet: Yep, good idea :+1:

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet: Why not. Identity uses `Add`, but `With` seems fine too.

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet, line 92: I don't think it's worth it. The check is trivial enough and different between the various methods, so I think it's fine here.

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet, line 231: Unfortunately, there's no way to retrieve an `ILogger` instance at this stage (this method is called from `ConfigureServices()`).

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet, line 427: Actually, `JWT` means `JSON Web Token`, so `JwtTokens` literally means `JSON Web Token Tokens` :smile: 
I'm considering renaming it in ASOS.

src/OpenIddict.Core/OpenIddictExtensions.cs
kevinchalet, line 94: Why not, but the name is optional and might be null here.

src/OpenIddict.Core/OpenIddictModule.cs
kevinchalet, line 17: Good idea, I'll add 2 constructors and remove the explicit property setters :+1:

src/OpenIddict.Core/Stores/IOpenIddictApplicationStore.cs
kevinchalet, line 17: No strong opinion, I used `Creates` because it's the terminology used by Identity.

src/OpenIddict.EntityFramework/Models/OpenIddictScope.cs
kevinchalet: Ah yeah, nice catch, thanks :+1:

src/OpenIddict.EntityFramework/Models/OpenIddictUser.cs
kevinchalet: Will remove it.

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
kevinchalet: Will make everything protected, thanks :smile:

src/OpenIddict.Mvc/OpenIddictController.cs
kevinchalet, line 24: Because you don't like the service locator pattern? :smile: 
I tried to get rid of it everywhere I could. `OpenIddictServices` is now only used in the provider itself (that's why it's in the `Infrastructure` namespace)

src/OpenIddict.Mvc/OpenIddictController.cs
kevinchalet, line 27: `applications` helps make the liner shorter :trollface:

src/OpenIddict.Mvc/OpenIddictController.cs
kevinchalet: Nice catch :+1:

src/OpenIddict.EntityFramework/Stores/OpenIddictTokenStore.cs
kevinchalet: Oh yeah, nice catch. Will fix that.

src/OpenIddict.EntityFramework/OpenIddictContext.cs
```diff
@@ -0,0 +1,102 @@
...
+                                 OpenIddictScope,
+                                 OpenIddictToken, IdentityRole, string>
+        where TUser : OpenIddictUser {
+        protected OpenIddictContext() { }
+
+        public OpenIddictContext(DbContextOptions options)
+            : base(options) { }
+    }
+
+    public class OpenIddictContext<TUser, TApplication, TAuthorization, TScope, TToken, TRole, TKey> : IdentityDbContext<TUser, TRole, TKey>
```
kevinchalet, line 59: I'll add some docs love :smile:

src/OpenIddict.Core/OpenIddictBuilder.cs
ilmax, line 427: :trollface:

src/OpenIddict.Mvc/OpenIddictController.cs
ilmax, line 24: nice!

kevinchalet:
Updated :smile:

ilmax:
LGTM

kevinchalet:
Merged.

Thanks for reviewing Massimiliano :clap:

Issue #112, opened 2016-05-14
Separating client_id and Application primary key


Ciantic:
I'm using a Integer as a keys for User, Role etc. Currently this also forces an application client_id to be a integer. This is not necessarily wanted, and they could be decoupled.

Right now I'm registering my official apps with e.g. with client_id 10000 and increment the number with another 10000 to make room for the possiblity of thirdparty apps (if there ever is any).

Ideally I would like to use string as a client_id, so I could register official apps with clear names such as "official-spa", "mobile-app" etc. so I could find them from logs if I need to.

kevinchalet:
Sounds like a good idea (as long as we can define the client_id column as an index for maximal performance).

/cc @Bartmax @ilmax thoughts? :smile:

Bartmax:
> Ideally I would like to use string as a client_id, so I could register official apps with clear names such as "official-spa", "mobile-app" etc. so I could find them from logs if I need to.

Agree, client_id type shouldn't be forced to be the same as user_id.

ilmax:
Ideally we could use a genircs here, so int or string can be used based on Dev needs

kevinchalet:
@ilmax the `Application` type supports custom key types thanks to its `TKey` generic argument (as requested by @khellang), but `TKey` is shared by all the OpenIddict/Identity models, so you can't use `string` for `Application` and `long` or `Guid` for the other models. We could of course use a key type per entity, but it would quickly become a real nightmare, as it would end up having an insane number of generic arguments (Identity doesn't support that for this exact reason).

I think having a separate indexed string property for `client_id` and a primary key whose type is chosen via `TKey` is a good compromise.

ilmax:
@PinpointTownes yes, that makes sense

kevinchalet:
Fixed, thanks to @xperiandri :clap:

PR #128, opened 2016-06-02
Separated ApplicationId and identity column of Application model

xperiandri:
Issue #112

Also eliminated dependency of Application primary key type from ASP.NET Identity primary key type.
Added consytaints of 38 characters for columns of Application model used for indexes.
Added XML documentation to modified types.
Updated sample Server project.


samples/Mvc.Server/Startup.cs
```diff
@@ -135,7 +135,7 @@ public class Startup {
                     var hasher = new PasswordHasher<Application>();
 
                     context.Applications.Add(new Application {
-                        Id = "myClient",
+                        ApplicationId = "myClient",
```
kevinchalet: Maybe it would be clearer to name it `ClientId`?

src/OpenIddict.EF/OpenIddictContext.cs
```diff
@@ -10,29 +10,120 @@
...
+    /// <summary>
+    /// Base class for the Entity Framework database context used for OpenIddict.
+    /// </summary>
+    /// <typeparam name="TApplication">The type of application objects.</typeparam>
+    /// <typeparam name="TApplicationKey">The type used for the primary key for the application.</typeparam>
+    /// <typeparam name="TUser">The type of user objects.</typeparam>
+    /// <typeparam name="TRole">The type of role objects.</typeparam>
+    /// <typeparam name="TKey">The type of the primary key for users and roles.</typeparam>
+    /// <seealso cref="Microsoft.AspNetCore.Identity.EntityFrameworkCore.IdentityDbContext{TUser, TRole, TKey}" />
+    public class OpenIddictContext<TApplication, TApplicationKey, TUser, TRole, TKey> : IdentityDbContext<TUser, TRole, TKey>
```
kevinchalet: I'm pretty sure we said "no" to a separate key type. Use TKey instead :smile:

src/OpenIddict.EF/OpenIddictExtensions.cs
```diff
@@ -74,6 +75,30 @@ where typeof(DbContext).IsAssignableFrom(argument)
             throw new InvalidOperationException("The type of the database context cannot be automatically inferred.");
         }
 
+        private static Type ResolveApplicationKeyType([NotNull] OpenIddictConfiguration configuration) {
```
kevinchalet: FYI, all this reflection sorcery is going to disappear with my experimental PR.

src/OpenIddict.Models/Application.cs
```diff
@@ -5,16 +5,82 @@
...
+    ///  which uses a string as a primary key.
+    /// </summary>
+    /// <seealso cref="Application{TKey}" />
+    /// <remarks>
+    /// An an <see cref="P:Application.ApplicationId"/> used <see cref="P:Application.Id"/>.
+    /// So <see cref="P:Application.ApplicationId"/> is not stored in database not to duplicate the same value.
+    /// </remarks>
+    public class Application : Application<string>
+    {
+        public override string ApplicationId
```
kevinchalet: Looks like rather unusual. Is this a pattern Identity uses somewhere? I'm concerned about introducing something totally new here.

kevinchalet:
Thanks for the PR! Minor a few remarks, it looks nice.

That said, it will directly conflict with my own changes. I'll try to merge my PR soon so we can progress with yours.

src/OpenIddict.EF/OpenIddictContext.cs
xperiandri: What if I want to use a `Guid` as a key for Identity and a `string` as an `ApplicationId` and a key at the same time? For me there is no reason to add additional column for primary key as `ApplicationId` is perfect option. But as long as `Application` primary key depends on Identity primary key I can't do that!
I implemented compromise variant. See additional `OpenIddictContext<TApplication, TUser, TRole, TKey>` implementation.

xperiandri:
OK
At least now I can build my current web application from my fork of sources and migrations work.
Because before I couldn't use Identity keys of `Guid` and ApplicationIds of `string`.

samples/Mvc.Server/Startup.cs
xperiandri: Then maybe we rename Application entity to Client entity? Or do you have some intent of having different names?

src/OpenIddict.EF/OpenIddictContext.cs
kevinchalet: > What if I want to use a Guid as a key for Identity and string as an ApplicationId and key at the same time?

You won't be able to do that, just like Identity doesn't allow you to use different key types for user/role. With the same logic, we'd have to add a `TAuthorizationKey`, a `TScopeKey`, a `TTokenKey` and honestly, it would be a real nightmare.

Don't forget that Identity added new generic arguments recently: https://github.com/aspnet/Identity/pull/798/files#diff-c72544a50eacab8ada8ca53f53daee15R84.

That would be great if we could avoid having 20 or 30 generic arguments :fearful:

samples/Mvc.Server/Startup.cs
kevinchalet: Having both `Id` and `ApplicationId` in the same `Application` entity is confusing. Renaming the second one would make clearer that `ClientId` is not a primary key.

kevinchalet:
> Because before I couldn't use Identity keys of Guid and ApplicationIds of string.

Curious: any particular reason not to use Guid for your client_id?

src/OpenIddict.EF/OpenIddictContext.cs
xperiandri: I understand you point however `Application` has nothing in common with Identity. Why must it depend on its generic parameter?
And your constraint breaks my primary scenario.

xperiandri: > TAuthorizationKey , a  TScopeKey , a  TTokenKey

I need some explanation. I haven't looked there.

kevinchalet: Similarly to what Identity does, the primary keys of all the OpenIddict entities are of type `TKey`.

kevinchalet: > And your constraint breaks my primary scenario.

It won't if we separate `Id` (the primary key, whose type is `TKey`) and `ApplicationId`/`ClientId`, whose type will necessarily be a string.

src/OpenIddict.Models/Application.cs
xperiandri: Actually they allow you to override `Id` on `IdentityUser` for example.

src/OpenIddict.EF/OpenIddictContext.cs
kevinchalet: > I need some explanation. I haven't looked there.

We are going to introduce new entities: `Authorization`, `Scope` and `Token`. If you want to introduce a key type specific to `Application`, then the same rule should apply to `Scope` and `Token`.

xperiandri: So you mean that you are going to have relationships with another entites from `Application` entity, right? And for that purpose you want to have the same key type on `Application` and that entities.

kevinchalet: No, I mean that if you want to have a key type specific to the `Application` entity, then someone else will want a key type for `Authorization`, another one for `Scope` and a last one for `Token`.

We can't afford having a key type for each entity.

xperiandri: What are `Authorization`, `Scope` and `Token`? I don't see such entities.
`Application` is a standalone entity with no relationships now. I don't see a relationship to what you say.

xperiandri: Are you going to save `Authorization`, `Scope` and `Token` to database and link them to `Application`?

kevinchalet: These entities are coming as part of my PR. E.g there's a direct relationship between `OpenIddictAuthorization` and `OpenIddictToken`:

``` csharp
/// <summary>
/// Represents an OpenIddict authorization.
/// </summary>
public class OpenIddictAuthorization<TToken, TKey> where TKey : IEquatable<TKey> {
    /// <summary>
    /// Gets or sets the unique identifier
    /// associated with the current authorization.
    /// </summary>
    public virtual TKey Id { get; set; }

    /// <summary>
    /// Gets or sets the space-delimited scopes
    /// associated with the current authorization.
    /// </summary>
    public virtual string Scope { get; set; }

    /// <summary>
    /// Gets or sets the list of tokens
    /// associated with the current authorization.
    /// </summary>
    public virtual IList<TToken> Tokens { get; } = new List<TToken>();

    /// <summary>
    /// Gets or sets the identifier of the user profile
    /// associated with the current authorization.
    /// </summary>
    public virtual TKey UserId { get; set; }
}
```

``` csharp
/// <summary>
/// Represents an OpenIddict token.
/// </summary>
public class OpenIddictToken<TKey> where TKey : IEquatable<TKey> {
    /// <summary>
    /// Gets or sets the identifier of the authorization attached with the current token.
    /// This property may be null if the token was issued without
    /// requiring the user consent or is bound to a client application.
    /// </summary>
    public virtual TKey AuthorizationId { get; set; }

    /// <summary>
    /// Gets or sets the unique identifier
    /// associated with the current token.
    /// </summary>
    public virtual TKey Id { get; set; }

    /// <summary>
    /// Gets or sets the type of the current token.
    /// </summary>
    public virtual string Type { get; set; }

    /// <summary>
    /// Gets or sets the identifier of the user attached with the current token.
    /// This property is null if the token represents a client application.
    /// </summary>
    public virtual TKey UserId { get; set; }
}
```

ilmax:
@xperiandri Are you still working on this?

xperiandri:
Will do an update based on latest changes and remarks next week

ilmax:
Nice!

src/OpenIddict.EntityFramework/Models/OpenIddictApplication.cs
```diff
@@ -19,54 +20,57 @@ public class OpenIddictApplication : OpenIddictApplication<string, OpenIddictTok
...
         /// </summary>
         public virtual string DisplayName { get; set; }
 
         /// <summary>
-        /// Gets or sets the unique identifier
-        /// associated with the current application.
+        /// Gets or sets the unique identifier for this application.
         /// </summary>
-        public virtual TKey Id { get; set; }
+        public virtual string ApplicationId { get; set; }
```
ilmax: I prefer call this `ClientId`

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
```diff
@@ -60,19 +60,19 @@ public class OpenIddictApplicationStore<TApplication, TToken, TContext, TKey> :
...
         /// </summary>
-        /// <param name="identifier">The unique identifier associated with the application.</param>
+        /// <param name="id">The ID of application to search for.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
         /// <returns>
         /// A <see cref="Task"/> that can be used to monitor the asynchronous operation,
         /// whose result returns the client application corresponding to the identifier.
         /// </returns>
-        public virtual Task<TApplication> FindByIdAsync(string identifier, CancellationToken cancellationToken) {
+        public virtual Task<TApplication> FindByIdAsync(string id, CancellationToken cancellationToken) {
```
ilmax, line 75: I'm pretty sure this method now need to search an application by `ApplicationId` and not by `Id`

xperiandri, line 75: Added one more method for that purpose.
Kept this one for style compatibility with Identity reason.

src/OpenIddict.EntityFramework/Models/OpenIddictApplication.cs
xperiandri: What is the idea behind this?
To correspond to `client_id` query parameter of the URL?

src/OpenIddict.EntityFramework/Models/OpenIddictApplication.cs
```diff
@@ -19,54 +20,57 @@ public class OpenIddictApplication : OpenIddictApplication<string, OpenIddictTok
...
-        public virtual string LogoutRedirectUri { get; set; }
+        /// <remarks>Use hash of the secret for security purposes.</remarks>
+        public virtual string Secret { get; set; }
 
         /// <summary>
-        /// Gets or sets the callback URL
-        /// associated with the current application.
+        /// Gets or sets the callback URL for this application.
         /// </summary>
         public virtual string RedirectUri { get; set; }
```
xperiandri, line 56: Maybe naming this as LoginRedirectUri would be more consistent?

src/OpenIddict.EntityFramework/Models/OpenIddictApplication.cs
ilmax: This is the client identifier called `client_id` in the specs

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
ilmax, line 75: Ok, but all the callers should be updated to call the new method (we should lookup a client by `client_id` and not by PK)

kevinchalet:
Merged, thanks :+1: 

https://github.com/openiddict/openiddict-core/commit/4a3b84ddb5a9a0a126ed9427dfd8e071a8b3de81

ilmax:
:tada:

PR #142, opened 2016-06-24
Updated model building

xperiandri:
Overridden names of ASP.NET Identity tables to distinct from other tables and be in the same group as OpenIddict.
Set ClientId as alternate key as it is a unique property that can act as key.
Added name for index on ClientId property.
Set ClientId, RedirectUri and Type properties as required because client application definition cannot be created without this data.

### DIFF
--- a/src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
+++ b/src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
@@ -150,62 +150,77 @@ namespace OpenIddict {
         protected override void OnModelCreating(ModelBuilder builder) {
             base.OnModelCreating(builder);
 
+            // Rename Identity tables.
+            builder.Entity<TUser>(e => e.ToTable("OpenIddictUsers"));
+            builder.Entity<TRole>(e => e.ToTable("OpenIddictRoles"));
+            builder.Entity<IdentityUserClaim<TKey>>(e => e.ToTable("OpenIddictUserClaims"));
+            builder.Entity<IdentityRoleClaim<TKey>>(e => e.ToTable("OpenIddictRoleClaims"));
+            builder.Entity<IdentityUserRole<TKey>>(e => e.ToTable("OpenIddictUserRoles"));
+            builder.Entity<IdentityUserLogin<TKey>>(e => e.ToTable("OpenIddictUserLogins"));
+            builder.Entity<IdentityUserToken<TKey>>(e => e.ToTable("OpenIddictUserTokens"));
+
             // Warning: optional foreign keys MUST NOT be added as CLR properties because
             // Entity Framework would throw an exception due to the TKey generic parameter
             // being non-nullable when using value types like short, int, long or Guid.
 
             // Configure the TApplication entity.
-            builder.Entity<TApplication>(entity => {
-                entity.HasKey(application => application.Id);
+            builder.Entity<TApplication>(e => {
+                e.HasKey(app => app.Id);
 
-                entity.HasIndex(application => application.ClientId)
-                      .IsUnique(unique: true);
+                e.HasAlternateKey(app => app.ClientId);
+                e.HasIndex(app => app.ClientId).HasName("ApplicationClientIdIndex");
 
-                entity.HasMany(application => application.Tokens)
-                      .WithOne()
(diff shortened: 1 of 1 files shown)


kevinchalet:
> Overridden names of ASP.NET Identity tables to distinct from other tables and be in the same group as OpenIddict.

Not a huge fan of that, personally. Specially since there's no point doing that.

> Set ClientId as alternate key as it is a unique property that can act as key.

Pretty sure it has consequences we don't want (e.g not being able to update the client_id value)

ilmax:
No strong opinion on table renaming, we might leave it as is, agreed on alternate key.

src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
```diff
@@ -150,63 +150,78 @@ public class OpenIddictDbContext<TUser, TRole, TApplication, TAuthorization, TSc
         protected override void OnModelCreating(ModelBuilder builder) {
             base.OnModelCreating(builder);
 
+            // Rename Identity tables.
+            builder.Entity<TUser>(e => e.ToTable("OpenIddictUsers"));
```
kevinchalet, line 154: Please revert.

src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
```diff
@@ -150,63 +150,78 @@ public class OpenIddictDbContext<TUser, TRole, TApplication, TAuthorization, TSc
...
+
             // Warning: optional foreign keys MUST NOT be added as CLR properties because
             // Entity Framework would throw an exception due to the TKey generic parameter
             // being non-nullable when using value types like short, int, long or Guid.
 
             // Configure the TApplication entity.
-            builder.Entity<TApplication>(entity => {
-                entity.HasKey(application => application.Id);
+            builder.Entity<TApplication>(e => {
+                e.HasKey(app => app.Id);
```
kevinchalet, line 168: When you contribute to a project, you MUST conform to its coding style and avoid changing things that don't need to be changed. Revert that.

src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
```diff
@@ -150,63 +150,78 @@ public class OpenIddictDbContext<TUser, TRole, TApplication, TAuthorization, TSc
...
+
+                e.HasAlternateKey(app => app.ClientId);
+                e.HasIndex(app => app.ClientId).HasName("ApplicationClientIdIndex");
```
kevinchalet, line 171: Pretty sure we had concluded an explicit index name was not necessary. Revert that.

src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
```diff
@@ -150,63 +150,78 @@ public class OpenIddictDbContext<TUser, TRole, TApplication, TAuthorization, TSc
...
             // Entity Framework would throw an exception due to the TKey generic parameter
             // being non-nullable when using value types like short, int, long or Guid.
 
             // Configure the TApplication entity.
-            builder.Entity<TApplication>(entity => {
-                entity.HasKey(application => application.Id);
+            builder.Entity<TApplication>(e => {
+                e.HasKey(app => app.Id);
+
+                e.HasAlternateKey(app => app.ClientId);
```
kevinchalet, line 170: Remove that too, please.

src/OpenIddict.EntityFramework/OpenIddictDbContext.cs
```diff
@@ -150,63 +150,78 @@ public class OpenIddictDbContext<TUser, TRole, TApplication, TAuthorization, TSc
...
-                entity.HasMany(user => user.Authorizations)
-                      .WithOne()
-                      .HasForeignKey("UserId")
-                      .IsRequired(required: false);
-
-                entity.HasMany(user => user.Tokens)
-                      .WithOne()
-                      .HasForeignKey("UserId")
-                      .IsRequired(required: false);
+            builder.Entity<TUser>(e => {
```
kevinchalet, line 214: Same remark, revert these cosmetic changes.

xperiandri:
Made a separate one

PR #144, opened 2016-06-27
Introduce alternate key for Application.ClientId property

xperiandri:
Added alternate key for Application.ClientId property which automatically creates index for it

Produced SQL is
CONSTRAINT [AK_OpenIddictApplications_ClientId] UNIQUE NONCLUSTERED ([ClientId] ASC)


kevinchalet:
https://github.com/openiddict/openiddict-core/pull/142#issuecomment-228562284

kevinchalet:
As mentioned in the linked comment, we had concluded it was not a good idea.

xperiandri:
@ilmax said that he agrees to this suggestion, don't you?

Issue #145, opened 2016-06-28
Error after upgrade to Asp.net core 1.0.0. 


ghost:
Hi there & thanks for your amazing work :)

I ran into an issue after upgrading asp.net core rc2 -> 1.0.0. yesterday:

When loading my dev-app, the line
`services.AddOpenIddict<Core.Entities.User, DataAccess.ApoNextDbContext>).DisableHttpsRequirement();
`
causes 

> An exception of type 'System.TypeLoadException' occurred in OpenIddict.dll but was not handled in user code
> 
> Additional information: Der Typ "Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionExtensions" in der Assembly "Microsoft.Extensions.DependencyInjection.Abstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60" konnte nicht geladen werden.
> 
> (translates as 'Type ... in assebly .. could not be loaded')

I am using Openiddict 1.0.0-alpha2-0331 .   Everything was running fine with asp.net core RC2 & OpenIddict alpha 302.

Let me know if you need additional information!

Mo

kevinchalet:
> Hi there & thanks for your amazing work :)

Haha, thanks :smile: 

The issue you're seeing is unfortunately a known one, caused by class name changes in the logging stack (and in a few other places): https://github.com/aspnet/DependencyInjection/issues/411

This incompatibility will be fixed when migrating to the RTM bits. In the meantime, I'd recommend using the ASP.NET Core RC2 packages :clap:

ghost:
Alright! Thanks for your quick response. Keep up the good work!

smbecker:
Not to push but any idea when a new package that targets the RTM bits will be available? I can't upgrade to RTM until it is so just want to plan accordingly.

kevinchalet:
> Not to push but any idea when a new package that targets the RTM bits will be available?

In the next few days. Actually, I already have a working "OpenIddict for ASP.NET Core RTM" prototype, but I'd like to share it with a few testers before releasing it publicly.

kevinchalet:
For those who want to try the experimental "OpenIddict for RTM" bits, you can download the corresponding packages here: https://ci.appveyor.com/project/openiddict/openiddict-core/build/1.0.0-alpha2-337/artifacts

ghost:
Thanks for the upload!

I get a new error now on Startup

`var host = new WebHostBuilder()
                .UseKestrel()
[...]

Issue #146, opened 2016-06-28
Update CryptoHelper to v2.0


henkmollema:
I released [CryptoHelper v2.0](https://www.nuget.org/packages/CryptoHelper/2.0.0) which targets .NET Core RTM. Either I'll send a pull request when you are updating to RTM or you can update yourself. :smile:

kevinchalet:
Ha great, thanks, I'll include that in my local branch! :tada: 

Curious: is the <net451 TFM deprecation the only breaking change?

henkmollema:
Yeah, I removed the `net40` and `net45` targets and the associated API's which were `ifdef`'d. The API for `net451` and `.netstandard1.3` stays the same.

kevinchalet:
Fixed as part of https://github.com/openiddict/openiddict-core/commit/dced1cd8fee9aa3c8ef9ff9a4271f204bb15f66c.

@henkmollema thanks! :clap:

PR #147, opened 2016-06-29
Migrate to the .NET Core/ASP.NET Core RTM packages

kevinchalet:
Closes https://github.com/openiddict/openiddict-core/issues/145 and https://github.com/openiddict/openiddict-core/issues/146.

Note: the final ASOS beta6 bits are not on NuGet.org yet because I'm still waiting for some feedback from the beta testers who're experimenting it in a real world app' before publishing it, so we'll temporarily use the latest beta6 nightly builds.


src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -13,16 +13,113 @@
...
 using Microsoft.Extensions.Logging;
 
 namespace OpenIddict.Infrastructure {
     public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TScope, TToken> : OpenIdConnectServerProvider
         where TUser : class where TApplication : class where TAuthorization : class where TScope : class where TToken : class {
+        public override async Task ExtractAuthorizationRequest([NotNull] ExtractAuthorizationRequestContext context) {
+            var services = context.HttpContext.RequestServices.GetRequiredService<OpenIddictServices<TUser, TApplication, TAuthorization, TScope, TToken>>();
+
+            // Reject requests using the unsupported request parameter.
+            if (!string.IsNullOrEmpty(context.Request.GetParameter(OpenIdConnectConstants.Parameters.Request))) {
```
kevinchalet, line 31: These 2 checks used to be part of ASOS but were removed in beta6 to allow implementers to support `request` and `request_uri` (http://openid.net/specs/openid-connect-core-1_0.html#RequestUriParameter).

Since OpenIddict doesn't implement it, these checks are added to reject requests that use one of these parameters.

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -13,16 +13,113 @@
...
+                context.Reject(
+                    error: OpenIdConnectConstants.Errors.RequestUriNotSupported,
+                    description: "The request_uri parameter is not supported.");
+
+                return;
+            }
+
+            // If a request_id parameter can be found in the authorization request,
+            // restore the complete authorization request stored in the user session.
+            if (!string.IsNullOrEmpty(context.Request.GetRequestId())) {
```
kevinchalet, line 56: 2 words about this (major) change:

ASOS used to store the authorization requests in a distributed cache to allow flowing large POST requests without interfering with the external authentication flow (e.g when logging in via Google). As of beta6, ASOS is now completely stateless and implementations must handle that part if they want to support POST requests.

The suggested implementation currently uses sessions, which requires adding the session middleware in your app if you want to support POST requests. If you think it's better to avoid using sessions, I'm fine using the distributed cache as-is.

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -13,16 +13,113 @@
...
+
+                // Restore the authorization request parameters from the serialized payload.
+                context.Request.Import(payload);
+            }
+        }
+
         public override async Task ValidateAuthorizationRequest([NotNull] ValidateAuthorizationRequestContext context) {
             var services = context.HttpContext.RequestServices.GetRequiredService<OpenIddictServices<TUser, TApplication, TAuthorization, TScope, TToken>>();
 
+            // Note: the OpenID Connect server middleware supports authorization code, implicit, hybrid,
```
kevinchalet, line 93: FYI, ASOS no longer rejects requests that use non-standard `response_type`/`response_mode` to allow you to implement custom flows (similarly to what's allowed for `grant_type`). Since we don't want to support non-standard flows in OpenIddict, I've copied these checks from ASOS to ensure we reject them properly.

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Exchange.cs
```diff
@@ -126,236 +126,283 @@ public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TSc
             context.Validate();
         }
 
-        public override async Task GrantClientCredentials([NotNull] GrantClientCredentialsContext context) {
+        public override async Task HandleTokenRequest([NotNull] HandleTokenRequestContext context) {
```
kevinchalet, line 129: Please ignore the changes made to this file, as there's no functional changes (only formatting changes caused by the `GrantAuthorizationCode`/`GrantRefreshToken`/`GrantClientCredentials`/`GrantResourceOwnerCredentials` -> `HandleTokenRequest` merge).

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Serialization.cs
```diff
@@ -17,14 +17,41 @@
 namespace OpenIddict.Infrastructure {
     public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TScope, TToken> : OpenIdConnectServerProvider
         where TUser : class where TApplication : class where TAuthorization : class where TScope : class where TToken : class {
-        public override async Task SerializeRefreshToken([NotNull] SerializeRefreshTokenContext context) {
+        public override async Task SerializeAuthorizationCode([NotNull] SerializeAuthorizationCodeContext context) {
```
kevinchalet, line 20: ASOS <beta6 used to support automatic authorization code revocation by storing codes in a distributed cache and returning a corresponding cache key. This is no longer true with beta6 (due to the completely stateless approach) so OpenIddict is now responsible of implementing authorization code revocation to prevent code reuse.

Since we already have the infrastructure we need for that (`OpenIddictTokenManager`), we're just leveraging it here.

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -431,7 +432,7 @@ public class OpenIddictBuilder {
         /// </summary>
         /// <returns>The <see cref="OpenIddictBuilder"/>.</returns>
         public virtual OpenIddictBuilder UseJsonWebTokens() {
-            return Configure(options => options.UseJwtTokens());
+            return Configure(options => options.AccessTokenHandler = new JwtSecurityTokenHandler());
```
kevinchalet, line 435: @ilmax FYI, ASOS beta6 no longer has a `UseJwtTokens()` extension to encourage people to use the opaque format (yeah, I'm evil :smile:)

src/OpenIddict.Mvc/OpenIddictController.cs
```diff
@@ -41,20 +39,7 @@ public class OpenIddictController<TUser, TApplication, TAuthorization, TToken> :
                 });
             }
 
-            // Note: authentication could be theorically enforced at the filter level via AuthorizeAttribute
```
kevinchalet, line 44: Note: this part is no longer necessary, as it's automatically done by `OpenIddictProvider.HandleAuthorizationRequest`: when receiving a POST request, we store it in the user session and generate a unique identifier. We then redirect the user to the same page but with the `request_id` appended in the query string, which allows us to use the classic `AuthorizeAttribute` logic.

kevinchalet:
Looks like NWebsec was updated to target ASP.NET Core RTM. I'll update the reference :clap:

src/OpenIddict.Core/OpenIddictBuilder.cs
ilmax, line 435: Hahah that'evil :) so no more JWT for access token right?

ilmax:
Will a deep look tomorrow night!

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet, line 435: JWT is still supported for access tokens, but you have to use the long form: `options.AccessTokenHandler = new JwtSecurityTokenHandler()`, which is exactly what OpenIddict's extension does.

ilmax, line 435: Ok, will look into opaque tokens!

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Exchange.cs
```diff
@@ -126,236 +126,283 @@ public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TSc
...
 
-            // Retrieve the application details corresponding to the requested client_id.
-            // Note: this call shouldn't return a null instance, but a race condition may occur
-            // if the application was removed after the initial check made by ValidateTokenRequest.
-            var application = await services.Applications.FindByClientIdAsync(context.ClientId);
-            if (application == null) {
-                throw new InvalidOperationException("The token request was aborted because the client application corresponding " +
-                                                   $"to the '{context.ClientId}' identifier was not found in the database.");
-            }
+            Debug.Assert(context.Request.IsAuthorizationCodeGrantType() ||
```
ilmax, line 132: Does it make sense to log something if the `grant_type` value is not supported? I.e. remove the Debug.Assert in favor of a proper check?

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Serialization.cs
```diff
@@ -17,14 +17,41 @@
...
 
-            Debug.Assert(context.Request.RequestType == OpenIdConnectRequestType.TokenRequest,
-                "The request should be a token request.");
+            Debug.Assert(context.Request.RequestType == OpenIdConnectRequestType.Authentication, "The request should be an authorization request.");
+            Debug.Assert(!string.IsNullOrEmpty(context.Request.ClientId), "The client_id parameter shouldn't be null or empty.");
+
+            // Note: a null value could be returned by FindByIdAsync. In this case, throw an exception to abort the token request.
+            var user = await services.Users.FindByIdAsync(context.Ticket.Principal.GetClaim(ClaimTypes.NameIdentifier));
+            if (user == null) {
+                throw new InvalidOperationException("The token request was aborted because the user associated " +
```
ilmax, line 29: Does those exception get's logged? If not can we consider to add logging before trowing?

src/OpenIddict.Mvc/OpenIddictExtensions.cs
```diff
@@ -154,6 +154,26 @@ public static class OpenIddictExtensions {
                     return provider.GetRequiredService(typeof(OpenIddictUserManager<>).MakeGenericType(builder.UserType));
                 });
 
+                // Register the OpenIddict services in the isolated container.
+                services.AddScoped(typeof(OpenIddictServices<,,,,>).MakeGenericType(
```
ilmax, line 164: nit: we are making a slow reflection call `typeof(OpenIddictServices<,,,,>).MakeGenericType(....)` twice, this type can be saved and reused in the `container.GetRequiredService(typeof(OpenIddictServices<,,,,>).MakeGenericType(...)`

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
ilmax, line 56: I'm fine with this new dependency now, we should make it clear in the readme that we have to register the session middelware. If distributed cache will be requested in the future, we can add an abstraction here to allow customizing the request storage

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -213,7 +316,56 @@ public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TSc
...
+
+                // Generate a request identifier. Note: using a crypto-secure
+                // random number generator is not necessary in this case.
+                var identifier = Guid.NewGuid().ToString();
+
+                // Load the session from the session store.
+                await context.HttpContext.Session.LoadAsync();
+
+                // Store the serialized authorization request parameters in the user session.
+                context.HttpContext.Session.Set("openiddict-request:" + identifier, context.Request.Export());
```
ilmax: nit: Can we introduce a const here?

ilmax:
few minor nits but the changes LGTM :+1:

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Exchange.cs
kevinchalet, line 132: The `grant_type` parameter is validated from `ValidateTokenRequest`: https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Exchange.cs#L25-L38

The `Debug.Assert` is just here to ensure we don't update `ValidateTokenRequest` to add new grant types without implementing them in `HandleTokenRequest` :smile:

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Serialization.cs
kevinchalet, line 29: They are ultimately caught and logged by the host (e.g Kestrel), so not sure it's worth logging the exception message before throwing.

src/OpenIddict.Mvc/OpenIddictExtensions.cs
kevinchalet, line 164: Ideally, I'd like to obsolete `OpenIddict.Mvc` :smile:

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
kevinchalet: Do you have a specific name in mind?

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
kevinchalet, line 56: Note that, by default, the session middleware relies on the distributed cache (that can be in-memory, SQL or Redis). AFAIK, there's no built-in session store that doesn't rely on the caching stack.

The question here is do we want to depend on sessions or directly on the caching stuff?

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Serialization.cs
ilmax, line 29: I didn't think to log the exception but rather the cause of the exception being thrown to have a more meaningful error message (as we already do in other parts)

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
ilmax: `OpenidDictRequestKey = "openiddict-request:"` ? I'm not a huge fan of strings :)

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Serialization.cs
kevinchalet, line 29: Well, we could. But seeing this exception should be extremely rare, as this method is never called if the user entity cannot be validated when receiving the authorization request. So it's only thrown if the user is removed after the initial check (which is basically a race condition).

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
ilmax, line 56: Ok, so I'm fine using session.

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Serialization.cs
ilmax, line 29: Ok, so I'm fine with this

src/OpenIddict.Mvc/OpenIddictExtensions.cs
```diff
@@ -169,30 +195,64 @@ public static class OpenIddictExtensions {
...
+                    // to remove the ending markers added to the generic type name.
+                    controller.ControllerName = "OpenIddict";
+                }
             }
         }
 
-        private class OpenIddictPart : ApplicationPart, IApplicationPartTypeProvider {
-            public OpenIddictPart(OpenIddictBuilder builder) {
-                Types = new[] {
+        private class OpenIddictPart : ApplicationPart, IApplicationPartTypeProvider, ICompilationReferencesProvider {
```
kevinchalet, line 208: Not a huge fan of all this sorcery, but it's pretty much the only way to work around the bug reported by Josh.

kevinchalet:
@ilmax updated :smile:

Issue #150, opened 2016-07-01
Error with OpenIddict.Mvc


papaytl185:
I am on the newest .netcore 1.0.0
When I add "OpenIddict.Mvc": "1.0.0-*" I get the below error when running

**InvalidOperationException: Can not find compilation library location for package 'OpenIddict.Mvc'**

```
stack:
System.InvalidOperationException: Can not find compilation library location for package 'OpenIddict.Mvc'
   at Microsoft.Extensions.DependencyModel.CompilationLibrary.ResolveReferencePaths()
   at System.Linq.Enumerable.<SelectManyIterator>d__157`2.MoveNext()
   at Microsoft.AspNetCore.Mvc.Razor.Compilation.MetadataReferenceFeatureProvider.PopulateFeature(IEnumerable`1 parts, MetadataReferenceFeature feature)
   at Microsoft.AspNetCore.Mvc.ApplicationParts.ApplicationPartManager.PopulateFeature[TFeature](TFeature feature)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.DefaultRoslynCompilationService.GetCompilationReferences()
   at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.DefaultRoslynCompilationService.get_CompilationReferences()
   at Microsoft.AspNetCore.Mvc.Razor.Internal.DefaultRoslynCompilationService.Compile(RelativeFileInfo fileInfo, String compilationContent)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.RazorCompilationService.Compile(RelativeFileInfo file)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.CompilerCache.CreateCacheEntry(String relativePath, String normalizedPath, Func`2 compile)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.CompilerCache.GetOrAdd(String relativePath, Func`2 compile)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.DefaultRazorPageFactoryProvider.CreateFactory(String relativePath)
[...]

PR #152, opened 2016-07-02
Remove OpenIddict.Mvc, OpenIddict.Assets and OpenIddict.Security

kevinchalet:
ASP.NET Core MVC RTM broke our MVC module: https://github.com/openiddict/openiddict-core/issues/150. Since there's no viable fix, the only reasonable approach is to remove `OpenIddict.Mvc` (and `OpenIddict.Assets`) and move to a "template" model, similar to the one used for `AccountController`.

(this PR is a WIP, please don't merge it yet).

/cc @ilmax @Bartmax


samples/Mvc.Server/Controllers/AuthorizationController.cs
```diff
@@ -0,0 +1,183 @@
...
+            SignInManager<ApplicationUser> signInManager,
+            OpenIddictUserManager<ApplicationUser> userManager) {
+            _applicationManager = applicationManager;
+            _signInManager = signInManager;
+            _userManager = userManager;
+        }
+
+        [Authorize, HttpGet, HttpPost, Route("~/connect/authorize")]
+        public async Task<IActionResult> Authorize() {
+            // Note: when a fatal error occurs during the request processing, an OIDC response
```
kevinchalet: @ilmax now that developers are responsible of creating their own controller, I believe we should handle "errored requests" directly in OpenIddict as there's a risk they won't handle them properly in their code.

I have 3 approaches in mind:
- Display a native error page, embedded in OpenIddict.Core. This simple approach has a downside: it's not very flexible and doesn't allow customizing it.
- Throw an exception and let the developer handle it using the error page middleware. I'm not a huge fan of throwing exceptions in this case, but it's simple and flexible.
- Rewrite the request path to something like /connect/error and invoke the rest of the pipeline. This allows creating a corresponding "error" action/controller to display the error code.

Any thoughts?

Bartmax:
what about return bad request ? is that possible?

kevinchalet:
> what about return bad request ? is that possible?

You mean just an empty 400 response? Sure it's possible, but it's not really user-friendly. People will likely want to customize it.

Bartmax:
Bad request can have a response body with the error.

kevinchalet:
Sure. That was even my first option :smile: 

> Display a native error page, embedded in OpenIddict.Core. This simple approach has a downside: it's not very flexible and doesn't allow customizing it.

Bartmax:
well that would be my choice without a "native view" just plain bad request error and text. Another thing that may be interesting is if we can plug/use the developmenterrorsmiddleware somehow.

ilmax:
Code changes LGTM, we might make it explicit we no longer provide a default UI and maybe point to the samples repo to find a basic implementation.

samples/Mvc.Server/Controllers/AuthorizationController.cs
ilmax: IMO 3 Can be the most flexible way to handle this so we can leverage the `ErrorHandlingPath` introduced in this PR

kevinchalet: Yeah, that's the approach I tried in the latest commits.

I'll also give the status code middleware a try before merging this PR. It might give us more flexibility.

kevinchalet:
Note for me: update README.md before merging.

kevinchalet:
Merged :clap: 

@ilmax @Bartmax thanks for your review :+1:

PR #157, opened 2016-07-07
Replace the internal error handling logic by the status code pages stack

kevinchalet:
This PR replaces the error handling logic I introduced recently in OpenIddict by a smarter one that relies on the status code pages stack:
- If the status code pages middleware is registered AND if the feature is not disabled, then OpenIddict will stop handling the request and avoid writing to the response stream. The status code pages middleware will then rewrite the response according to the policy chosen by the developer (e.g re-execute the request)
- If it's not registered or if it's disabled, ASOS will render a default error page (a plain text one :smile:)

/cc @ilmax @Bartmax


Bartmax:
I LOVE this change!! :+1:

kevinchalet:
Note that this new approach is not functionally different from the previous one, it just better integrates with the rest of ASP.NET Core instead of having an OpenIddict-specific way to handle errored requests :smile: 

I still expect most people to use `app.UseStatusCodePagesWithReExecute`, which will re-execute the request with a different path, like what I had implemented in my previous PR.

Bartmax:
I know!

> it just better integrates with the rest of ASP.NET Core

that is the core objective of this library. Hence why I love this kind of changes. ;)

kevinchalet:
@ilmax please give me the squirrel if you approve this change :shipit: :smile:

PR #170, opened 2016-07-20
Throw an exception when no signing key is explicitly registered and introduce OpenIddictBuilder.AddEphemeralSigningKey()

kevinchalet:
Automatic RSA key generation is likely going to be removed in ASOS beta7 (https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/pull/318).

OpenIddict uses beta6-final so this change can theoretically wait, but I want to make sure it's introduced soon enough to get some feedback for ASOS beta7-final.

I'm 100% aware this change will make ASOS/OpenIddict a bit more painful to configure so to compensate, both will get a new `AddEphemeralSigningKey()` helper generating an in-memory RSA key during development.

/cc @ilmax

### DIFF
--- a/src/OpenIddict.Core/OpenIddictBuilder.cs
+++ b/src/OpenIddict.Core/OpenIddictBuilder.cs
@@ -10,11 +10,13 @@ using System.IdentityModel.Tokens.Jwt;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using AspNet.Security.OpenIdConnect.Extensions;
 using JetBrains.Annotations;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.IdentityModel.Tokens;
 using OpenIddict;
 
 namespace Microsoft.AspNetCore.Builder {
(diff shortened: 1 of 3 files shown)


src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -277,6 +279,59 @@ public class OpenIddictBuilder {
...
+                algorithm.Dispose();
+                algorithm = new RSACryptoServiceProvider(2048);
+            }
+#endif
+
+            if (algorithm.KeySize < 2048) {
+                throw new InvalidOperationException("The ephemeral key generation failed.");
+            }
+
+            // Note: the RSA instance cannot be flowed as-is due to a bug in IdentityModel that disposes
```
kevinchalet, line 311: AFAIK, this bug is not known, I'll open a new ticket and ping someone from the Azure AD team, in charge of the IdentityModel stack.

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -277,6 +279,59 @@ public class OpenIddictBuilder {
         }
 
         /// <summary>
+        /// Registers a new ephemeral key used to sign the tokens issued by OpenIddict: the key
+        /// is discarded when the application shuts down and tokens signed using this key are
+        /// automatically invalidated. This method should only be used during development:
+        /// on production, using a X.509 certificate stored in the machine store is recommended.
+        /// </summary>
+        /// <returns>The <see cref="OpenIddictBuilder"/>.</returns>
+        public virtual OpenIddictBuilder AddEphemeralSigningKey() {
```
kevinchalet, line 288: FYI, this helper is identical to ASOS' `AddEphemeralKey()` helper but we can't use it as it's not part of beta6. We'll migrate to the "official" helper when moving to beta7.

ilmax:
Sorry for being late at the party, this is a LGTM, we should mention that in the readme

kevinchalet:
Good idea!

I added `AddEphemeralSigningKey()` in the 2 snippets we have in README.md: https://github.com/openiddict/openiddict-core/commit/a94ec16a1b5c2a433fe0e7811b46d46305b145a2

ilmax:
nice! Thank you

src/OpenIddict.Core/OpenIddictBuilder.cs
kevinchalet, line 311: Ticket opened: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/477

PR #178, opened 2016-07-30
Update OpenIddict.Core to target the ASOS beta7 nightly builds

kevinchalet:
Related ticket: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/pull/319.

/cc @ilmax

### DIFF
--- a/samples/Mvc.Server/Controllers/AuthorizationController.cs
+++ b/samples/Mvc.Server/Controllers/AuthorizationController.cs
@@ -5,6 +5,7 @@
  */
 
 using System;
+using System.Linq;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
@@ -52,7 +53,9 @@ namespace Mvc.Server {
 
             return View(new AuthorizeViewModel {
                 ApplicationName = application.DisplayName,
-                Parameters = request.Parameters,
+                Parameters = request.ToDictionary(
+                    parameter => parameter.Key,
+                    parameter => (string) parameter.Value),
                 Scope = request.Scope
             });
         }
@@ -101,7 +104,9 @@ namespace Mvc.Server {
             var request = HttpContext.GetOpenIdConnectRequest();
 
             return View(new LogoutViewModel {
-                Parameters = request.Parameters
+                Parameters = request.ToDictionary(
+                    parameter => parameter.Key,
+                    parameter => (string) parameter.Value)
             });
         }
 
--- a/samples/Mvc.Server/Controllers/AuthorizationController.cs
+++ b/samples/Mvc.Server/Controllers/AuthorizationController.cs
@@ -5,7 +5,6 @@
  */
 
 using System;
-using System.Linq;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
@@ -51,11 +50,11 @@ namespace Mvc.Server {
                 });
             }
 
+            // Flow the request_id to allow OpenIddict to restore
+            // the original authorization request from the cache.
             return View(new AuthorizeViewModel {
                 ApplicationName = application.DisplayName,
-                Parameters = request.ToDictionary(
-                    parameter => parameter.Key,
-                    parameter => (string) parameter.Value),
+                RequestId = request.RequestId,
                 Scope = request.Scope
             });
         }
@@ -103,10 +102,10 @@ namespace Mvc.Server {
             // Extract the authorization request from the ASP.NET environment.
             var request = HttpContext.GetOpenIdConnectRequest();
 
+            // Flow the request_id to allow OpenIddict to restore
+            // the original logout request from the distributed cache.
             return View(new LogoutViewModel {
-                Parameters = request.ToDictionary(
-                    parameter => parameter.Key,
-                    parameter => (string) parameter.Value)
+                RequestId = request.RequestId
             });
         }
 
--- a/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
+++ b/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
@@ -20,6 +21,8 @@ using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Bson;
 
 namespace OpenIddict.Infrastructure {
     public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TScope, TToken> : OpenIdConnectServerProvider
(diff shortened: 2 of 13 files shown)


src/OpenIddict.Core/OpenIddictOptions.cs
```diff
@@ -16,9 +16,6 @@ namespace OpenIddict {
     /// </summary>
     public class OpenIddictOptions : OpenIdConnectServerOptions {
         public OpenIddictOptions() {
-            AuthorizationEndpointPath = IntrospectionEndpointPath = LogoutEndpointPath =
```
kevinchalet, line 19: FYI, "nullifying" the endpoint paths is no longer necessary with ASOS beta7 because all the endpoints are disabled by default.

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -378,27 +396,23 @@ public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TSc
...
+                // If it was not registered or enabled, let the OpenID Connect server middleware render
                 // a default error page instead of delegating the rendering to the status code middleware.
                 var feature = context.HttpContext.Features.Get<IStatusCodePagesFeature>();
                 if (feature != null && feature.Enabled) {
                     // Replace the default status code to return a 400 response.
                     context.HttpContext.Response.StatusCode = 400;
 
-                    // Store the OpenID Connect response in the HTTP context to allow retrieving it
-                    // from user code (e.g from an ASP.NET Core MVC controller or a Nancy module).
-                    context.HttpContext.SetOpenIdConnectResponse(context.Response);
```
kevinchalet, line 400: FYI, this is something ASOS beta7 now does automatically before invoking the `ApplyAuthorizationResponse` event.

src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -67,7 +70,14 @@ public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TSc
                 }
 
                 // Restore the authorization request parameters from the serialized payload.
-                context.Request.Import(payload);
```
kevinchalet, line 70: FYI, the removed `Import`/`Export` helpers used binary serialization (via `BinaryReader`/`BinaryWriter`). To support complex serialization, we're going to move to BSON.

samples/Mvc.Server/Controllers/AuthorizationController.cs
```diff
@@ -52,7 +53,9 @@ public class AuthorizationController : Controller {
 
             return View(new AuthorizeViewModel {
                 ApplicationName = application.DisplayName,
-                Parameters = request.Parameters,
+                Parameters = request.ToDictionary(
```
kevinchalet: @ilmax I'm considering simply flowing the `request_id` instead of the entire OIDC payload, which would make views and view models a bit simpler and would prevent devs from messing with the OIDC parameters. Any objection?

samples/Mvc.Server/Controllers/AuthorizationController.cs
```diff
@@ -101,7 +104,9 @@ public class AuthorizationController : Controller {
             var request = HttpContext.GetOpenIdConnectRequest();
 
             return View(new LogoutViewModel {
-                Parameters = request.Parameters
+                Parameters = request.ToDictionary(
```
kevinchalet: @ilmax note that if we agree to replace the entire OIDC payload by a simple request_id, we'll have to update the logout endpoint to make the same request_id conversion as the authorization endpoint.

ilmax:
Will tale a look tommorow (i.e. this) evening!

samples/Mvc.Server/Controllers/AuthorizationController.cs
ilmax: No objection to this change, should we update the samples in the samples repo accordingly?

ilmax:
LGTM, Nice work!

src/OpenIddict.Core/OpenIddictBuilder.cs
```diff
@@ -286,49 +284,7 @@ public class OpenIddictBuilder {
...
-
-            else {
-#endif
-                key = new RsaSecurityKey(algorithm);
-#if NET451
-            }
-#endif
-
-            return Configure(options => options.SigningCredentials.AddKey(key));
+            return Configure(options => options.SigningCredentials.AddEphemeralKey());
```
ilmax, line 287: We are now calling into [this](https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/0df4b07e3b19ccf5bdb80e0eff44a07a4b217da7/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerExtensions.cs#L227) ASOS helper right?

kevinchalet, line 287: Right!

samples/Mvc.Server/Controllers/AuthorizationController.cs
kevinchalet: Yep. I'll push a new commit tomorrow and ping you when it's ready :smile:

kevinchalet:
@ilmax ICYMI, ASOS beta7 now allows us to handle token requests outside the authorization provider (e.g in a controller or in a middleware).

I'm currently working on a tiny prototype that allows devs to provide a "custom token endpoint" action, very similar to how the `Authorize` works:

``` csharp
[HttpPost("~/connect/token")]
public async Task<IActionResult> Exchange() {
    var request = HttpContext.GetOpenIdConnectRequest();

    var user = await _userManager.FindByNameAsync(request.Username);
    if (user == null) {
        return Json(new OpenIdConnectResponse {
            Error = OpenIdConnectConstants.Errors.InvalidGrant
        });
    }

    // Ensure the password is valid.
    if (!await _userManager.CheckPasswordAsync(user, request.Password)) {
        if (_userManager.SupportsUserLockout) {
            await _userManager.AccessFailedAsync(user);
        }

        return Json(new OpenIdConnectResponse {
            Error = OpenIdConnectConstants.Errors.InvalidGrant
        });
    }

    if (_userManager.SupportsUserLockout) {
        await _userManager.ResetAccessFailedCountAsync(user);
    }

    var identity = await _userManager.CreateIdentityAsync(user, request.GetScopes());

    // Create a new authentication ticket holding the user identity.
    var ticket = new AuthenticationTicket(
        new ClaimsPrincipal(identity),
        new AuthenticationProperties(),
        OpenIdConnectServerDefaults.AuthenticationScheme);

    ticket.SetResources(request.GetResources());
    ticket.SetScopes(request.GetScopes());

    return SignIn(ticket.Principal, ticket.Properties, ticket.AuthenticationScheme);
}
```

To make that optional, I'm considering creating a new `OpenIddictMiddleware` that would catch unhandled `grant_type=password` token requests and that would apply the default logic (the one we currently have in `OpenIddictProvider.HandleTokenRequest`).

We could also extend that to the authorization endpoint by providing a default UI-less experience (for simple consentless scenarios).

Any thoughts? :smile:

kevinchalet:
Updated.

ilmax:
> @ilmax ICYMI, ASOS beta7 now allows us to handle token requests outside the authorization provider (e.g in a controller or in a middleware).

I saw that, very nice addition though :smile: 
Thumbs up for consentless scenario!

Could this middelware be used as a hook for injecting custom logic?

kevinchalet:
> Could this middelware be used as a hook for injecting custom logic?

Not the middleware itself, but yeah, one will be able to inject custom logic in the controller action handling the /connect/token requests :smile:

ilmax:
LGTM

kevinchalet:
@ilmax thanks!

PR #183, opened 2016-08-02
Introduce a stricter PKCE policy than the one used in ASOS

kevinchalet:
/cc @ilmax

### DIFF
--- a/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
+++ b/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
@@ -243,14 +243,16 @@ namespace OpenIddict.Infrastructure {
                 return;
             }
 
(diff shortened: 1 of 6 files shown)


src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Authentication.cs
```diff
@@ -139,6 +139,16 @@ public partial class OpenIddictProvider<TUser, TApplication, TAuthorization, TSc
                 return;
             }
 
+            // Reject authorization requests that specify scope=offline_access if the refresh token flow is not enabled.
```
kevinchalet, line 142: Note: this block was already there. I just moved it higher :smile:

ilmax:
LGTM! :tada:

kevinchalet:
@ilmax thanks for the review! :+1:

PR #190, opened 2016-08-09
Allow handling grant_type=password token requests in user code and introduce OpenIddictMiddleware

kevinchalet:
This design change allows handling the resource owner password credentials grant token requests in a custom "token endpoint action", similarly to how we handle authorization requests. It also provides a default authorization request handling logic that "directly" returns an authorization code/access token/identity token without requiring user consent.

So how does it work?
- When a token request is received, it is first validated by ASOS/OpenIddict. If it's invalid, the request is immediately rejected with an OAuth2 error.
- If it corresponds to an ROPC request, OpenIddict calls `context.SkipToNextMiddleware()` to inform ASOS that the response won't be handled at this stage and that the next middleware in the pipeline should be invoked (for more information about that, you can read http://kevinchalet.com/2016/07/13/creating-your-own-openid-connect-server-with-asos-creating-your-own-authorization-provider/).
- Thanks to `context.SkipToNextMiddleware()`, the user controllers/middleware registered after `app.UseOpenIddict()` will be invoked. If one of these components handles the response (e.g by calling `SignInAsync` to inform ASOS that a token response should be returned), the request processing stops immediately.
- If the response is not handled by user code, `OpenIddictMiddleware` will detect it when rewinding the middleware chain. In this case, it will handle the token response by leveraging the same logic as the one we used to have in `OpenIddictProvider.HandleTokenRequest`.

The same concept also applies to authorization and logout requests. We may also want to extend that to userinfo requests.

Note: this PR is mainly a prototype (i.e it shouldn't be merged as-is).

/cc @ilmax

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,12 +152,6 @@ services.AddOpenIddict<ApplicationUser, IdentityRole<int>, ApplicationDbContext,
 
 ## Enabling interactive flows support
 
-  - **Create your own authorization controller and your own views**:
-
-To **enable authorization code/implicit flows support, you must create your own controller** and your own views/view models. The Mvc.Server sample comes with an [`AuthorizationController` that you can easily reuse in your application](https://github.com/openiddict/openiddict-core/blob/dev/samples/Mvc.Server/Controllers/AuthorizationController.cs).
-
-![](https://cloud.githubusercontent.com/assets/6998306/10988233/d9026712-843a-11e5-8ff0-e7addffd727b.png)
-
   - **Enable the corresponding flows in the OpenIddict options**:
 
 ```csharp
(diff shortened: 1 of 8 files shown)


kevinchalet:
Question: should we also extend that to `grant_type=client_credentials` token requests?

kevinchalet:
@ilmax ping :smile:

ilmax:
I really like the approach, I'm not so sure that supporting `grant_type=client_credentials` is high priority but we might support it later on.

We should put some notes on the readme to point out this extension point once this is merged in.

Nice job!

kevinchalet:
@ilmax I updated the README.md file to indicate that creating your own authorization view is no longer required (for "consentless" scenarios, at least).

Question: now that OpenIddict is able to process authorization and logout requests without having to add custom code, should we add new `EnableXYZEndpoint` overloads taking no path parameter? (i.e that would use a default /connect/xyz value)

kevinchalet:
@ilmax thanks for the review :+1:

Issue #182, opened 2016-08-01
Save access tokens to database


jayrulez:
Does openiddict support saving access tokens to the database out of the box?

I see where there is a table for token storage.

If it does, how can I enable it?

When I get a token ropc style, I do not see the record in the database.

My motivation for doing this is to implement two factor auth for ropc flow.

I am thinking of creating a TokenClaim table that stores claims for a token. One of those claims will be a two factor key. If the claim exists in the db then the user is two-factor-authenticated.

So when a user successfully gets a token, if they have two factor enabled then they will have a code sent to them.

I will have a policy to check that the currently logged in token has this token claim on any action that requires 2fa.

Also, is there any way I can subscribe to an event to know when a token is granted and run some custom code (Sending 2fa code if 2fa is enabled).

kevinchalet:
> Does openiddict support saving access tokens to the database out of the box?

No. Mainly for performance reasons. Only refresh tokens are stored and can be revoked.

> My motivation for doing this is to implement two factor auth for ropc flow.

Very bad idea. ROPC has never and will never support 2FA. Of course you could probably hack it and make your own non-standard sauce, but I don't recommend it.

You should **strongly** consider using an interactive flow like code or implicit.

jayrulez:
I do agree with you that it is a bad hack having done it in another project.

There is a native android app consuming the api. 

They would rather not to deal with non-native views... Can I still satisfy them without using ropc?

kevinchalet:
> They would rather not to deal with non-native views... Can I still satisfy them without using ropc?

The recommended approach is to open the native browser and handle everything there (registration, login, 2FA, consent, etc.).

If it's definitely not an option for you, you could consider creating your own non-standard grant... but OpenIddict deliberately doesn't allow that. I'd recommend going with ASOS in this case: http://kevinchalet.com/2016/07/13/creating-your-own-openid-connect-server-with-asos-introduction/

jayrulez:
@PinpointTownes I'll evaluate this. Thank you.

kevinchalet:
Closing. Feel free to reopen if you still need help.

jayrulez:
So a custom grant type is definitely the way to go based on my considerations over the past few weeks.

I'm at a loss here though as I see openiddict does not support custom grant types.

I would hate to have to implement all the conveniences that openiddict offers which we definitely want to make use of just to implement this.

Is there any chance that openiddict will support custom grant types in the future?

[...]

PR #205, opened 2016-08-28
Update OpenIddictProvider to allow using custom grant types

kevinchalet:
This PR removes the internal checks that reject token requests using an unknown grant type, which will allow implementing your own grant type.

Related ticket: https://github.com/openiddict/openiddict-core/issues/182.

Here's an example of how one could implement a "Facebook token exchange" grant (based on http://stackoverflow.com/questions/39008642/use-openidconnectserver-and-try-to-connect-to-facebook-through-api-service):

``` csharp
[HttpPost("~/connect/token")]
public IActionResult Exchange() {
    var request = HttpContext.GetOpenIdConnectRequest();

    if (request.GrantType == "urn:ietf:params:oauth:grant-type:facebook_access_token") {
        // Reject the request if the "assertion" parameter is missing.
        if (string.IsNullOrEmpty(request.Assertion)) {
            return Json(new OpenIdConnectResponse {
                Error = OpenIdConnectConstants.Errors.InvalidRequest,
                ErrorDescription = "The mandatory 'assertion' parameter was missing."
            });
        }

        // Create a new ClaimsIdentity containing the claims that
        // will be used to create an id_token and/or an access token.
        var identity = new ClaimsIdentity(OpenIdConnectServerDefaults.AuthenticationScheme);

        // Validate the access token using Facebook's token validation
        // endpoint and add the user claims you retrieved here.

        identity.AddClaim(ClaimTypes.NameIdentifier, "c508135b-b72a-4498-f57e-08d396b1cf4b");

        // Create a new authentication ticket holding the user identity.
        var ticket = new AuthenticationTicket(
            new ClaimsPrincipal(identity),
            new AuthenticationProperties(),
            OpenIdConnectServerDefaults.AuthenticationScheme);

        ticket.SetResources(request.GetResources());
        ticket.SetScopes(request.GetScopes());

        return SignIn(ticket.Principal, ticket.Properties, ticket.AuthenticationScheme);
    }

    return Json(new OpenIdConnectResponse {
        Error = OpenIdConnectConstants.Errors.UnsupportedGrantType
    });
}
```

/cc @ilmax @jayrulez

### DIFF
--- a/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Discovery.cs
+++ b/src/OpenIddict.Core/Infrastructure/OpenIddictProvider.Discovery.cs
@@ -4,7 +4,6 @@
  * the license and the contributors participating to this project.
  */
 
-using System.Diagnostics;
 using System.Threading.Tasks;
 using AspNet.Security.OpenIdConnect.Extensions;
 using AspNet.Security.OpenIdConnect.Server;
@@ -30,13 +29,6 @@ namespace OpenIddict.Infrastructure {
 
             // Copy the supported grant types list to the discovery document.
             foreach (var type in services.Options.GrantTypes) {
-                Debug.Assert(type == OpenIdConnectConstants.GrantTypes.AuthorizationCode ||
-                             type == OpenIdConnectConstants.GrantTypes.ClientCredentials ||
-                             type == OpenIdConnectConstants.GrantTypes.Implicit ||
-                             type == OpenIdConnectConstants.GrantTypes.Password ||
-                             type == OpenIdConnectConstants.GrantTypes.RefreshToken,
-                             "Unsupported or non-standard OAuth2/OIDC grant types should not be exposed.");
-
                 context.GrantTypes.Add(type);
             }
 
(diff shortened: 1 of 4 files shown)


jayrulez:
Awesome. We are very pleased with this change :)

Kukks:
@PinpointTownes Would this be the recommended way of adding facebook authentication?

kevinchalet:
@Kukks wooooops, sorry, looks like I missed your post :sweat: 

Though it's now officially supported, it's not the approach I'd recommend, as it's harder to implement for beginners and not really standard (the token validation part is necessarily provider-specific). That said, if using interactive flows like the code flow or the implicit flow is not an option (e.g because you absolutely want to rely on the mobile SDKs offered by Facebook instead of using the device browser), then, you can consider using it if you know what you're doing.

Related ticket: https://github.com/openiddict/openiddict-core/issues/232.

Kukks:
@PinpointTownes That's fine. I've created a repo to create a working proof of concept of this use case as this is what I'll be needing. Thank you for your great work and effort! 

https://github.com/Kukks/openiddict-custom-grants-example/

Issue #209, opened 2016-08-31
Example in readme produces unexpected results


jayrulez:
The ROPC example in the README file returns 200 OK when credential are invalid.

return Json(... will return 200.

jayrulez:
Should be BadResult(... when there is an error

kevinchalet:
Nice catch. I'll replace it by `BadRequest` and decorate the action with `[Produces("application/json")]` to ensure the JSON formatter is always used to serialize the response.

PR #210, opened 2016-08-31
Replace Controller.Json() by ControllerBase.BadRequest() to return a 400 response for invalid token requests

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/209.

@jayrulez care to review? :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,14 +156,16 @@ To **support the password or the client credentials flow, you must provide your 
 
 ```csharp
 [HttpPost("~/connect/token")]
+[Produces("application/json")]
 public async Task<IActionResult> Exchange() {
     var request = HttpContext.GetOpenIdConnectRequest();
 
     if (request.IsPasswordGrantType()) {
         var user = await _userManager.FindByNameAsync(request.Username);
         if (user == null) {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -156,14 +156,16 @@ To **support the password or the client credentials flow, you must provide your
 
 ```csharp
 [HttpPost("~/connect/token")]
+[Produces("application/json")]
```
jayrulez, line 159: Is this line necessary?

I get application/json content type in the response headers when i sent that header.

jayrulez:
Looks good @PinpointTownes

jayrulez:
Ah, I didn't see your comment on the issue before I reviewed. Since we want to ensure that the json formatter serializes the object then disregard my question above.

README.md
kevinchalet: If you have multiple formatters (e.g if you register the XML formatter), I think there's a risk of getting XML if the client sent an `Accept` with XML specified before JSON.

jayrulez: I never really thought about that before. Thanks for pointing that out.

kevinchalet:
Thanks for the review :+1:

PR #248, opened 2016-10-22
Override default port

ORuban:
allows to run both Client and Server out of the box


kevinchalet:
@ORuban I'm not opposed to this change, as long as it doesn't break the IIS hosting scenario. Can you please triple check it's still working as intended?

PR #273, opened 2016-11-16
Migrate to .NET Core/ASP.NET Core 1.1.0

kevinchalet:


kevinchalet:
Postponed to the next milestone, as ASP.NET Core 1.1.0 is impacted by a bunch of annoying bugs.

JeanCollas:
Any update on this issue? :)

kevinchalet:
> Any update on this issue? :)

All the aspnet-contrib and OpenIddict projects will keep using the LTS packages (1.0.x) and there's no plan to change that, so this PR won't be merged.

The cool news is that you can use OpenIddict with your ASP.NET Core 1.1.0 app. You still have to rework your code a bit to work around the "multiple automatic handlers issue" (https://github.com/aspnet/Security/issues/1062), but it will be fixed in the 1.1.1 bits.

JeanCollas:
Yes, so it means that we just have to wait for the ASP.NET Core 1.1.1, right? :)
Any idea when it should be? (I couldn't find it, and dates are often not up to date!)

kevinchalet:
> Yes, so it means that we just have to wait for the ASP.NET Core 1.1.1, right? :)

Not really, the link I posted mentions everything you need to "fix" your app to work with the 1.1.0 bits.

> Any idea when it should be? (I couldn't find it, and dates are often not up to date!)

No idea. That's a question you should ask to @Eilon or @Tratcher :smile:

Eilon:
We don't have a date set for 1.1.1, but we're working on getting fixes ready. Given the US holiday schedule, it'll almost certainly be early next year (as opposed to December).

kevinchalet:
@Eilon thanks for the precision! :+1:

> Given the US holiday schedule, it'll almost certainly be early next year (as opposed to December).

Damn, it was on my wishlist! Santa is not cool this year :trollface:

JeanCollas:
Thanks! :)

PR #274, opened 2016-11-20
Update README.md

JeanCollas:
Extraction of dev-specific options to be applied only in dev environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,12 @@ To use OpenIddict, you need to:
   - **Configure the OpenIddict services** in `Startup.ConfigureServices`:
 
 ```csharp
+public Startup(IHostingEnvironment env) {
+        CurrentEnvironment = env;
+}
+
+private IHostingEnvironment CurrentEnvironment { get; set; }
+
 public void ConfigureServices(IServiceCollection services) {
     services.AddMvc();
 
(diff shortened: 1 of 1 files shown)


kevinchalet:
Thanks for the PR :smile: 

Unfortunately, I'm not sure the suggested changes are really the right thing to do: don't forget that an exception will be thrown if no signing key is registered (e.g if you don't call `AddEphemeralSigningKey()` nor `AddSigningCertificate(...)`).

JeanCollas:
I didn't know that (yet!) :) 

I guess you were talking about that part of the code:

```
// On production, using a X.509 certificate stored in the machine store is recommended.
// You can generate a self-signed certificate using Pluralsight's self-cert utility:
// https://s3.amazonaws.com/pluralsight-free/keith-brown/samples/SelfCert.zip
// 
// services.AddOpenIddict<ApplicationUser, IdentityRole<Guid>, ApplicationDbContext, Guid>()
//     .AddSigningCertificate("7D2A741FE34CC2C7369237A5F2078988E17A6A75");
// 
```

Should it be the same certificate that the one used by the https connection? Or an independant one? (sorry I am not expert in identification issues :) )

Is it linked to the following call?

```
// During development, you can disable the HTTPS requirement.
   .DisableHttpsRequirement()
```

kevinchalet:
I'd recommend using a separate (self-signed) certificate. And no, it's not linked to `DisableHttpsRequirement()`, whose only objective is to ask OpenIddict to stop rejecting non-HTTPS requests.

PR #295, opened 2016-12-13
Update the samples to register the OpenIddict entity sets directly from Startup.cs

kevinchalet:
It looks like the new VS tooling has a weird issue that makes `builder.UseOpenIddict()` really hard to use from `OnModelCreating()`. This PR removes this extension to favor the DI approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,15 @@ To use OpenIddict, you need to:
 public void ConfigureServices(IServiceCollection services) {
     services.AddMvc();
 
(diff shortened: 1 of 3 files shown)


kevinchalet:
Update: to support scenarios using `options.UseInternalServiceProvider(provider)` (where `options.UseOpenIddict()` cannot be used), we'll keep the `ModelBuilder` extensions.

This PR will simply update the Mvc.Server sample to use the DI approach.

PR #346, opened 2017-01-26
Relax the default signing credentials policy to allow using OpenIddict in degraded mode

kevinchalet:
Related PR: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/pull/377.

With this change, registering a signing key/certificate will no longer be necessary to use the code flow, the password flow, the client client credentials, the refresh token flow or any custom flow. Of course, it will still be needed when using JWT as the access token format or when using the implicit flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ public void ConfigureServices(IServiceCollection services) {
 
     // Register the Identity services.
 	services.AddIdentity<ApplicationUser, IdentityRole>()
-	    .AddEntityFrameworkCoreStores<ApplicationDbContext>()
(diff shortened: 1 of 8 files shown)


Bartmax:
is still recommended to have a signing key/certificate or it was never useful ?

kevinchalet:
The signing key/certificate will still be used to sign the JWT tokens returned by OpenIddict, but it will now accept returning unsigned tokens under certain circumstances, as allowed by the specification (i.e it's not a JWT access token and it's returned from the token endpoint, where TLS server validation can be enforced).

PR #376, opened 2017-03-13
Migrate to MSBuild 15 csproj #375

### LINKED ISSUE #375: Migrate to MSBuild 15 csproj
@PinpointTownes, can we migrate to MSBuild 15 csproj that works in VS 2017?

Bartmax:
this attempt to fix: #375

See if this makes sense.
I keep the lines about `Microsoft.NETCore.Platforms` commented. The auto migration brings those from json to csproj. (pretty sure is not needed, just need confirmation)


kevinchalet:
@Bartmax FYI, I spent a few hours migrating most the aspnet-contrib repos to .csproj and it looks like everything is now working fine on both AppVeyor and Travis (see https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/pull/63/files for an example).

That would be great if you could update your PR to use the same approach :+1:

Bartmax:
Great! I will be able to take a look in 2 - 3 days.

kevinchalet:
Replaced by https://github.com/openiddict/openiddict-core/commit/112c46deb63d5c1084cd7c9a6ed727036a4d8346.

Thanks! :clap:

xperiandri:
@Bartmax, where have you seen `Microsoft.NETCore.Platforms`?

Bartmax:
@xperiandri on csprojs

PR #397, opened 2017-05-05
Update CryptoHelper to 2.0.1 patch release

henkmollema:
I've created a patch release for 2.0.x including the LTS packages of ASP.NET Core. Specifically `Microsoft.AspNetCore.Cryptography.KeyDerivation` version 1.0.2.


kevinchalet:
Thanks for the PR :+1: 

I'm curious: is the `Microsoft.AspNetCore.Cryptography.KeyDerivation` version being updated the only change? 'cause we usually don't target patched versions in the aspnet-contrib/OpenIddict packages, leaving that to the end users (explicitly or implicitly if they transitively reference newer packages).

henkmollema:
@PinpointTownes yes, updating the package is the only change. Just to make sure the latest patch release is pulled in transitively. Microsoft also uses this approach for their own packages. If you want to leave that choice up to the end user, it's fine and you can ignore this PR.

kevinchalet:
@henkmollema thanks for the confirmation :+1:

PR #420, opened 2017-07-14
Add options for customize database tables.

mczachurski:
We cannot have OpenIddict prefix in table names. Pull request add possibilities to customize names for OpenIddict tables. 

```csharp
builder.UseOpenIddict(options =>
{
    options.ApplicationsTableName = "Applications";
    options.AuthorizationsTableName = "Authorizations";
    options.ScopesTableName = "Scopes";
    options.TokensTableName = "Tokens";
});
```

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/OpenIddictCustomizer.cs
+++ b/src/OpenIddict.EntityFrameworkCore/OpenIddictCustomizer.cs
@@ -24,9 +24,14 @@ namespace OpenIddict.EntityFrameworkCore
         where TToken : OpenIddictToken<TKey, TApplication, TAuthorization>, new()
         where TKey : IEquatable<TKey>
     {
-        public OpenIddictCustomizer([NotNull] ModelCustomizerDependencies dependencies)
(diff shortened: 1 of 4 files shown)


kevinchalet:
Interesting. Is there any particular reason you couldn't do that in your own code? This is the pattern we usually use for both Identity and OpenIddict:

```csharp
public class ApplicationDbContext : IdentityDbContext<ApplicationUser>
{
    public ApplicationDbContext(DbContextOptions options)
        : base(options) { }

    protected override void OnModelCreating(ModelBuilder builder)
    {
        base.OnModelCreating(builder);

        // Customize the ASP.NET Identity model and override the defaults if needed.
        // For example, you can rename the ASP.NET Identity table names and more.
        // Add your customizations after calling base.OnModelCreating(builder);

        builder.Entity<[entity]>(entity => entity.ToTable("[table name]"));
    }
}
```

mczachurski:
Indeed it works :+1: We can use pattern proposed by you. Thanks.

PR #429, opened 2017-08-18
Update to CryptoHelper 3.0.0

henkmollema:
CryptoHelper 3.0 targets .NET Standard 2.0 and the 2.0 version of ASP.NET Core Data Protection


kevinchalet:
Awesome, thanks @henkmollema!

Out of curiosity, do you plan to release a strong-signed version of your library? That's currently the only dependency we have that would prevent OpenIddict from being strong-signed.

henkmollema:
You're welcome! 👍 

R.e.: strong naming. Would [StrongNamer](https://www.nuget.org/packages/StrongNamer/) help you out?

PR #431, opened 2017-08-19
Update OpenIddictCustomizer to inherit from RelationalModelCustomizer instead of ModelCustomizer

kevinchalet:
/cc @MoxxNull @gizmohd @bradmarder


Bartmax:
this was the problem with the table names?

kevinchalet:
@Bartmax yep. This is the workaround suggested by the EF team: https://github.com/aspnet/EntityFrameworkCore/issues/9225#issuecomment-323409665.

kevinchalet:
Note: backporting this workaround to OpenIddict 1.x is not necessary as EF Core 1.x is not affected.

PR #432, opened 2017-08-19
Add IssuedUtc and ExpiresUtc fields to OpenIddictToken

vyfster:
Fields (`IssuedUtc` and `ExpiresUtc`) added to help keep track of tokens added to the `OpenIddictTokens` table so that expired tokens can be easily identified.

Unit tests have been updated with the change, however the tests `SerializeAuthorizationCode_AuthorizationCodeIsCorrectlyPersisted` and `SerializeRefreshToken_RefreshTokenIsCorrectlyPersisted` should probably include a check the verifies that the values that have been persisted for these two fields is correct.  For this to happen, the `AuthenticationTicket` that is created in `OpenIddictProviderTests` line 142 needs to be available to the unit tests.  I'm not sure on the best approach to achieve this - any hints / guidelines on doing this?

*** NOTE - this breaks if used with an existing database that has not been updated.  Two new columns must be added to the table `OpenIddictTokens`.  Alternatively, if the data in the database is not needed, as it's in dev, then simply drop the database.

**MSSQL**
IssuedUtc DateTimeOffset (NULL)
ExpiresUtc DateTimeOffset (NULL)

**MySQL || MariaDB**
IssuedUtc DATETIME (NULL)
ExpiresUtc DATETIME (NULL)

### DIFF
--- a/src/OpenIddict.Core/Managers/OpenIddictTokenManager.cs
+++ b/src/OpenIddict.Core/Managers/OpenIddictTokenManager.cs
@@ -61,11 +61,13 @@ namespace OpenIddict.Core
         /// </summary>
         /// <param name="type">The token type.</param>
         /// <param name="subject">The subject associated with the token.</param>
+        /// <param name="issuedUtc">The date time that the token was issued.</param>
+        /// <param name="expiresUtc">The date time that the token will expire.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
         /// <returns>
         /// A <see cref="Task"/> that can be used to monitor the asynchronous operation, whose result returns the token.
         /// </returns>
-        public virtual async Task<TToken> CreateAsync([NotNull] string type, [NotNull] string subject, CancellationToken cancellationToken)
+        public virtual async Task<TToken> CreateAsync([NotNull] string type, [NotNull] string subject, DateTimeOffset? issuedUtc, DateTimeOffset? expiresUtc, CancellationToken cancellationToken)
         {
             if (string.IsNullOrEmpty(type))
             {
@@ -83,7 +85,7 @@ namespace OpenIddict.Core
                 throw new ArgumentException("The subject cannot be null or empty.");
             }
 
-            return await Store.CreateAsync(type, subject, cancellationToken);
+            return await Store.CreateAsync(type, subject, issuedUtc, expiresUtc, cancellationToken);
         }
 
         /// <summary>
--- a/src/OpenIddict.Core/Stores/IOpenIddictTokenStore.cs
+++ b/src/OpenIddict.Core/Stores/IOpenIddictTokenStore.cs
@@ -7,6 +7,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
+using System;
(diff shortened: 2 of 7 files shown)


kevinchalet:
> For this to happen, the AuthenticationTicket that is created in OpenIddictProviderTests line 142 needs to be available to the unit tests. I'm not sure on the best approach to achieve this - any hints / guidelines on doing this?

There's a simple way to do that: after line 142, add `ticket.Properties.IssuedUtc = new DateTimeOffset(... fixed date...)`, which will force OpenIddict to use a fixed date instead of the current date time. Then, call `ticket.SetAuthorizationCodeLifetime(TimeSpan.FromDays(1))` (or `ticket.SetRefreshTokenLifetime(TimeSpan.FromDays(1))`) to set the expiration date to `ticket.Properties.IssuedUtc` + 1 day, and update your tests to use your hardcoded dates instead of `It.IsAny<DateTimeOffset?>()`.

kevinchalet:
Note: I'll probably wait a bit before merging your PR. I'd like to combine it with https://github.com/openiddict/openiddict-core/issues/433 - which will require massive changes in the `OpenIddictToken` entity - so we avoid breaking people twice :sweat_smile:

vyfster:
> There's a simple way to do that: after line 142, add ticket.Properties.IssuedUtc = new DateTimeOffset(... fixed date...), which will force OpenIddict to use a fixed date instead of the current date time.

Unfortunately this doesn't seem to be the case.  In the file `OpenIddictProvider.Serialization.cs`, method `SerializeAuthorizationCode`, line 37, the value for `context.Ticket.Properties.IssuedUtc` seems to be set to DateTime.UtcNow (or similar value), regardless of whether the value for this property has been set.

I've had a dig through the AspNet.Security.OpenidConnect.Server code and it looks like the `IssuedUtc` property is set there (code below is present in multiple places):
```c#
            // Create a new ticket containing the updated properties.
            var ticket = new AuthenticationTicket(identity, properties);
            ticket.Properties.IssuedUtc = Options.SystemClock.UtcNow;
```

I'm not familiar with the code so I might be wrong wrt ASOS, but IssuedUtc is always set to a utc date regardless of whether or not it has a value.

Simple way to test - set to a hardcoded value as described in the comments above and then set a breakpoint at line 37 in `OpenIddictProvider.Serialization.cs`.

kevinchalet:
Arg yeah, that's true (I now remember I changed that at some point, when I decided to ignore `ExpiresUtc` and replace it by the `ticket.SetXYZLifetime()` extensions).

Another approach is to use a mocked clock to ensure a fixed date is used for `IssuedUtc`:

E.g:

```csharp
var server = CreateAuthorizationServer(builder =>
{
    builder.Configure(options =>
    {
        options.SystemClock = Mock.Of<ISystemClock>(mock => mock.UtcNow ==
            new DateTimeOffset(2017, 01, 01, 00, 00, 00, TimeSpan.FromHours(0)));
    });

    builder.Services.AddSingleton(manager);
});
```

kevinchalet:
Your commit has been integrated into https://github.com/openiddict/openiddict-core/pull/443 and backported to OpenIddict 1.x.

Thanks for your contribution! :clap:

PR #525, opened 2017-12-10
Rename OpenIddictToken.Ciphertext/Hash to Payload/ReferenceId and add new methods to allow overriding the default obfuscation

kevinchalet:
An OpenIddict consumer wants to be able to opt out of our default hashing method to use encryption instead. This PR introduces a new method in the token manager to allow overriding how the identifier is obfuscated in the database (by default, it hashes the generated identifier using SHA256 and stores it as a base64-encoded in the database).

Note: OpenIddict users moving to a package version including this commit will have to add a migration to take the schema changes into account.


kevinchalet:
@asadsahi FYI this PR introduced a schema change. You may want to add a migration in your project if you use the latest OI bits.

asadsahi:
@PinpointTownes thanks for letting me know. Will do the necessary changes.

PR #537, opened 2018-01-17
Removed mention of myget

erikrenaud:
MyGet mention was removed, not sure if that delivery mechanism will be maintained.
Since there are instructions for adding the packages directly to csproj, no specific mention of having to tick pre-release in the nuget manager is done.


kevinchalet:
Merged, thanks @tuath!

PR #562, opened 2018-02-15
Introduce OpenIddictApplication.ConsentType and add new authorization helpers

kevinchalet:
(most likely the last work item for RC2 :tada:)

### DIFF
--- a/samples/Mvc.Server/Startup.cs
+++ b/samples/Mvc.Server/Startup.cs
@@ -90,9 +90,10 @@ namespace Mvc.Server
                        .AllowPasswordFlow()
                        .AllowRefreshTokenFlow();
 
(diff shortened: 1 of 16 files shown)


xperiandri:
What is the purpose of `ConsentTypes.External`?
How would I use it?

Could you add XML comments with a short description for each of `ConsentTypes`?

danieljee:
> What is the purpose of `ConsentTypes.External`?
> How would I use it?
> 
> Could you add XML comments with a short description for each of `ConsentTypes`?

I think it is used for application that is granted authorization externally (E.g. by system admins). System admin creates authorization for the application so it doesn't require End-User's consent during Authorization flow.

Issue #433, opened 2017-08-22
Implement reference tokens support as an opt-in feature


kevinchalet:
## Objective:

Offering built-in support for reference tokens in OpenIddict, so that things like immediate access token revocation can be easily achieved in OpenIddict-based apps.

## Proposed API:

```csharp
/// <summary>
/// Configures to OpenIddict to use reference tokens, so that authorization codes,
/// access tokens and refresh tokens are stored as ciphertext in the database
/// (only an identifier is returned to the client application). Enabling this option
/// is useful to keep track of all the issued tokens or when storing a very large
/// number of claims in the authorization codes, access tokens and refresh tokens.
/// Note: this option cannot be used when configuring JWT as the access token format.
/// </summary>
/// <param name="builder">The services builder used by OpenIddict to register new services.</param>
/// <returns>The <see cref="OpenIddictBuilder"/>.</returns>
public static OpenIddictBuilder UseReferenceTokens([NotNull] this OpenIddictBuilder builder)
{
    if (builder == null)
    {
        throw new ArgumentNullException(nameof(builder));
    }

    return builder.Configure(options => options.UseReferenceTokens = true);
}
````

## Usage:

```csharp
public void ConfigureServices(IServiceCollection services)
{
    // Register the OpenIddict services.
    services.AddOpenIddict(options =>
    {
        // ...

        // Configure OpenIddict to use reference tokens.
        options.UseReferenceTokens();
    });
}
```

Open question: should we have an equivalent to the aspnet-contrib validation middleware for reference tokens, or is the introspection middleware enough for this task?

kinosang:
one more question, if we use reference tokens without a new validation middleware, how can we get Userinfo endpoint working? I don't want to use the introspection middleware in the OpenID Connect Server project, it's ugly and waste.

kevinchalet:
Currently, it's introspection or nothing.

kinosang:
@PinpointTownes I see. but when I try to use the introspection endpoint to validate the token with Postman, it always return false.

e.g.

POST http://localhost:55179/Api/Introspect
token=_YMBhvhLMzZm9th6kcreFY0s243-9EyK7qCmUqQmKyE

```json
{
    "active": false
}
```

kevinchalet:
What do the logs say?

kinosang:
It says `The introspection request was rejected because the access token was issued to a different client or for another resource server.`

but I've already set the RP to "http://localhost:55179/"

[...]

PR #589, opened 2018-04-16
Introduce a validation middleware with reference tokens support

kinosang:
#433 and #443 introduced the reference tokens support,

but there's no validation middleware for that,

we should use Introspection middleware and Introspection API with HTTP calling to validate the tokens on a OpenIdConnect server itself,

so I'm trying to make the middleware.

### DIFF
--- a/OpenIddict.sln
+++ b/OpenIddict.sln
@@ -52,6 +52,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenIddict.Server", "src\Op
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenIddict.Server.Tests", "test\OpenIddict.Server.Tests\OpenIddict.Server.Tests.csproj", "{07B02B98-8A68-432D-A932-48E6D52B221A}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenIddict.Validation.Tests", "test\OpenIddict.Validation.Tests\OpenIddict.Validation.Tests.csproj", "{F470E734-F4B6-4355-AF32-53412B619E41}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenIddict.Validation", "src\OpenIddict.Validation\OpenIddict.Validation.csproj", "{6AB8F9E7-47F8-4A40-837F-C8753362AF54}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -122,6 +126,14 @@ Global
 		{07B02B98-8A68-432D-A932-48E6D52B221A}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{07B02B98-8A68-432D-A932-48E6D52B221A}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{07B02B98-8A68-432D-A932-48E6D52B221A}.Release|Any CPU.Build.0 = Release|Any CPU
+		{F470E734-F4B6-4355-AF32-53412B619E41}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{F470E734-F4B6-4355-AF32-53412B619E41}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{F470E734-F4B6-4355-AF32-53412B619E41}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{F470E734-F4B6-4355-AF32-53412B619E41}.Release|Any CPU.Build.0 = Release|Any CPU
+		{6AB8F9E7-47F8-4A40-837F-C8753362AF54}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{6AB8F9E7-47F8-4A40-837F-C8753362AF54}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{6AB8F9E7-47F8-4A40-837F-C8753362AF54}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{6AB8F9E7-47F8-4A40-837F-C8753362AF54}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -143,6 +155,8 @@ Global
 		{275D888A-B4C8-4E93-AC4B-B1AA25D98159} = {D544447C-D701-46BB-9A5B-C76C612A596B}
 		{21A7F241-CBE7-4F5C-9787-F2C50D135AEA} = {D544447C-D701-46BB-9A5B-C76C612A596B}
 		{07B02B98-8A68-432D-A932-48E6D52B221A} = {5FC71D6A-A994-4F62-977F-88A7D25379D7}
+		{F470E734-F4B6-4355-AF32-53412B619E41} = {5FC71D6A-A994-4F62-977F-88A7D25379D7}
+		{6AB8F9E7-47F8-4A40-837F-C8753362AF54} = {D544447C-D701-46BB-9A5B-C76C612A596B}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {A710059F-0466-4D48-9B3A-0EF4F840B616}
--- a/src/OpenIddict.Validation/OpenIddict.Validation.csproj
+++ b/src/OpenIddict.Validation/OpenIddict.Validation.csproj
@@ -1,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\packages.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>OpenIdDict reference tokens validation middleware for ASP.NET Core.</Description>
+    <Authors>Kévin Chalet;Chino Chang</Authors>
+    <PackageTags>aspnetcore;authentication;jwt;openidconnect;openiddict;security</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\OpenIddict.Core\OpenIddict.Core.csproj" />
+    <ProjectReference Include="..\OpenIddict.Models\OpenIddict.Models.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AspNet.Security.OAuth.Validation" Version="$(AspNetContribOpenIdExtensionsVersion)" />
+    <PackageReference Include="AspNet.Security.OpenIdConnect.Extensions" Version="$(AspNetContribOpenIdExtensionsVersion)" />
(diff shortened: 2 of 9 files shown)


kevinchalet:
Thanks for contributing to this task, it's much appreciated! :clap: 

Things that would be nice to do:
  - Supporting both the default token format and reference tokens. The logic is so similar that is should only require a 2/3-line change (reference tokens are just IDs for opaque tokens stored in the DB).
  - Adding a `UseReferenceTokens` property to `OpenIddictValidationOptions.cs` 
  - Renaming the project to just `OpenIddict.Validation` and the validation handler to `OpenIddictValidationHandler`. I'll consider renaming the server one to `OpenIddictServerHandler` in a separate PR.
  - Once this PR is done, adding a similar middleware in the `release` branch (for 1.0).

> This middleware does not work at this time

I'm pretty sure it doesn't work yet because you must replace the default access token format when reference tokens are used. Take a look at https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/Internal/OpenIddictInitializer.cs#L72-L95 to see how it's done for the server-side.

kevinchalet:
Note: instead of copying the entire validation handler to this repo, we may want to add some protected virtual methods in the aspnet-contrib validation handler and only override the methods we need in the OpenIddict validation handler. This way, we wouldn't have two implementations to keep in sync and we would avoid code duplication.

kevinchalet:
Second note: we'll also need to make the handler class generic to ensure folks can use it even if they don't use our `OpenIddictToken` entity.

kinosang:
@PinpointTownes Now the middleware is named `OpenIddict.Validation`, it extends the aspnet-contrib oauth validation middleware.

The handler class is generic now and the extension also provides generic methods.

kevinchalet:
Thanks! :+1: 

Quick question: can we make `OpenIddictValidationHandler` derive from `OAuthValidationHandler` so we can remove all the challenge stuff from `OpenIddictValidationHandler`?

In a future PR, we may want to consider making the validation extensions API extend `OpenIddictBuilder` instead of `AuthenticationBuilder`. This way, we wouldn't have to specify the `TToken` parameter a second time, as it could be retrieved from `OpenIddictBuilder.TokenType`:

```csharp
services.AddOpenIddict<[entity types]>(builder =>
{
    builder.AddValidation(options => options.UseReferenceTokens = true);
});

services.AddOpenIddict<[entity types]>()
    .AddValidation(options => options.UseReferenceTokens = true);
```

We could also adopt the same pattern for the server part, instead of magically registering the server handler when `Configure` is called:

```csharp
services.AddOpenIddict<[entity types]>(builder =>
{
    builder.AddServer(options => options.UseReferenceTokens = true);
    builder.AddValidation(options => options.UseReferenceTokens = true);
});

services.AddOpenIddict<[entity types]>()
    .AddServer(options => options.UseReferenceTokens = true)
    .AddValidation(options => options.UseReferenceTokens = true);
```

(no need to update your PR to take that into account, I'll think more about that before changing anything)

extensions/OpenIddict.Validation/OpenIddict.Validation.csproj
```diff
@@ -0,0 +1,27 @@
...
+
+  <Import Project="..\..\build\packages.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>OpenIdDict reference tokens validation middleware for ASP.NET Core.</Description>
+    <Authors>Kévin Chalet,Chino Chang</Authors>
```
kevinchalet: `,` -> `;`

build/dependencies.props
```diff
@@ -1,6 +1,7 @@
 <Project>
 
   <PropertyGroup Label="Package Versions">
+    <AspNetContribOauthExtensionsVersion>2.0.0-rc2-final</AspNetContribOauthExtensionsVersion>
```
kevinchalet: Consider using the existing `AspNetContribOpenIdExtensionsVersion` prop.

extensions/OpenIddict.Validation/OpenIddict.Validation.csproj
```diff
@@ -0,0 +1,27 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\OpenIddict.Core\OpenIddict.Core.csproj" />
+    <ProjectReference Include="..\..\src\OpenIddict.Models\OpenIddict.Models.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AspNet.Security.OAuth.Validation" Version="$(AspNetContribOauthExtensionsVersion)" />
+    <PackageReference Include="AspNet.Security.OpenIdConnect.Server" Version="$(AspNetContribOpenIdServerVersion)" />
```
kevinchalet: Consider removing that. Feel free to hardcode the Data Protection purposes instead of using `nameof`.

extensions/OpenIddict.Validation/OpenIddictValidationExtensions.cs
```diff
@@ -0,0 +1,155 @@
...
+        
+        /// <summary>
+        /// Adds a new instance of the OpenIddict validation middleware in the ASP.NET Core pipeline.
+        /// </summary>
+        /// <param name="builder">The authentication builder.</param>
+        /// <param name="scheme">The authentication scheme associated with this instance.</param>
+        /// <returns>The authentication builder.</returns>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        public static AuthenticationBuilder AddOpenIddictValidation(
+            [NotNull] this AuthenticationBuilder builder, [NotNull] string scheme)
```
kevinchalet: Let's remove all the overloads accepting a `scheme` (we don't support customizing the scheme name of the server handler, so we shouldn't allow replacing the default validation scheme).

extensions/OpenIddict.Validation/OpenIddict.Validation.csproj
kinosang: @PinpointTownes `AspNet.Security.OpenIdConnect.Server` also provides the `AuthenticationProperties.GetProperty()` method which is required in the handler

extensions/OpenIddict.Validation/OpenIddictValidationHandler.cs
```diff
@@ -0,0 +1,431 @@
...
+                }
+            }
+
+            var ticket = Options.AccessTokenFormat.Unprotect(payload);
+
+            if (ticket == null)
+            {
+                return AuthenticateResult.Fail("Authentication failed because the access token was invalid.");
+            }
+
```
kevinchalet: For reference tokens, don't forget to add the code logic that restores the issued/expires dates in the authentication ticket:

```csharp
if (Options.UseReferenceTokens)
{
    // Restore the token identifier using the unique
    // identifier attached with the database entry.
    ticket.Properties.Items[OpenIddictConstants.Properties.TokenId] = identifier;

    // Dynamically set the creation and expiration dates.
    ticket.Properties.IssuedUtc = await Tokens.GetCreationDateAsync(token);
    ticket.Properties.ExpiresUtc = await Tokens.GetExpirationDateAsync(token);

    // Restore the authorization identifier using the identifier attached with the database entry.
    ticket.Properties.Items[OpenIddictConstants.Properties.AuthorizationId] =
        await Tokens.GetAuthorizationIdAsync(token);
}
```

extensions/OpenIddict.Validation/OpenIddict.Validation.csproj
kevinchalet: > @PinpointTownes AspNet.Security.OpenIdConnect.Server also provides the AuthenticationProperties.GetProperty() method which is required in the handler

Consider using `ticket.Properties.Items.TryGetValue("property name", out string value)` instead of the convenient helper :smile:

kevinchalet:
Note: we'll also need to port the validation handler tests to this repo.

kinosang:
@PinpointTownes 
> Quick question: can we make OpenIddictValidationHandler derive from OAuthValidationHandler so we can remove all the challenge stuff from OpenIddictValidationHandler?

I've tried that, but if we derive `OpenIddictValidationHandler` from `OAuthValidationHandler`, `OpenIddictValidationHandler.Options` will be type of `OAuthValidationOptions`, so we'll got:

```
There is no implicit reference conversion from 'OpenIddict.Validation.OpenIddictValidationHandler<TToken>' to 'Microsoft.AspNetCore.Authentication.AuthenticationHandler<OpenIddict.Validation.OpenIddictValidationOptions>'
```

kevinchalet:
You can workaround it by doing that:

```csharp
private new OpenIddictValidationOptions Options => (OpenIddictValidationOptions) base.Options;
```

kinosang:
@PinpointTownes Nope, the error is reported at `OpenIddictValidationExtensions` L88.

If we derive `OpenIddictValidationHandler` from `OAuthValidationHandler`, it will not accept `OpenIddictValidationOptions` as the first parameter for `AuthenticationBuilder.AddScheme`.

kevinchalet:
> If we derive OpenIddictValidationHandler from OAuthValidationHandler, it will not accept OpenIddictValidationOptions as the first parameter for AuthenticationBuilder.AddScheme.

Replace the `builder.AddScheme` call by:

```csharp
// Register the OpenIddict validation handler in the authentication options,
// so it can be discovered by the default authentication handler provider.
builder.Services.Configure<AuthenticationOptions>(options =>
{
    // Note: similarly to Identity, OpenIddict should be registered only once.
    // To prevent multiple schemes from being registered, a check is made here.
    if (options.SchemeMap.ContainsKey(OpenIddictValidationDefaults.AuthenticationScheme))
    {
        return;
    }

    options.AddScheme(OpenIddictValidationDefaults.AuthenticationScheme, scheme =>
    {
        scheme.HandlerType = typeof(OpenIddictValidationHandler<TToken>);
    });
});

builder.Services.Configure(OpenIddictValidationDefaults.AuthenticationScheme, configuration);

return builder;
```

I just tested and it works like a charm.

OpenIddict.sln
```diff
@@ -52,6 +52,12 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenIddict.Server", "src\Op
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenIddict.Server.Tests", "test\OpenIddict.Server.Tests\OpenIddict.Server.Tests.csproj", "{07B02B98-8A68-432D-A932-48E6D52B221A}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "extensions", "extensions", "{C45757B6-F8F0-4556-8A8A-16BBD1C59FDA}"
```
kevinchalet: Could you please remove that and move the validation folder to the main `src`?

kevinchalet:
> In a future PR, we may want to consider making the validation extensions API extend OpenIddictBuilder instead of AuthenticationBuilder.

I made a prototype for the server part: https://github.com/openiddict/openiddict-core/pull/590

kinosang:
@PinpointTownes gotcha, I'll update this PR asap.

kinosang:
@PinpointTownes All done, and I'll follow the changes of #590 once it's merged.

kevinchalet:
Merged (with some major modifications to use the new builders API): https://github.com/openiddict/openiddict-core/commit/03a27051ec51061151895e54458bf1ebf51bd484

It was also backported to OpenIddict 1.x: https://github.com/openiddict/openiddict-core/commit/59c3d51ae751a71847793ea35e9affa80358989c

Thanks for your contribution! :tada:

kevinchalet:
Note: I'll take some time tomorrow to publish an announcement for both the validation middleware and the revamped OpenIddict services registration APIs.

PR #615, opened 2018-06-01
Made store properties protected virtual

xperiandri:
They must be such to be able to add additional filtering without whole class reimplementation.

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictApplicationStore.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictApplicationStore.cs
@@ -108,17 +108,17 @@ namespace OpenIddict.EntityFrameworkCore
         /// <summary>
         /// Gets the database set corresponding to the <typeparamref name="TApplication"/> entity.
         /// </summary>
-        private DbSet<TApplication> Applications => Context.Set<TApplication>();
+        protected virtual DbSet<TApplication> Applications => Context.Set<TApplication>();
 
         /// <summary>
         /// Gets the database set corresponding to the <typeparamref name="TAuthorization"/> entity.
         /// </summary>
-        private DbSet<TAuthorization> Authorizations => Context.Set<TAuthorization>();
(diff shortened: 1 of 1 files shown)


kevinchalet:
> They must be such to be able to add additional filtering without whole class reimplementation.

Care to share an example of what you'd like to do exactly?

xperiandri:
``` C#
private string AuthorizedTenant => httpContextAccessor.HttpContext.GetAuthorizedTenant();

protected IQueryable<TenantApplication> Applications
 => Context.Set<TenantApplication>()
               .Where(a => a.AllowedTenants.Contains(AuthorizedTenant));

```

xperiandri:
Ah, I see this requires a bit more changes than I expected

kevinchalet:
This won't work as you expect because...

  - the properties are of type `DbSet<T>`: you can't do filtering on sets, even using casts.
  - all the `Find` methods in the EF Core stores use compiled queries, that don't even use these properties (that I'll likely remove).

Your scenario is best achieved using global query filters: https://docs.microsoft.com/en-us/ef/core/querying/filters

xperiandri:
Indeed, thanks

xperiandri:
Do you know any good solution to apply Query Filters to JSON Array column?

kevinchalet:
EF Core has zero support for JSON columns and that's why we represent them as `string` properties and use manual filtering: https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictApplicationStore.cs#L336

What are you trying to do exactly?

xperiandri:
1. I have OpenIddictApplication with the AllowedTenants JSON array column.
I need to filter them by allowed tenants. And allow creating OpenIddictApplication with AllowedTenants containing a single item of a currently authorized tenant.

2. I have ApplicaitonUser with Tenant column and need to filter them by it and only allow to create a user with a currently authorized tenant.

xperiandri:
Well, looks like it is impossible to implement my scenario with filters.
Tenant info is retrieved from authorized claims principal. But `DbContext` is being created before authentication happens.

PR #638, opened 2018-07-05
Replaced new() constraints at the stores by Activator.CreateInstance

### LINKED ISSUE #631: Application store must not require OpenIddict application to be nonabstract and has parameterless constructor
You can apply this constraint to generic Descriptor and get the type from it.

Current implementation breaks my scenario where I want to have `WebApplication` and `ExternalApplication` classes derived from `abstract TenantApplication: OpenIddictApplication<Guid>`.
 Where WebApplicaton is used for in-house MVC app and ExternalApplication for third-party apps.

vmartynets:
fixes #631

### DIFF
--- a/src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
+++ b/src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
@@ -657,23 +657,13 @@ namespace OpenIddict.EntityFramework
                 return new ValueTask<TApplication>(Activator.CreateInstance<TApplication>());
             }
 
-            catch (MissingMemberException exception)
-            {
-                return new ValueTask<TApplication>(Task.FromException<TApplication>(
-                    new InvalidOperationException(new StringBuilder()
-                        .AppendLine("An error occurred while trying to create a new application instance.")
-                        .Append("Make sure that the application entity has a public parameterless constructor or create ")
-                        .Append("a custom store that overrides the 'InstantiateAsync()' method to use a custom factory.")
-                        .ToString(), exception)));
-            }
-
             catch (MemberAccessException exception)
             {
                 return new ValueTask<TApplication>(Task.FromException<TApplication>(
                     new InvalidOperationException(new StringBuilder()
                         .AppendLine("An error occurred while trying to create a new application instance.")
-                        .Append("Make sure that the application entity is not an abstract class or create a ")
-                        .Append("custom store that overrides the 'InstantiateAsync()' method to use a custom factory.")
+                        .Append("Make sure that the application entity is not abstract and has a public parameterless constructor ")
+                        .Append("or create a custom application store that overrides 'InstantiateAsync()' to use a custom factory.")
                         .ToString(), exception)));
             }
         }
--- a/src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
+++ b/src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
@@ -651,7 +651,16 @@ namespace OpenIddict.EntityFramework
         /// whose result returns the instantiated application, that can be persisted in the database.
         /// </returns>
         public virtual ValueTask<TApplication> InstantiateAsync(CancellationToken cancellationToken)
-            => new ValueTask<TApplication>(new TApplication());
+        {
+            try
+            {
+                return new ValueTask<TApplication>(Activator.CreateInstance<TApplication>());
+            }
+            catch (MissingMethodException)
+            {
+                throw new ArgumentException("You are trying to create OpenIddict application from application description in store of abstract application type or type that has no parameterless constructor. Register generic store and inject store typed of derived class for abstract base class.");
+            }
+        }
 
         /// <summary>
         /// Executes the specified query and returns all the corresponding elements.
--- a/src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
+++ b/src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
@@ -656,9 +656,12 @@ namespace OpenIddict.EntityFramework
             {
                 return new ValueTask<TApplication>(Activator.CreateInstance<TApplication>());
             }
-            catch (MissingMethodException)
+            catch (MissingMethodException exception)
             {
-                throw new ArgumentException("You are trying to create OpenIddict application from application description in store of abstract application type or type that has no parameterless constructor. Register generic store and inject store typed of derived class for abstract base class.");
+                throw new OpenIddictException(OpenIddictConstants.Exceptions.InstantiationError, new StringBuilder()
+                    .AppendLine("You are trying to create OpenIddict application from application description in store of abstract application type or type that has no parameterless constructor.")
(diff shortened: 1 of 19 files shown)


kevinchalet:
Thanks for your PR. I don't see the `new()` C# constraints being removed. Did you forget to push some commits?

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
```diff
@@ -651,7 +651,19 @@ public virtual ValueTask<ImmutableArray<string>> GetRedirectUrisAsync([NotNull]
...
         public virtual ValueTask<TApplication> InstantiateAsync(CancellationToken cancellationToken)
-            => new ValueTask<TApplication>(new TApplication());
+        {
+            try
+            {
+                return new ValueTask<TApplication>(Activator.CreateInstance<TApplication>());
+            }
+            catch (MissingMethodException exception)
+            {
+                throw new OpenIddictException(OpenIddictConstants.Exceptions.InstantiationError, new StringBuilder()
```
kevinchalet: It would be better to throw the exception "asynchronously". E.g:

```csharp
try
{
    return new ValueTask<TApplication>(Activator.CreateInstance<TApplication>());
}

catch (MissingMemberException exception)
{
    return new ValueTask<TApplication>(Task.FromException<TApplication>(
        new OpenIddictException(OpenIddictConstants.Exceptions.InstantiationError, new StringBuilder()
            .AppendLine("An error occurred while trying to create a new application instance.")
            .Append("Make sure that the application entity has a public parameterless constructor or create ")
            .Append("a custom store that overrides the 'InstantiateAsync()' method to use a custom factory.")
            .ToString(), exception)));
}

catch (MemberAccessException exception)
{
    return new ValueTask<TApplication>(Task.FromException<TApplication>(
        new OpenIddictException(OpenIddictConstants.Exceptions.InstantiationError, new StringBuilder()
            .AppendLine("An error occurred while trying to create a new application instance.")
            .Append("Make sure that the application entity is not an abstract class or create a ")
            .Append("custom store that overrides the 'InstantiateAsync()' method to use a custom factory.")
            .ToString(), exception)));
}
```

kevinchalet: BTW, I'm not sure it's worth using `OpenIddictException` here. `InvalidOperationException` might be better.

kevinchalet:
Note: there are a few others `new()` constraints in the code base. Consider doing a global search to eliminate them.

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
```diff
@@ -651,7 +651,32 @@ public virtual ValueTask<ImmutableArray<string>> GetRedirectUrisAsync([NotNull]
         /// whose result returns the instantiated application, that can be persisted in the database.
         /// </returns>
         public virtual ValueTask<TApplication> InstantiateAsync(CancellationToken cancellationToken)
-            => new ValueTask<TApplication>(new TApplication());
+        {
+            try
```
kevinchalet, line 655: Looks good. Please apply the same pattern to the other stores.

xperiandri: I guess methods can be made async to simplify the logic.

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
```diff
@@ -651,7 +651,32 @@ public virtual ValueTask<ImmutableArray<string>> GetRedirectUrisAsync([NotNull]
...
+            {
+                return new ValueTask<TApplication>(Activator.CreateInstance<TApplication>());
+            }
+
+            catch (MissingMemberException exception)
+            {
+                return new ValueTask<TApplication>(Task.FromException<TApplication>(
```
xperiandri: Let's move this duplicated code to an internal method

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
kevinchalet: Thing is this implementation doesn't use `await` so we'd get a compiler warning and pay for a state machine that wouldn't be useful. Let's keep it as-is.

src/OpenIddict.EntityFramework/Stores/OpenIddictApplicationStore.cs
kevinchalet: Don't worry about that in this PR. We'll likely refactor that when/if we move to .resx-based exception messages.

kevinchalet:
@xperiandri once this PR is ready, could you please test it on your end to ensure it works with your scenario?

xperiandri:
@PinpointTownes, sure

xperiandri (commented):
What about the message that developer can register a generic store and resolve it with a class derived from abstract?

kevinchalet:
> What about the message that developer can register a generic store and resolve it with a class derived from abstract?

It's something that will work with the MongoDB stores but **not with the Entity Framework 6/Core** stores due to the inter-dependent constraints at the stores level between `TApplication/TAuthorization/TScope/TToken`.

I don't think it's a big deal as `InstantiateAsync()` is actually only used by the `CreateAsync()` overloads that take a descriptor instead of the entity directly. These overloads are typically used by libraries or components that can't create the entity themselves (e.g the OpenIddict server components for authorizations and tokens or OrchardCore for all the entities, as the OpenID module is entity-agnostic):

```csharp
var application = new MyAppDerived1()
{
    // ...
};

// manager is of type OpenIddictApplicationManager<MyApp>().
await manager.CreateAsync(application, "secret");
```

... or if you want to fill an app from a descriptor:

```csharp
var application = new MyAppDerived1();

// manager is of type OpenIddictApplicationManager<MyApp>().
await manager.PopulateAsync(application, descriptor);
await manager.CreateAsync(application, "secret");
```

kevinchalet:
@xperiandri let me know when you have a chance to test it.

xperiandri:
You no longer use MyGet CI feed, do you?

kevinchalet:
We do (but we don't push PR packages to the MyGet feed, as it would be a terrible security flaw :sweat_smile:)

Consider building the .nukpg locally and referencing them in your project.

kevinchalet:
@xperiandri did that work? :smile:

kevinchalet:
I tested it on my side and it looks like it works as expected with your scenario so I'll merge this PR later today.

xperiandri:
Good

kevinchalet:
@xperiandri did you have a chance to test it too?

xperiandri:
Not the way I want. If you can wait till tomorrow I will test today

kevinchalet:
Okay, I'll wait a bit then.

kevinchalet:
@xperiandri ping.

xperiandri:
Still checking.
I will let you know soon

xperiandri:
Yes, it works as expected

kevinchalet:
@Amialc thanks for your contribution :clap:  
@xperiandri thanks for testing :+1:

PR #709, opened 2018-12-27
Add validation log message for applications without logout permission

### LINKED ISSUE #687: Use more meaningful message for ValidatePostLogoutRedirectUriAsync
@PinpointTownes thanks for amazing library.

Today I tried to logout using post_logout_redirect_uri parameter and got the following message:

> The specified 'post_logout_redirect_uri' parameter is not valid.

but after review the source code, I figure out that the actual reason was my client permissions. I had to simply add logout permission:

`OpenIddictConstants.Permissions.Endpoints.Logout`

Is it possible for ValidatePostLogoutRedirectUriAsync function throw more meaningful exception? Thank you!

ahanoff:
Close #687

### DIFF
--- a/src/OpenIddict.Server/Internal/OpenIddictServerProvider.Session.cs
+++ b/src/OpenIddict.Server/Internal/OpenIddictServerProvider.Session.cs
@@ -120,8 +120,6 @@ namespace OpenIddict.Server.Internal
                     var applications = await _applicationManager.FindByPostLogoutRedirectUriAsync(address);
                     if (applications.IsDefaultOrEmpty)
                     {
-                        _logger.LogError("The logout request was rejected because the specified post_logout_redirect_uri " +
-                                         "was unknown: {address}.", address);
                         return false;
                     }
 
@@ -138,13 +136,13 @@ namespace OpenIddict.Server.Internal
                             return true;
                         }
                     }
-
-                    _logger.LogError("The logout request was rejected because the specified application has no logout permission: {address}.", address);
                     return false;
                 }
 
                 if (!await ValidatePostLogoutRedirectUriAsync(context.PostLogoutRedirectUri))
                 {
+                    _logger.LogError("The logout request was rejected because no application with the specified post_logout_redirect_uri " +
+                                     "and with a logout endpoint permission was found: {PostLogoutRedirectUri}.", context.PostLogoutRedirectUri);
                     context.Reject(
                         error: OpenIddictConstants.Errors.InvalidRequest,
                         description: "The specified 'post_logout_redirect_uri' parameter is not valid.");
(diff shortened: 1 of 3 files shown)


test/OpenIddict.Server.Tests/Internal/OpenIddictServerProviderTests.cs
```diff
@@ -1525,7 +1525,7 @@ public async Task ProcessSignoutResponse_CustomPublicParametersAreAddedToLogoutR
             Assert.Equal("value", (string) response["custom_string_parameter"]);
         }
 
-        private static TestServer CreateAuthorizationServer(Action<OpenIddictServerBuilder> configuration = null)
+        private static TestServer CreateAuthorizationServer(Action<OpenIddictServerBuilder> configuration = null, bool ignoreEndpointPermission = true)
```
ahanoff: I don't really like this field injection, but I didn't find better solution for now. Any suggestion @PinpointTownes ? Thanks

kinosang: @ahanoff why not use `OpenIddictServerOptions.IgnoreEndpointPermissions()`

ahanoff: @kinosang so I refer to [OpenIddictServerProvider.Session.cs](https://github.com/openiddict/openiddict-core/blob/42d0cd3be39ade88c2714c4cacd8d1f558419780/src/OpenIddict.Server/Internal/OpenIddictServerProvider.Session.cs#L128-L131). Tests by default ignore endpoint permissions, but in that case app permission validation code is unreachable for tests

src/OpenIddict.Server/Internal/OpenIddictServerProvider.Session.cs
```diff
@@ -137,14 +139,12 @@ async Task<bool> ValidatePostLogoutRedirectUriAsync(string address)
                         }
                     }
 
+                    _logger.LogError("The logout request was rejected because the specified application has no logout permission: {address}.", address);
```
kevinchalet: At this point, it's not really "exact" to say "specified application" as end session requests don't include a `client_id` (the specification clearly sucks in this regard).

Shouldn't we keep the log message at the original place (l. 145/147) but reword it a bit to explain why it's unknown? (i.e no application with the specified `post_logout_redirect_uri` AND with a logout endpoint permission endpoint was found)

test/OpenIddict.Server.Tests/Internal/OpenIddictServerProviderTests.cs
kevinchalet: @ahanoff a simple (and quite elegant) option consists in directly configuring the underlying `OpenIddictServerOptions` using the `Configure` method (which is what `OpenIddictServerBuilder` does for you):

```csharp
builder.Configure(options => options.IgnoreEndpointPermissions = false);
```

kevinchalet:
@ahanoff thanks for your contribution! :clap: 
@kinosang thanks for taking a look :+1:

PR #797, opened 2019-09-12
Prevent Arcade from replacing the project copyright by Microsoft's copyright

kevinchalet:
@martincostello FYI the aspnet-contrib providers are also impacted, so we may want to adopt a similar fix there 😄 

image


martincostello:
@PinpointTownes Good catch! I’ll fix up OAuth and OpenId repos tomorrow.

PR #831, opened 2019-10-15
add filter for RequireDegradedModeDisabled to RevokeToken handler

knoxi:
An exception in revoke token handler occurs when trying to revoke a token with enabled degraded mode. Adding filter for degraded mode fixes it.

/cc @marc-mueller


kevinchalet:
Nice catch, thanks for the PR!

Merged: https://github.com/openiddict/openiddict-core/commit/ff7d5d56107403743450c2616de3e4d8c7f0a122.

PR #851, opened 2019-12-01
Port the AspNet.Security.OpenIdConnect.Extensions tests

### LINKED ISSUE #774: Port the AspNet.Security.OpenIdConnect.Extensions tests
In 1.x/2.x, OpenIddict relied on a bunch of extensions provided by ASOS as part of the `AspNet.Security.OpenIdConnect.Extensions` package.

When decoupling OpenIddict from ASP.NET Core, most of these extensions have been updated to act on `ClaimsPrincipal` instead of `AuthenticationTicket` (which means what used to be stored in authentication properties is now stored as claims)

We'll need to port the corresponding unit tests: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/dev/test/AspNet.Security.OpenIdConnect.Extensions.Tests/OpenIdConnectExtensionsTests.cs

IgorHrabrov:
Resolved #774

### DIFF
--- a/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
@@ -812,7 +812,7 @@ namespace OpenIddict.Abstractions
                 throw new ArgumentException("The claim type cannot be null or empty.", nameof(type));
             }
 
-            return principal.FindFirst(claim => string.Equals(claim.Type, type))?.Value;
+            return principal.FindFirst(type)?.Value;
         }
 
         /// <summary>
@@ -1791,15 +1791,6 @@ namespace OpenIddict.Abstractions
         /// <returns>The claims principal.</returns>
         public static ClaimsPrincipal SetInternalTokenId([NotNull] this ClaimsPrincipal principal, string identifier)
             => principal.SetClaim(Claims.Private.TokenId, identifier);
-
-        /// <summary>
-        /// Sets the token usage associated with the claims principal.
-        /// </summary>
-        /// <param name="principal">The claims principal.</param>
-        /// <param name="tokenUsage">The token usage to store.</param>
-        /// <returns>The claims principal.</returns>
-        public static ClaimsPrincipal SetTokenUsage([NotNull] this ClaimsPrincipal principal, string tokenUsage)
-             => principal.SetClaim(Claims.Private.TokenUsage, tokenUsage);
 
         private static IEnumerable<string> GetValues(string source, char[] separators)
         {
--- a/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
@@ -812,7 +812,7 @@ namespace OpenIddict.Abstractions
                 throw new ArgumentException("The claim type cannot be null or empty.", nameof(type));
             }
 
-            return principal.FindFirst(type)?.Value;
+            return principal.FindFirst(claim => string.Equals(claim.Type, type))?.Value;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
```diff
@@ -812,7 +812,7 @@ public static string GetClaim([NotNull] this ClaimsPrincipal principal, [NotNull
                 throw new ArgumentException("The claim type cannot be null or empty.", nameof(type));
             }
 
-            return principal.FindFirst(type)?.Value;
+            return principal.FindFirst(claim => string.Equals(claim.Type, type))?.Value;
```
kevinchalet: Is there any particular reason to be case-sensitive here (instead of just using `FirstFirst(type)`, that performs a case-insensitive comparison)?

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
```diff
@@ -1798,6 +1792,15 @@ public static ClaimsPrincipal SetInternalAuthorizationId([NotNull] this ClaimsPr
         public static ClaimsPrincipal SetInternalTokenId([NotNull] this ClaimsPrincipal principal, string identifier)
             => principal.SetClaim(Claims.Private.TokenId, identifier);
 
+        /// <summary>
+        /// Sets the token usage associated with the claims principal.
+        /// </summary>
+        /// <param name="principal">The claims principal.</param>
+        /// <param name="tokenUsage">The token usage to store.</param>
+        /// <returns>The claims principal.</returns>
+        public static ClaimsPrincipal SetTokenUsage([NotNull] this ClaimsPrincipal principal, string tokenUsage)
```
kevinchalet: I'm tempted to say adding this extension is not super useful, but I'm all ears if you have any particular scenario in mind.

kevinchalet:
Thanks for your PR!

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
IgorHrabrov: There is a test for this ( GetClaim_IsCaseSensitive() ) in old project - that's why I changed testing code. If it should be case-insensitive now - let's remove this test method and revert to "FindFirst(type)"

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
IgorHrabrov: I've just added this for consistency since we have Get*** method for this but don't have Set*** method. And again there is a test for Set*** method in old project. If it's not needed in new version - let's remove it.

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
kevinchalet: Oh yeah, good point. Can you please revert it and rename the test to `GetClaim_IsCaseInsensitive`?

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
kevinchalet: Makes sense. Since it's likely not very useful, let's remove it.

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
IgorHrabrov: Done

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
IgorHrabrov: Done

kevinchalet (approved):

kevinchalet:
Merged, thanks again!

PR #903, opened 2020-01-20
Adds a check for token requests containing a code_verifier parameter.

ajfleming1:
Adds a check for a token request containing code_verifier when the authorization request was not sent code_challenge and code_challenge_method.


kevinchalet:
Great, thanks! Could you please add an integration test [here](https://github.com/openiddict/openiddict-core/blob/dev/test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTests.Exchange.cs#L564) to ensure this scenario is covered?

ajfleming1:
Done - let me know if there is anything else I can do to help out. :)

kevinchalet:
Merged, thanks for your contribution! 👏 

If you're interested in contributing to other parts of OpenIddict, please don't hesitate (tickets that don't require much design are usually flagged as "help wanted" so it's easier to find them: https://github.com/openiddict/openiddict-core/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)

Issue #807, opened 2019-09-27
Add unit tests for the new OpenIddict.Validation, OpenIddict.Validation.AspNetCore, OpenIddict.Validation.Owin, OpenIddict.Validation.DataProtection and OpenIddict.Validation.ServerIntegration handlers


kevinchalet:
Starting with 3.0, most of the validation logic is now contained in implementations of `IOpenIddictValidationHandler<TContext>`, that can be easily unit-tested without using `TestServer`. As part of the 3.0 work, we need to add unit tests for all these handlers.

gustavdw:
I've started on this one, working on the validation handlers at the moment.

kevinchalet:
@gustavdw excellent, thanks for contributing! 👏

kevinchalet:
@gustavdw don't hesitate to ping me if you have any question or need help.

Issue #894, opened 2020-01-14
Port the OpenIddict.Server unit tests from OpenIddict 2.x


kevinchalet:
The following classes will need to be adapted from 2.x:
  - https://github.com/openiddict/openiddict-core/blob/2.0.1/test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
  - https://github.com/openiddict/openiddict-core/blob/2.0.1/test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
  - https://github.com/openiddict/openiddict-core/blob/2.0.1/test/OpenIddict.Server.Tests/Internal/OpenIddictServerConfigurationTests.cs (for that one, we should remove the dependency on `TestServer` and make it a proper unit tests class).

Darthruneis:
@kevinchalet I'm considering taking a look at this one based on our discussion in #956 - can you clarify specifically what you mean when you say 'Port' the tests? My best guess is to fix them/update them after they have broken during changes for 3.0, is that correct?

kevinchalet:
Thanks @Darthruneis!

> My best guess is to fix them/update them after they have broken during changes for 3.0, is that correct?

Yep, exactly. There are also new methods that were added to `OpenIddictServerBuilder` in 3.0 for which we'll need to add new unit tests after porting the tests class from 2.x (but it can be done in a separate PR).

Issue #761, opened 2019-08-16
Add unit tests for the new OpenIddict.Server, OpenIddict.Server.AspNetCore, OpenIddict.Server.Owin and OpenIddict.Server.DataProtection handlers


kevinchalet:
Starting with 3.0, most of the server logic is now contained in implementations of `IOpenIddictServerHandler<TContext>`, that can be easily unit-tested without using `TestServer`. As part of the 3.0 work, we need to add unit tests for all these handlers.

kevinchalet:
**OpenIddict.Server**:

  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.Authentication.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.Discovery.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.Exchange.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.Introspection.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.Session.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.Userinfo.cs

**OpenIddict.Server.AspNetCore**:
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.Authentication.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.Session.cs

Once the `OpenIddict.Server.AspNetCore` tests are ready, we'll be able to reuse most of them for `OpenIddict.Server.Owin`.

PR #956, opened 2020-05-30
807 | OpenIddict.Validation.AspNetCore : InferIssuerFromHost tests

Darthruneis:
Taking a stab at writing some tests for #807 . 

I wanted to get an initial test class written and see if there's any feedback before continuing on to more of the handlers inside of `OpenIddict.Validation.AspNetCore`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
   global:
     - DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     - DOTNET_CLI_TELEMETRY_OPTOUT: 1
-    - VERBOSE: true
 os:
   - linux
   - osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
   global:
     - DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     - DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    - VERBOSE: true
 os:
   - linux
   - osx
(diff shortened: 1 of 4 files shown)


kevinchalet:
Thanks for your PR, @Darthruneis! :clap:

Here's my feeling: when I opened https://github.com/openiddict/openiddict-core/issues/807 last year (and its server counterpart), I didn't anticipate how big - and time-consuming - 3.0 would become.

I ported the server integration tests and it already took longer than I expected (and while it's essential, writing/porting tests is rarely a fun task 😅). Given how time-consuming this can be, I'm wondering if adding unit tests for the handlers is the right thing to do, as most of the logic they would test would be also covered by the integration tests. Of course, contributions like yours definitely help, but what's unclear is if contributions will cover all the handlers (I'm not so sure), or if I'll have to dedicate time to that to fill in the blanks eventually.

I'd love to hear your thoughts.

Darthruneis:
@kevinchalet I can't say I'm any good at predicting the future, and thus I can't say how much test coverage will be supplied by contributions. That said, I think having unit tests in addition to integration tests is a nice to have, and will generally cover the units more thoroughly than the integration tests will. Integration and unit tests have different goals, after all. 

The tests in this PR, for example. I wouldn't expect to see integration tests around some of the edge cases, such as the null checks. That seems like it would be more time intensive to set up in an integration test than it would in a unit test. Additionally, I can see it becoming a pain point for maintenance, as integration tests tend to be more brittle than unit tests.

Finally, I'm hoping that by starting with unit testing various pieces of code, I'll get up to speed with the repo as a whole, and eventually might be able to contribute in other ways. 

I'm quite excited by the ASOS merge that 3.0 is aiming to do, I'd like to help in some way if I can, even if its just unit testing for now :).

kevinchalet:
> I'm quite excited by the ASOS merge that 3.0 is aiming to do, I'd like to help in some way if I can, even if its just unit testing for now :).

Thanks a lot for your interest! 😃 

I agree with everything you said. Sadly, I have limited bandwidth these days, and while having unit tests for all the handlers would be valuable, I think focusing on the integration tests first will be a better option: they are indeed less fine-grained and won't really cover corner cases, but they help guarantee that the handlers chain forms a coherent pipeline that reacts exactly how it should, which is not something we can test with per-handler unit tests.

I'll remove https://github.com/openiddict/openiddict-core/issues/807 and https://github.com/openiddict/openiddict-core/issues/761 from the next milestone so we can focus on the other tasks first (including porting the validation/introspection `TestServer`-based tests).

If you're still interested in contributing to the unit tests, maybe you could take a look at https://github.com/openiddict/openiddict-core/issues/894?

Thanks again!

kevinchalet:
Closing for now. We'll reconsider later if we have contributions for this area 😃

PR #970, opened 2020-06-09
Begin porting OpenIddictServerBuilderTests from 2.x.

Darthruneis:
- Some tests would not work without the changes to how the Options were retrieved.
- The skipped tests are those that are currently failing on all frameworks.
Will be continuing to try to fix those.
- The commented out tests have compiler errors.
Will need to go through and see which, if any, are still valid.

Related Work Items: #894

Posting this in a somewhat WIP state in case @kevinchalet has any feedback on the changes so far, and also to see what happens with the CI builds.


test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net461;net472;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Moq" Version="4.13.1" />
```
kevinchalet: Please use `$(MoqVersion)` to avoid hardcoding the Moq version.

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
```
kevinchalet: This dependency is automagically added by Arcade and specifying it explicitly causes the exception seen in the CI logs.

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\OpenIddict.Abstractions\OpenIddict.Abstractions.csproj" />
```
kevinchalet: This package is referenced by `OpenIddict.Server` so you should be able to remove it.

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <ProjectReference Include="..\..\src\OpenIddict.Server\OpenIddict.Server.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
+    <Reference Include="System.Net.Http" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.0" />
```
kevinchalet: Once you replace `3.1.0` by `$(ExtensionsVersion)`, you'll be able to remove most of these `ItemGroup`s.

kevinchalet (commented):
Thanks for your PR! I added a couple of notes that should help you with dependencies.

Darthruneis:
@kevinchalet One of the tests I was having trouble with were the certificate ones. What is the process used for generating the certificates used in the test projects? Is it just a self-signed cert created locally?

Reviewing your other comments now.

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
Darthruneis: Ah, nice. Updated!

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
Darthruneis: Yeah, good call. I added the Abstractions reference first while going through the 'fix all the red squigglies from namespaces that could not be found'. Didn't think about it being a transitive dependency.

Updated.

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
Darthruneis: Removed. I swore I had seen another of the Test projects including it, but looking again I can't. Maybe it was because it showed up in the NuGet Package Manager window in VS? Either way, removed.

test/OpenIddict.Server.Tests/OpenIddict.Server.Tests.csproj
Darthruneis: Good to know, updated!

kevinchalet:
> @kevinchalet One of the tests I was having trouble with were the certificate ones. What is the process used for generating the certificates used in the test projects? Is it just a self-signed cert created locally?

The certificates stored as embedded resources, they are indeed self-signed certificates.

Assuming you're referring to the 2 skipped tests - that use certificate generation - I suspect that's because the `SUPPORTS_CERTIFICATE_GENERATION` compilation constant is not set. You'll need to set if for all TFMs except `net461` (e.g https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddict.Server.csproj#L33).

Darthruneis:
Yeah, I've got that line in there, but currently I don't have a certificate in the project. I was initially thinking I'd need to generate a new one, but I guess it should also be possible to just go back and get the one that was previously in the test project back in 2.x. That might be easier/less prone to new issues popping up in the test.

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
```diff
@@ -0,0 +1,2000 @@
...
+            // Arrange
+            var services = (IServiceCollection)null;
+
+            // Act and assert
+            var exception = Assert.Throws<ArgumentNullException>(() => new OpenIddictServerBuilder(services));
+
+            Assert.Equal("services", exception.ParamName);
+        }
+
+        private class CustomContext : BaseContext
```
kevinchalet: nit: can you please put them at the end of the file?

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
```diff
@@ -0,0 +1,2000 @@
...
+        {
+            /// <summary>
```
kevinchalet: You can remove the XML comments, they are not very useful in this case.

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
```diff
@@ -0,0 +1,2000 @@
...
+        [Fact]
+        public void AddEventHandler_ThrowsAnExceptionWhenConfigurationIsNull()
+        {
+            // Arrange
+            var services = CreateServices();
+            var builder = CreateBuilder(services);
+            Action<OpenIddictServerHandlerDescriptor.Builder<CustomContext>> configuration = null;
+
+            // Act
+            Action action = () => builder.AddEventHandler<CustomContext>(configuration);
```
kevinchalet: Technically, the exception is not thrown at this stage, so it's weird to put it in the "act" phase. Could you please combine the act and assert into a single "Act and assert" phase?

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
```diff
@@ -0,0 +1,2000 @@
...
+            }
+        }
+
+        [Fact]
+        public void AddEventHandler_ThrowsAnExceptionWhenConfigurationIsNull()
+        {
+            // Arrange
+            var services = CreateServices();
+            var builder = CreateBuilder(services);
+            Action<OpenIddictServerHandlerDescriptor.Builder<CustomContext>> configuration = null;
```
kevinchalet, line 36: To simply things, feel free to remove that and do `builder.AddEventHandler<CustomContext>(configuration: null)` like in the other tests.

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
```diff
@@ -0,0 +1,2000 @@
...
+
+            // Act
+            Action action = () => builder.AddEncryptionKey(key);
+
+            // Assert
+            var exception = Assert.Throws<ArgumentNullException>(action);
+            Assert.Equal(nameof(key), exception.ParamName);
+        }
+
+        private class FakeAsymmetricSecurityKey : AsymmetricSecurityKey
```
kevinchalet: May I suggest using Moq here? I used it in the ASOS tests and it worked fine in this case: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/dev/test/AspNet.Security.OpenIdConnect.Server.Tests/OpenIdConnectServerExtensionsTests.cs#L530

kevinchalet (commented):
It looks excellent, thanks a lot!

Do you need help with the remaining tests?

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
Darthruneis: Resharper generated those when I used its 'generate missing members' functionality. I can remove them if you like, but it didn't take any time to add them either :).

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
Darthruneis: I used the variable so that I could use `nameof(configuration)` in the assertion, rather than hardcoding the string value for the assert. I avoid hardcoding strings when possible - do you think there's a negative to using `nameof` in this way in the context of a unit test?

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
Darthruneis: Done.

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
Darthruneis: Updated all cases where this was being done in this way.

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
Darthruneis: Verified tests still pass using Moq, so replaced with Moq.

Darthruneis:
> Do you need help with the remaining tests?

I think I've got most of the coverage of the ServerBuilder class now. Anything related to the certificates or encryption may need to be built out a bit more. I've shied away those due to a lack of experience working with certs/encryption.

Regarding the other classes which are referenced in #894 , I intend to start on those next. Do you have any preference between:
- A different PR for each of the class that is being tested, or 
- a single PR (this one) to encompass all of 894?

test/OpenIddict.Server.Tests/OpenIddictServerBuilderTests.cs
kevinchalet: It's certainly not a bad practice, it's just that I haven't used this approach anywhere else, so it's a bit inconsistent. But I'm fine keeping it this way if you like it.

kevinchalet:
FYI, the fact you're using `action` delegate variables results in VS adding many "info" messages suggesting replacing them by local functions:

image

This is not blocking for this PR, but to reduce the noise in VS, you'll want probably want to avoid the intermediate variable and do `Assert.Throws<ArgumentNullException>(() => builder.AddEncryptionKey(key))`.

> I think I've got most of the coverage of the ServerBuilder class now. Anything related to the certificates or encryption may need to be built out a bit more. I've shied away those due to a lack of experience working with certs/encryption.

Great. I'll take a look once this PR is merged.

> Regarding the other classes which are referenced in #894 , I intend to start on those next. Do you have any preference between.

No preference. Feel free to opt for the one who better suits your flow. If you opt for the second option, please let me know and I'll merge this PR today.

Thanks!

Darthruneis:
Updated to clean up those info messages. I've had info messages turned off in VS for ages so I had no idea it was polluting those up so much.

As for the PR, I prefer to merge it in then. Easier to make a new branch then and work on the next set of tests that need ported, and it allows the tests to be in place as the safety net they hopefully will act as.

kevinchalet:
PR merged, thanks a lot for your contribution! 👏

PR #1048, opened 2020-08-03
894: Port the 2.x tests for OpenIddictServerExtensions for 3.x.

Darthruneis:
- Add tests similar to ported tests for the current set of services which are configured by `AddServer()`.
- Add tests for the various default handlers that are configured.
This could be done with less code, but the test names when doing so are not as useful as the ones
generated in this manner.

### DIFF
--- a/test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
+++ b/test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
@@ -1,0 +1,209 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/openiddict/openiddict-core for more information concerning
+ * the license and the contributors participating to this project.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace OpenIddict.Server.Tests
+{
+    public class OpenIddictServerExtensionsTests
+    {
+        [Fact]
+        public void AddServer_ThrowsAnExceptionForNullBuilder()
+        {
+            // Arrange
+            var builder = (OpenIddictBuilder) null!;
+
+            // Act and assert
+            var exception = Assert.Throws<ArgumentNullException>(() => builder.AddServer());
+
+            Assert.Equal("builder", exception.ParamName);
+        }
+
+        [Fact]
+        public void AddServer_ThrowsAnExceptionForNullConfiguration()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act and assert
+            var exception = Assert.Throws<ArgumentNullException>(() => builder.AddServer(configuration: null!));
+
+            Assert.Equal("configuration", exception.ParamName);
+        }
+
+        [Fact]
+        public void AddServer_RegistersLoggingServices()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.ServiceType == typeof(ILogger<>));
+        }
+
+        [Fact]
+        public void AddServer_RegistersOptionsServices()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.ServiceType == typeof(IOptions<>));
+        }
+        
+        [Fact]
+        public void AddServer_RegistersServerDispatcher()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.ServiceType == typeof(IOpenIddictServerDispatcher) &&
+                                                 service.ImplementationType == typeof(OpenIddictServerDispatcher) &&
+                                                 service.Lifetime == ServiceLifetime.Scoped);
+        }
+        
+        [Fact]
+        public void AddServer_RegistersServerFactory()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.ServiceType == typeof(IOpenIddictServerFactory) &&
+                                                 service.ImplementationType == typeof(OpenIddictServerFactory) &&
+                                                 service.Lifetime == ServiceLifetime.Scoped);
+        }
+
+        public static IEnumerable<object[]> DefaultHandlers
+            => OpenIddictServerHandlers.DefaultHandlers.Select(descriptor => new object[] { descriptor });
+        
+        [Theory]
+        [MemberData(nameof(DefaultHandlers))]
+        public void AddServer_RegistersDefaultHandler(OpenIddictServerHandlerDescriptor descriptor)
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.Lifetime == descriptor.ServiceDescriptor.Lifetime &&
+                                                 service.ServiceType == descriptor.ServiceDescriptor.ServiceType &&
+                                                 service.ImplementationType == descriptor.ServiceDescriptor.ImplementationType);
+        }
+
+        [Theory]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireAuthorizationStorageEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireAuthorizationRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireClientIdParameter))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireConfigurationRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireCryptographyRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireDegradedModeDisabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireDeviceRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireEndpointPermissionsEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireGrantTypePermissionsEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireIntrospectionRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireLogoutRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequirePostLogoutRedirectUriParameter))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireReferenceAccessTokensEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireReferenceRefreshTokensEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireRevocationRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireSlidingRefreshTokenExpirationEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireScopePermissionsEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireScopeValidationEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireTokenStorageEnabled))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireTokenRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireUserinfoRequest))]
+        [InlineData(typeof(OpenIddictServerHandlerFilters.RequireVerificationRequest))]
+        public void AddServer_RegistersRequiredSingletons(Type type)
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var builder = new OpenIddictBuilder(services);
+
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.ServiceType == type &&
+                                                 service.ImplementationType == type &&
+                                                 service.Lifetime == ServiceLifetime.Singleton);
+        }
+
+        [Fact]
+        public void AddServer_ResolvingProviderThrowsAnExceptionWhenCoreServicesAreNotRegistered()
+        {
+            // Arrange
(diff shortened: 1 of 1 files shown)


test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.TestCases.cs
```diff
@@ -0,0 +1,417 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OpenIddict.Server.Tests
+{
+    public partial class OpenIddictServerExtensionsTests
+    {
+        private static IEnumerable<AddServer_RegistersDefaultHandler_TestCase> GetOpenIdDictServerHandlerRegistrationTestCases()
```
kevinchalet: I get the intent, but I suspect maintaining these test cases will be painful in the long term.

Instead, couldn't we just make sure all the `OpenIddictServerHandlerDescriptor`s contained in `OpenIddictServerHandlers.DefaultHandlers` array have their `ServiceDescriptor` registered by `AddServer()`? It would make all this stuff unnecessary for your `AddServer_RegistersDefaultHandler` test.

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
```diff
@@ -0,0 +1,403 @@
...
+            var builder = new OpenIddictBuilder(services);
+
+            // Act and assert
+            var exception = Assert.Throws<ArgumentNullException>(() => builder.AddServer(configuration: null));
+
+            Assert.Equal("configuration", exception.ParamName);
+        }
+
+        //[Fact]
+        //public void AddServer_RegistersAuthenticationServices()
```
kevinchalet: FYI, the authentication services are no longer registered by the generic server services (aka `AddServer()`) but by the ASP.NET Core host itself for layering reasons (`options.UseAspNetCore()`).

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
```diff
@@ -0,0 +1,403 @@
...
+            var provider = services.BuildServiceProvider();
+
+            var exception = Assert.Throws<InvalidOperationException>(() => provider.GetRequiredService<OpenIddictServerConfiguration>());
+
+            Assert.NotNull(exception);
+        }
+
+        [Theory]
+        [InlineData(typeof(IPostConfigureOptions<OpenIddictServerOptions>), typeof(OpenIddictServerConfiguration))]
+        //[InlineData(typeof(IPostConfigureOptions<OpenIddictServerOptions>), typeof(OpenIdConnectServerInitializer))]
```
kevinchalet: You can remove that.

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
```diff
@@ -0,0 +1,403 @@
...
+            // Act
+            builder.AddServer();
+
+            // Assert
+            Assert.Contains(services, service => service.ServiceType == serviceType &&
+                                                 service.ImplementationType == implementationType);
+        }
+
+        //[Fact]
+        //public void AddServer_RegistersAuthenticationScheme()
```
kevinchalet: It's no longer something the generic server services or the ASP.NET Core host do in 3.0: the `OpenIddictServerAspNetCoreHandler` is now lazily registered by `OpenIddictServerAspNetCoreConfiguration`.

kevinchalet (commented):
Thanks for your PR!

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.TestCases.cs
Darthruneis: So, my initial concern was that if you had a test failure come up, the test name would be unhelpful in tracking down what registration was missing - this was particularly true for types which end up being a generic.

One possible way these tests could be rewritten:
```csharp
public static IEnumerable<object[]> GetAddServer_RegistersDefaultHandler_V2TestCases()
{
	foreach (var descriptor in OpenIddictServerHandlers.DefaultHandlers)
	{
		yield return new object[] { descriptor };
	}
}

[Theory]
[MemberData(nameof(GetAddServer_RegistersDefaultHandler_V2TestCases))]
public void AddServer_RegistersDefaultHandler_v2(OpenIddictServerHandlerDescriptor descriptor)
{
	//Arrange
	var services = new ServiceCollection();
	var builder = new OpenIddictBuilder(services);

	//Act
	builder.AddServer();

	//Assert
	Assert.Contains(services, service => service.Lifetime == descriptor.ServiceDescriptor.Lifetime &&
			service.ServiceType == descriptor.ServiceDescriptor.ServiceType &&
			service.ImplementationType == descriptor.ServiceDescriptor.ImplementationType);
}
```
 Here's an example output shown for a failure (I commented out the line in AddServer that registers all the default handlers):
> AddServer_RegistersDefaultHandler_v2(descriptor: OpenIddictServerHandlerDescriptor { ContextType = typeof(OpenIddict.Server.OpenIddictServerEvents+ApplyAuthorizationResponseContext), FilterTypes = [], Order = -2147381648, ServiceDescriptor = ServiceType: OpenIddict.Server.OpenIddictServerHandlers+Authentication+AttachResponseState Lifetime: Singleton ImplementationType: OpenIddict.Server.OpenIddictServerHandlers+Authentication+AttachResponseState, Type = BuiltIn }) [0:00.000] Failed: Assert.Contains() Failure
> Assert.Contains() Failure
> Not found: (filter expression)
> In value:  ServiceCollection [ServiceType: Microsoft.Extensions.Options.IOptions`1[TOptions] Lifetime: Singleton ImplementationType: Microsoft.Extensions.Options.OptionsManager`1[TOptions], ServiceType: Microsoft.Extensions.Options.IOptionsSnapshot`1[TOptions] Lifetime: Scoped ImplementationType: Microsoft.Extensions.Options.OptionsManager`1[TOptions], ServiceType: Microsoft.Extensions.Options.IOptionsMonitor`1[TOptions] Lifetime: Singleton ImplementationType: Microsoft.Extensions.Options.OptionsMonitor`1[TOptions], ServiceType: Microsoft.Extensions.Options.IOptionsFactory`1[TOptions] Lifetime: Transient ImplementationType: Microsoft.Extensions.Options.OptionsFactory`1[TOptions], ServiceType: Microsoft.Extensions.Options.IOptionsMonitorCache`1[TOptions] Lifetime: Singleton ImplementationType: Microsoft.Extensions.Options.OptionsCache`1[TOptions], ...]
>    at OpenIddict.Server.Tests.OpenIddictServerExtensionsTests.AddServer_RegistersDefaultHandler_v2(OpenIddictServerHandlerDescriptor descriptor) in ...\test\OpenIddict.Server.Tests\OpenIddictServerExtensionsTests.cs:line 215

Another downside of this approach, is that if a change were to happen where one of the handlers in question was removed from the DefaultHandlers array, no test would fail, as that test case would be removed.

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
Darthruneis: Alright - I have removed that commented out test. There are a few more that I was not sure about - are the remaining commented out tests also able to be removed completely?

The list of tests that are still commented out is:
- `AddServer_RegistersCachingServices()`
- `AddServer_RegistersEventService()`
- `AddServer_RegistersHandler()`
- `AddServer_RegistersProvider()`
- `AddServer_ThrowsAnExceptionWhenSchemeIsAlreadyRegisteredWithDifferentHandlerType()`

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
Darthruneis: Done!

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
Darthruneis: Alright, removed!

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.TestCases.cs
kevinchalet: > One possible way these tests could be rewritten:

I like this approach 👍 

> Another downside of this approach, is that if a change were to happen where one of the handlers in question was removed from the DefaultHandlers array, no test would fail, as that test case would be removed.

I'd expect this case to be caught by the integration tests, where a missing or invalid registration should result in an invalid pipeline.

test/OpenIddict.Server.Tests/OpenIddictServerExtensionsTests.cs
kevinchalet: > `AddServer_RegistersEventService()`
> `AddServer_RegistersProvider()`

The event service was replaced by `IOpenIddictServerDispatcher` and `IOpenIddictServerFactory`.

> `AddServer_RegistersHandler()`
> `AddServer_ThrowsAnExceptionWhenSchemeIsAlreadyRegisteredWithDifferentHandlerType()`

The authentication handler is now registered by the ASP.NET Core host in `OpenIddictServerAspNetCoreConfiguration`, so you can remove these tests from `OpenIddictServerExtensionsTests`.

test/OpenIddict.Server.Tests/AddServerRegistersDefaultHandlerDataAttribute.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit.Sdk;
+
+namespace OpenIddict.Server.Tests
+{
+    public class AddServerRegistersDefaultHandlerDataAttribute : DataAttribute
```
kevinchalet: Since it's not used in multiple class tests, we should probably avoid this extra attribute and use `[MemberData]` instead (e.g https://github.com/openiddict/openiddict-core/blob/1e87e917a688087d8612a6adab4c203b95f9187f/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictResponseTests.cs#L113)

test/OpenIddict.Server.Tests/AddServer_RegistersRequireSingletons_TestCase.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+
+namespace OpenIddict.Server.Tests
+{
+    public class AddServer_RegistersRequireSingletons_TestCase
```
kevinchalet: Same remark.

test/OpenIddict.Server.Tests/AddServer_RegistersRequiredSingletonDataAttribute.cs
```diff
@@ -0,0 +1,53 @@
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit.Sdk;
+
+namespace OpenIddict.Server.Tests
+{
+    public class AddServer_RegistersRequiredSingletonDataAttribute : DataAttribute
```
kevinchalet: Same remark, let's used a property in the main tests class instead.

kevinchalet (commented):
Looks like it's also ready to merge 😄 

If you don't mind, please add #nullable enable at the top of the tests class to ensure it won't cause errors when https://github.com/openiddict/openiddict-core/pull/1116 is merged.

test/OpenIddict.Server.Tests/AddServer_RegistersRequiredSingletonDataAttribute.cs
Darthruneis: Removed.

test/OpenIddict.Server.Tests/AddServer_RegistersRequireSingletons_TestCase.cs
Darthruneis: Removed.

test/OpenIddict.Server.Tests/AddServerRegistersDefaultHandlerDataAttribute.cs
Darthruneis: Removed.

Darthruneis:
The automated builds are likely going to fail, I can't build the solution locally right now. Not sure why, I downloaded the tools noted in `global.json`, but I'm getting several build errors in Visual Studio right now. 

@kevinchalet Is there a place I can refer to to see what changes I need to make locally in order for the solution to build again? I'm getting errors in files and projects that shouldn't have been modified in this PR at all.

kevinchalet:
> @kevinchalet Is there a place I can refer to to see what changes I need to make locally in order for the solution to build again? I'm getting errors in files and projects that shouldn't have been modified in this PR at all.

The errors you're seeing are caused by the [nullable reference feature](https://docs.microsoft.com/en-US/dotnet/csharp/nullable-references), that is now used everywhere in OpenIddict. You can update your tests to make them nullable-aware or, if you're not familiar with this feature, replace `#nullable enable` by `#nullable disable` and I'll take care of updating your tests to be nullable-aware when it's merged.

Darthruneis:
@kevinchalet Right, that is what the PR build is failing for, but locally in Visual Studio, I can't even get that far - I get a bunch of errors throughout the solution, which didn't happen prior to pulling latest (I was a month or two out of date on my fork).

For example:
`OpenIddict.Abstractions.Primitives.OpenIddictConverter.cs, line 53`:         
> Error CS0173: Type of conditional expression cannot be determined because there is no implicit conversion between 'OpenIddict.Abstractions.OpenIddictRequest' and 'OpenIddict.Abstractions.OpenIddictResponse' (53, 20)

The fact that the PR build doesn't hit these errors is what makes me think that I have something missing on my machine, which would impair my ability to work on anything else besides this PR (for which I could use `#nullable disable`, as you mentioned, to get this merged). I am not sure which tool or version update I'm missing - I checked the `global.json` file and grabbed the Net 5.0 preview tools it mentioned.

Some of the other errors I get when building locally do seem to be related to nullable reference types, such as in some of the generic EF models:
`OpenIddict.EntityFramework.Models.OpenIddictEntityFrameworkApplication.cs, line 76`:
> Error CS8627: A nullable type parameter must be known to be a value type or non-nullable reference type. Consider adding a 'class', 'struct', or type constraint. (61, 24)

kevinchalet:
What VS version are you using? The 5.0 beta/RC tooling requires VS preview.

Darthruneis:
Ah, okay. I didn't realize it required VS2019 16.8+. I was using VS2019 16.7.4.

I've got 16.8 installed now and it has gotten past those build issues. 

I'll take a stab at fixing the tests for `#enable nullable`.

Darthruneis:
> If you don't mind, please add #nullable enable at the top of the tests class to ensure it won't cause errors when #1116 is merged.

Since some of the tests are validating behavior that currently still exists around nulls, I wrapped those tests inside of `#disable nullable`, and used `#enable nullable` for the rest.

Whenever the code under test is set up to handle `#enable nullable`, those tests can just be removed.

kevinchalet:
> Since some of the tests are validating behavior that currently still exists around nulls, I wrapped those tests inside of #disable nullable, and used #enable nullable for the rest.

It would be better to null the null-forgiving operator. See https://github.com/openiddict/openiddict-core/pull/1116/files#diff-a6c046f35044d3e84bd4617441b3b548 for an example.

Darthruneis:
> > Since some of the tests are validating behavior that currently still exists around nulls, I wrapped those tests inside of #disable nullable, and used #enable nullable for the rest.
> 
> It would be better to null the null-forgiving operator. See https://github.com/openiddict/openiddict-core/pull/1116/files#diff-a6c046f35044d3e84bd4617441b3b548 for an example.

Oh, didn't know that was a thing with nullable reference types! Applied to the relevant tests.

kevinchalet:
Looks like there's a conflict. Can you please rebase your branch on top of `dev`?

Thanks!

kevinchalet:
PR merged, thanks for your contribution! 👍 👏

PR #1052, opened 2020-08-09
Introduce OpenIddict.Server.Quartz

### LINKED ISSUE #361: Implement automatic expired token flushing
(I was pretty sure we already had a work item to track that, but obviously, nope...)

So, we basically have 2 potential options to implement such a feature:
  - Remove the expired tokens when receiving a new request.
  - Use a dedicated background task/job that would remove the expired tokens from the database at defined intervals. It could potentially be used with something like Hangfire.

I'd love to hear your thoughts about that.

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/361 by introducing a new OpenIddict server/Quartz integration package.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,6 +15,8 @@
     <PackageVersion Include="MongoDB.Driver" Version="2.10.4" />
     <PackageVersion Include="Moq" Version="4.14.5" />
     <PackageVersion Include="Portable.BouncyCastle" Version="1.8.6.7" />
+    <PackageVersion Include="Quartz.Extensions.DependencyInjection" Version="3.2.0-preview-20200806-1639" />
+    <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.2.0-preview-20200806-1639" />
     <PackageVersion Include="System.Collections.Immutable" Version="1.7.1" />
     <PackageVersion Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageVersion Include="System.Linq.Async" Version="4.1.1" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,8 +15,8 @@
     <PackageVersion Include="MongoDB.Driver" Version="2.10.4" />
     <PackageVersion Include="Moq" Version="4.14.5" />
     <PackageVersion Include="Portable.BouncyCastle" Version="1.8.6.7" />
-    <PackageVersion Include="Quartz.Extensions.DependencyInjection" Version="3.2.0-preview-20200806-1639" />
-    <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.2.0-preview-20200806-1639" />
+    <PackageVersion Include="Quartz.Extensions.DependencyInjection" Version="3.2.0-preview-20200810-0611" />
+    <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.2.0-preview-20200810-0611" />
     <PackageVersion Include="System.Collections.Immutable" Version="1.7.1" />
     <PackageVersion Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageVersion Include="System.Linq.Async" Version="4.1.1" />
--- a/samples/Mvc.Server/Startup.cs
+++ b/samples/Mvc.Server/Startup.cs
@@ -5,6 +5,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Mvc.Server.Models;
 using Mvc.Server.Services;
+using Quartz;
 using static OpenIddict.Abstractions.OpenIddictConstants;
 
 namespace Mvc.Server
@@ -45,6 +46,18 @@ namespace Mvc.Server
                 options.ClaimsIdentity.UserIdClaimType = Claims.Subject;
                 options.ClaimsIdentity.RoleClaimType = Claims.Role;
             });
+
+            // OpenIddict offers native integration with Quartz.NET to perform scheduled tasks
+            // (like pruning orphaned authorizations/tokens from the database) at regular intervals.
+            services.AddQuartz(options =>
+            {
+                options.UseMicrosoftDependencyInjectionJobFactory();
+                options.UseSimpleTypeLoader();
+                options.UseInMemoryStore();
+            });
+
+            // Register the Quartz.NET service and configure it to block shutdown until jobs are complete.
+            services.AddQuartzHostedService(options => options.WaitForJobsToComplete = true);
 
             services.AddOpenIddict()
 
@@ -96,6 +109,9 @@ namespace Mvc.Server
                            .EnableUserinfoEndpointPassthrough()
                            .EnableVerificationEndpointPassthrough()
                            .DisableTransportSecurityRequirement(); // During development, you can disable the HTTPS requirement.
+
+                    // Enable Quartz.NET integration.
+                    options.UseQuartz();
 
                     // Note: if you don't want to specify a client_id when sending
                     // a token or revocation request, uncomment the following line:
--- a/src/OpenIddict.Server.Quartz/OpenIddictServerQuartzExtensions.cs
+++ b/src/OpenIddict.Server.Quartz/OpenIddictServerQuartzExtensions.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Extensions.DependencyInjection
                     JobBuilder.Create<OpenIddictServerQuartzJob>()
                         .StoreDurably()
                         .WithIdentity(OpenIddictServerQuartzJob.Identity)
-                        .WithDescription(SR.GetResourceString(SR.ID8000))
+                        .WithDescription(SR.GetResourceString(SR.ID9000))
                         .Build());
             }
 
@@ -65,7 +65,7 @@ namespace Microsoft.Extensions.DependencyInjection
                     TriggerBuilder.Create()
                         .ForJob(OpenIddictServerQuartzJob.Identity)
                         .WithSimpleSchedule(options => options.WithIntervalInHours(1).RepeatForever())
-                        .WithDescription(SR.GetResourceString(SR.ID8001))
+                        .WithDescription(SR.GetResourceString(SR.ID9001))
                         .StartNow()
                         .Build());
             }
--- a/src/OpenIddict.Server.Quartz/OpenIddictServerQuartzExtensions.cs
+++ b/src/OpenIddict.Server.Quartz/OpenIddictServerQuartzExtensions.cs
@@ -1,0 +1,101 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/openiddict/openiddict-core for more information concerning
+ * the license and the contributors participating to this project.
+ */
+
+using System;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using OpenIddict.Server.Quartz;
+using Quartz;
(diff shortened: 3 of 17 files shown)


Directory.Packages.props
```diff
@@ -15,6 +15,8 @@
     <PackageVersion Include="MongoDB.Driver" Version="2.10.4" />
     <PackageVersion Include="Moq" Version="4.14.5" />
     <PackageVersion Include="Portable.BouncyCastle" Version="1.8.6.7" />
+    <PackageVersion Include="Quartz.Extensions.DependencyInjection" Version="3.2.0-preview-20200806-1639" />
+    <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.2.0-preview-20200806-1639" />
```
lahma: I'm trying to get [Quartz.Extensions.Hosting](https://www.nuget.org/packages/Quartz.Extensions.Hosting/) to Quartz project, until then need to use the preview feed. Fortunately Quartz is not protected on NuGet and people cannot use the prefix without being in Quartz.NET organization, less things to worry...

samples/Mvc.Server/Startup.cs
```diff
@@ -46,6 +47,18 @@ public void ConfigureServices(IServiceCollection services)
                 options.ClaimsIdentity.RoleClaimType = Claims.Role;
             });
 
+            // OpenIddict offers native integration with Quartz.NET to perform scheduled tasks
+            // (like pruning orphaned authorizations/tokens from the database) at regular intervals.
+            services.AddQuartz(options =>
```
lahma, line 52: This `AddQuartz` registers also the Microsoft logging "bridge", a.k.a. `MicrosoftLoggingProvider`  which will enable Quartz to log to global logger factory. This will affect all schedulers build, so it's global and not scheduler factory global. If you would take the route to have scheduler per tenant then we would need `options.UseLoggerFactory(yourScopedFactory);`?

samples/Mvc.Server/Startup.cs
```diff
@@ -46,6 +47,18 @@ public void ConfigureServices(IServiceCollection services)
...
+            {
+                options.UseMicrosoftDependencyInjectionJobFactory();
+                options.UseSimpleTypeLoader();
+                options.UseInMemoryStore();
+            });
+
+            // Register the Quartz.NET service and configure it to block shutdown until jobs are complete.
+            services.AddQuartzHostedService(options => options.WaitForJobsToComplete = true);
```
lahma, line 60: For use case where there would be multiple scheduler, we would need an overload or something to resolve specific scheduler, just commenting on Quartz logic, not your implementation 🙂

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            builder.Services.TryAddTransient<OpenIddictServerQuartzJob>();
+
+            // To ensure this method can be safely called multiple times, the job details
+            // of the OpenIddict server job are only added if no existing IJobDetail instance
+            // pointing to OpenIddictServerQuartzJob was already registered in the DI container.
+            if (!builder.Services.Any(descriptor => descriptor.ServiceType == typeof(IJobDetail) &&
+                                                    descriptor.ImplementationInstance is IJobDetail job &&
+                                                    job.Key == OpenIddictServerQuartzJob.Identity))
+            {
+                builder.Services.AddSingleton(
```
lahma, line 46: Jobs should be transient, new instance for each invocation.

lahma: Scratch that, this was the detail.. sorry

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
```diff
@@ -0,0 +1,154 @@
...
+    public class OpenIddictServerQuartzJob : IJob
+    {
+        private readonly ILogger _logger;
+        private readonly IOptionsMonitor<OpenIddictServerQuartzOptions> _options;
+        private readonly IServiceProvider _provider;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="OpenIddictServerQuartzJob"/> class.
+        /// </summary>
+        public OpenIddictServerQuartzJob() => throw new InvalidOperationException(SR.GetResourceString(SR.ID1081));
```
lahma, line 30: Quartz options for job factory has `AllowDefaultConstructor` which defaults to false, so Quartz shouldn't call this by default. I think could also be private?

lahma: Ah you now when I track this back it seems that you want to catch invalid configuration, neat.

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
```diff
@@ -0,0 +1,154 @@
...
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // TODO: determine whether errors should be swallowed or should be re-thrown to be logged by Quartz.NET.
+
+            // Note: this job is registered as a transient service. As such, it cannot directly depend on scoped services
+            // like the core managers. To work around this limitation, a scope is manually created for each invocation.
+            var scope = _provider.CreateScope();
```
lahma, line 62: There's also `MicrosoftDependencyInjectionScopedJobFactory` which should provide scoping.

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                // Note: the authorization/token managers MUST be resolved from the scoped provider
+                // as they depend on scoped stores that should be disposed as soon as possible.
+
+                if (!_options.CurrentValue.DisableAuthorizationsPruning)
+                {
+                    var manager = scope.ServiceProvider.GetService<IOpenIddictAuthorizationManager>();
+                    if (manager == null)
+                    {
+                        throw new InvalidOperationException(SR.GetResourceString(SR.ID1277));
```
lahma: Ideally you should only throw `JobExecutionException` from Execute, any other exception signals scheduler error which "a little bit more severe". `JobExecutionException` also allows you to instruct scheduler like "remove triggers, this won't ever work" or "re-fire immedialtely, we can do this!"

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
```diff
@@ -0,0 +1,154 @@
...
+                        _logger.LogError(exception, SR.GetResourceString(SR.ID7118));
+                    }
+                }
+
+                if (!_options.CurrentValue.DisableTokensPruning)
+                {
+                    var manager = scope.ServiceProvider.GetService<IOpenIddictTokenManager>();
+                    if (manager == null)
+                    {
+                        throw new InvalidOperationException(SR.GetResourceString(SR.ID1277));
```
lahma: Again prefer `JobExecutionException`

Directory.Packages.props
kevinchalet: Good to know, thanks! We've had similar issues with a few aspnet-contrib packages, but luckily, all the libs authors we've spoken to all have been kind enough to transfer the ownership to the aspnet-contrib org 😃

samples/Mvc.Server/Startup.cs
kevinchalet: If you wanted to support having multiple scheduler factories in the DI container but wanted a different logger factory for each, yeah, that would be a nice option. Now, if you implement multi-tenancy a-la-Orchard-Core, it manages an `IServiceProvider` per tenant, so each tenant would get a unique scheduler and a unique logger factory.

samples/Mvc.Server/Startup.cs
kevinchalet: Yep. Since MS's DI doesn't support named services, it's something you'd have to implement yourself, likely by using derived scheduler types. Named options w/ `Microsoft.Extensions.Options` would be a very nice alternative if you wanted to support multiple schedulers and different options per scheduler.

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
kevinchalet: Absolutely, the parameterless constructor is just here to inform the user that enabling DI is mandatory for the OpenIddict server job to work correctly. Once DI is enabled, the other constructor will always be preferred (whether you're using the MS DI or any other compliant DI like Autofac).

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
kevinchalet: `MicrosoftDependencyInjectionScopedJobFactory` is a nice option, but it's hard for a lib to assume it's the option chosen by the developer (you'll generally get a cryptic error message from DI when trying to resolve a scoped service from a transient one).

We could probably find a way to force users to use `MicrosoftDependencyInjectionScopedJobFactory` instead of `MicrosoftDependencyInjectionJobFactory`, but it would affect every other job (not just ours), so I'm not sure it would be a nice thing to do in a library 😕

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
kevinchalet: Wooot, that's indeed much better, thanks for the info!

> "re-fire immedialtely, we can do this!"

It would be a nice option for transient exceptions. Is there a limit to prevent a job from returning `JobExecutionException`s with `RefireImmediately` set to `true` indefinitely or is it something that should be done inside the job itself?

Directory.Packages.props
lahma: Yes the two other integration package names were also taken, but fortunately there was no problem for the authors to contribute the package id.

src/OpenIddict.Server.Quartz/OpenIddictServerQuartzJob.cs
lahma: If you signal the `RefireImmediately = true` you [start looping inside the job invocation logic](https://github.com/quartznet/quartznet/blob/9dfe5ee5ee93a53cea87b9950234c0012afac9f1/src/Quartz/Core/JobRunShell.cs#L255-L271). This is basically the same thing as if you would do the loop with try-catch inside your job trying to get successful execution. So if you consider exception transient, you should signal the flag. This however means that you job execution time could vary from normal 2 seconds to 5 minutes. If you job would run every 1 minute there would be misfires (concurrent execution disallowed) and you would need to have misfire policy in place, like Ignore, continue with next scheduled time when it comes.

You will also get the misfire count from job execution context in your job if you want to abort trying after some trials.

kevinchalet: > This however means that you job execution time could vary from normal 2 seconds to 5 minutes. If you job would run every 1 minute there would be misfires (concurrent execution disallowed) and you would need to have misfire policy in place, like Ignore, continue with next scheduled time when it comes.

Thanks for those precious details!

> You will also get the misfire count from job execution context in your job if you want to abort trying after some trials.

Perfect. I'll update the job to catch those transient exceptions and re-throw `JobExecutionException`s (and check the `RefireCount` property to ensure it doesn't cause endless loops 😃)

kevinchalet:
PR merged, thanks a lot for your help reviewing it @lahma! 🎉 

I opened https://github.com/openiddict/openiddict-core/issues/1053 to track marking `OpenIddict.Server.Quartz` as a shipping package.

lahma:
Awesome! I'll try to address the options thing on Quartz's side hopefully in the near future.

PR #1090, opened 2020-09-17
Update OpenIddictResources.resx

kevinchalet:
@dfederm would that help make the exception message clearer about what you need to do?


dfederm:
@kevinchalet I think so, yes. Thanks!

dfederm (approved):

PR #1116, opened 2020-10-06
Enable nullable reference types support in the unit/integration tests…

### LINKED ISSUE #1037: Enable nullable reference types support in the unit/integration tests projects
As part of https://github.com/openiddict/openiddict-core/pull/1035, all the OpenIddict projects have been updated to use nullable reference types.
We'll now need to enable them in the unit/integration tests projects and update them accordingly.

johanndev:
Closes #1037 

As I was not sure about the most sensible approach at several places, I just "blindly" fixed some syntax errors for now. I will add comments to the affected lines to discuss the best approach.

### DIFF
--- a/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictParameterTests.cs
+++ b/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictParameterTests.cs
@@ -323,7 +323,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
                 JsonSerializer.Deserialize<JsonElement>(@"{""parameter"":""value""}"));
 
             // Act and assert
-            Assert.Equal("value", (string) parameter.GetNamedParameter("parameter"));
+            Assert.Equal("value", (string?) parameter.GetNamedParameter("parameter"));
         }
 
         [Fact]
@@ -375,7 +375,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
             });
 
             // Act and assert
-            Assert.Equal("Fabrikam", (string) parameter.GetUnnamedParameter(0));
+            Assert.Equal("Fabrikam", (string?) parameter.GetUnnamedParameter(0));
         }
 
         [Fact]
@@ -408,7 +408,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
                 JsonSerializer.Deserialize<JsonElement>(@"[""Fabrikam"",""Contoso""]"));
 
             // Act and assert
-            Assert.Equal("Fabrikam", (string) parameter.GetUnnamedParameter(0));
+            Assert.Equal("Fabrikam", (string?) parameter.GetUnnamedParameter(0));
         }
 
         [Fact]
@@ -472,7 +472,8 @@ namespace OpenIddict.Abstractions.Tests.Primitives
                 JsonSerializer.Deserialize<JsonElement>(@"{""parameter"":""value""}"));
 
             // Act and assert
-            Assert.Equal(parameters, parameter.GetNamedParameters().ToDictionary(pair => pair.Key, pair => (string) pair.Value));
+            // TODO: NRT
+            Assert.Equal(parameters, parameter.GetNamedParameters().ToDictionary(pair => pair.Key, pair => (string) pair.Value!));
         }
 
         [Fact]
@@ -510,7 +511,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
 
             // Act and assert
             Assert.Equal(parameters, from element in parameter.GetUnnamedParameters()
-                                     select (string) element);
+                                     select (string?) element);
         }
 
         [Fact]
@@ -539,7 +540,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
 
             // Act and assert
             Assert.Equal(parameters, from element in parameter.GetUnnamedParameters()
-                                     select (string) element);
+                                     select (string?) element);
         }
 
         [Fact]
@@ -559,8 +560,8 @@ namespace OpenIddict.Abstractions.Tests.Primitives
             // Arrange, act and assert
             Assert.True(OpenIddictParameter.IsNullOrEmpty(new OpenIddictParameter((bool?) null)));
             Assert.True(OpenIddictParameter.IsNullOrEmpty(new OpenIddictParameter((long?) null)));
-            Assert.True(OpenIddictParameter.IsNullOrEmpty(new OpenIddictParameter((string) null)));
-            Assert.True(OpenIddictParameter.IsNullOrEmpty(new OpenIddictParameter((string[]) null)));
+            Assert.True(OpenIddictParameter.IsNullOrEmpty(new OpenIddictParameter((string?) null)));
+            Assert.True(OpenIddictParameter.IsNullOrEmpty(new OpenIddictParameter((string[]?) null)));
         }
 
         [Fact]
@@ -760,7 +761,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
 
             // Act and assert
             Assert.True(parameter.TryGetNamedParameter("parameter", out var value));
-            Assert.Equal("value", (string) value);
+            Assert.Equal("value", (string?) value);
         }
 
         [Fact]
@@ -815,7 +816,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
 
             // Act and assert
             Assert.True(parameter.TryGetUnnamedParameter(0, out var value));
-            Assert.Equal("Fabrikam", (string) value);
+            Assert.Equal("Fabrikam", (string?) value);
         }
 
         [Fact]
@@ -851,7 +852,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
 
             // Act and assert
             Assert.True(parameter.TryGetUnnamedParameter(0, out var value));
-            Assert.Equal("Fabrikam", (string) value);
+            Assert.Equal("Fabrikam", (string?) value);
         }
 
         [Fact]
@@ -861,7 +862,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
             var parameter = new OpenIddictParameter();
 
             // Act and assert
-            var exception = Assert.Throws<ArgumentNullException>(() => parameter.WriteTo(writer: null));
+            var exception = Assert.Throws<ArgumentNullException>(() => parameter.WriteTo(writer: null!));
             Assert.Equal("writer", exception.ParamName);
         }
 
@@ -890,11 +891,11 @@ namespace OpenIddict.Abstractions.Tests.Primitives
         public void BoolConverter_CanCreateParameterFromBooleanValue()
         {
             // Arrange, act and assert
-            Assert.True((bool) new OpenIddictParameter(true).Value);
-            Assert.True((bool) new OpenIddictParameter((bool?) true).Value);
+            Assert.True((bool?) new OpenIddictParameter(true).Value);
+            Assert.True((bool?) new OpenIddictParameter((bool?) true).Value);
 
-            Assert.False((bool) new OpenIddictParameter(false).Value);
-            Assert.False((bool) new OpenIddictParameter((bool?) false).Value);
+            Assert.False((bool?) new OpenIddictParameter(false).Value);
+            Assert.False((bool?) new OpenIddictParameter((bool?) false).Value);
         }
 
         [Fact]
@@ -1037,15 +1038,15 @@ namespace OpenIddict.Abstractions.Tests.Primitives
         public void LongConverter_CanCreateParameterFromLongValue()
         {
             // Arrange, act and assert
-            Assert.Equal(42, (long) new OpenIddictParameter(42).Value);
-            Assert.Equal(42, (long) new OpenIddictParameter((long?) 42).Value);
+            Assert.Equal(42, (long?) new OpenIddictParameter(42).Value);
+            Assert.Equal(42, (long?) new OpenIddictParameter((long?) 42).Value);
         }
 
         [Fact]
         public void LongConverter_ReturnsDefaultValueForNullValues()
         {
             // Arrange, act and assert
-            Assert.Equal(0, (long) new OpenIddictParameter());
+            Assert.Equal(0, (long)new OpenIddictParameter());
             Assert.Null((long?) new OpenIddictParameter());
         }
 
@@ -1053,7 +1054,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
         public void LongConverter_ReturnsDefaultValueForUnsupportedPrimitiveValues()
         {
             // Arrange, act and assert
-            Assert.Equal(0, (long) new OpenIddictParameter("Fabrikam"));
+            Assert.Equal(0, (long)new OpenIddictParameter("Fabrikam"));
             Assert.Null((long?) new OpenIddictParameter("Fabrikam"));
         }
 
@@ -1061,7 +1062,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
         public void LongConverter_ReturnsDefaultValueForUnsupportedArrays()
         {
             // Arrange, act and assert
-            Assert.Equal(0, (long) new OpenIddictParameter(new[] { "Contoso", "Fabrikam" }));
+            Assert.Equal(0, (long)new OpenIddictParameter(new[] { "Contoso", "Fabrikam" }));
             Assert.Null((long?) new OpenIddictParameter(new[] { "Contoso", "Fabrikam" }));
         }
 
@@ -1069,7 +1070,7 @@ namespace OpenIddict.Abstractions.Tests.Primitives
         public void LongConverter_ReturnsDefaultValueForUnsupportedJsonValues()
         {
             // Arrange, act and assert
-            Assert.Equal(0, (long) new OpenIddictParameter(JsonSerializer.Deserialize<JsonElement>("[]")));
+            Assert.Equal(0, (long)new OpenIddictParameter(JsonSerializer.Deserialize<JsonElement>("[]")));
             Assert.Null((long?) new OpenIddictParameter(JsonSerializer.Deserialize<JsonElement>("[]")));
         }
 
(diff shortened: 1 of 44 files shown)


test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictParameterTests.cs
```diff
@@ -472,7 +472,8 @@ public void GetNamedParameters_ReturnsExpectedParametersForJsonObjects()
                 JsonSerializer.Deserialize<JsonElement>(@"{""parameter"":""value""}"));
 
             // Act and assert
-            Assert.Equal(parameters, parameter.GetNamedParameters().ToDictionary(pair => pair.Key, pair => (string) pair.Value));
+            // TODO: NRT
+            Assert.Equal(parameters, parameter.GetNamedParameters().ToDictionary(pair => pair.Key, pair => (string) pair.Value!));
```
johanndev, line 476: Is the null-forgiving operator appropriate here?

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTestClient.cs
```diff
@@ -410,7 +410,8 @@ static string UnescapeDataString(string value)
 
                     var value = UnescapeDataString(segment.Substring(index + 1, segment.Length - (index + 1)));
 
-                    parameters.Add(new KeyValuePair<string, string>(name, value));
+                    // TODO NRT
+                    parameters.Add(new KeyValuePair<string, string>(name, value!));
```
johanndev: Is the null-forgiving operator appropriate here?

test/OpenIddict.Server.Owin.IntegrationTests/OpenIddictServerOwinIntegrationTests.Exchange.cs
```diff
@@ -30,7 +30,8 @@ public async Task ExtractTokenRequest_MultipleClientCredentialsCauseAnError()
                     builder.UseInlineHandler(context =>
                     {
                         var request = context.Transaction.GetOwinRequest();
-                        request.Headers["Authorization"] = "Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW";
+                        // TODO: NRT
+                        request?.Headers!["Authorization"] = "Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW";
```
johanndev: Without the `!`, `Headers` gets flagged as possible null here.

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictParameterTests.cs
johanndev: Is the null-forgiving operator appropriate here?

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs
```diff
@@ -57,8 +57,10 @@ ValueTask DisposeAsync()
 
 #if SUPPORTS_GENERIC_HOST
             // Stop and dispose of the underlying generic host.
-            await Host.StopAsync();
-            Host.Dispose();
+            if (Host is object) {
```
johanndev: Very unsure about this one 🤔

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.cs
```diff
@@ -604,7 +604,7 @@ void ConfigurePipeline(IApplicationBuilder app)
                         await context.Response.WriteAsync(JsonSerializer.Serialize(
                             new OpenIddictResponse(result.Principal.Claims.GroupBy(claim => claim.Type)
                                 .Select(group => new KeyValuePair<string, string[]>(
-                                    group.Key, group.Select(claim => claim.Value).ToArray())))));
+                                    group.Key, group.Select(claim => claim.Value).ToArray()))!)));
```
johanndev: 🤷

johanndev (commented):
I would be grateful about some comments on the changes :)

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictParameterTests.cs
kevinchalet: Would `(string?)` work here (instead of using `!` on `Value`)?

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictRequestTests.cs
```diff
@@ -315,7 +315,7 @@ public void PropertyGetter_ReturnsExpectedParameter(string property, string name
             request.SetParameter(name, value);
 
             // Act and assert
-            Assert.Equal(value.Value, typeof(OpenIddictRequest).GetProperty(property).GetValue(request));
+            Assert.Equal(value.Value, typeof(OpenIddictRequest)?.GetProperty(property)?.GetValue(request));
```
kevinchalet: Would it make more sense to use `!` here (for both ones)?

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictRequestTests.cs
```diff
@@ -326,7 +326,7 @@ public void PropertySetter_AddsExpectedParameter(string property, string name, O
             var request = new OpenIddictRequest();
 
             // Act
-            typeof(OpenIddictRequest).GetProperty(property).SetValue(request, value.Value);
+            typeof(OpenIddictRequest)?.GetProperty(property)?.SetValue(request, value.Value);
```
kevinchalet: Same remark.

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictResponseTests.cs
```diff
@@ -118,7 +118,7 @@ public void PropertyGetter_ReturnsExpectedParameter(string property, string name
             response.SetParameter(name, value);
 
             // Act and assert
-            Assert.Equal(value.Value, typeof(OpenIddictResponse).GetProperty(property).GetValue(response));
+            Assert.Equal(value.Value, typeof(OpenIddictResponse)?.GetProperty(property)?.GetValue(response));
```
kevinchalet: Here too.

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictResponseTests.cs
```diff
@@ -129,7 +129,7 @@ public void PropertySetter_AddsExpectedParameter(string property, string name, O
             var response = new OpenIddictResponse();
 
             // Act
-            typeof(OpenIddictResponse).GetProperty(property).SetValue(response, value.Value);
+            typeof(OpenIddictResponse).GetProperty(property)?.SetValue(response, value.Value);
```
kevinchalet: And here? 😄

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs
```diff
@@ -37,7 +37,7 @@ public OpenIddictServerAspNetCoreIntegrationTestServer(IHost host)
         /// <summary>
         /// Gets the generic host used by this instance.
         /// </summary>
-        public IHost Host { get; }
+        public IHost? Host { get; }
```
kevinchalet: I'd keep the host property non-nullable, as it's always set by the constructor.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs
kevinchalet: I think you should be able to revert this change if you make `Host` non-nullable.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.Exchange.cs
```diff
@@ -31,7 +31,7 @@ public async Task ExtractTokenRequest_MultipleClientCredentialsCauseAnError()
                     builder.UseInlineHandler(context =>
                     {
                         var request = context.Transaction.GetHttpRequest();
-                        request.Headers[HeaderNames.Authorization] = "Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW";
+                        request?.Headers![HeaderNames.Authorization] = "Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW";
```
kevinchalet: Since `request` is not supposed to be null here, I'd simply use `!` on `request`. Regarding the other one, not sure why it's required. Do you have an error if you do `request!.Headers`?

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.cs
```diff
@@ -328,7 +328,7 @@ public async Task ProcessRequest_AllowsHandlingResponse(string address)
             var response = await client.PostAsync(address, new OpenIddictRequest());
 
             // Assert
-            Assert.Equal("Bob le Bricoleur", (string) response["name"]);
+            Assert.Equal("Bob le Bricoleur", (string?)response["name"]);
```
kevinchalet: nit: please keep the space after cast operations 😄

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.cs
```diff
@@ -363,7 +363,7 @@ public async Task ProcessRequest_AllowsSkippingHandler(string address)
             var response = await client.PostAsync(address, new OpenIddictRequest());
 
             // Assert
-            Assert.Equal("Bob le Magnifique", (string) response["name"]);
+            Assert.Equal("Bob le Magnifique", (string?)response["name"]);
```
kevinchalet: Here too (I see other occurrences too).

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.cs
kevinchalet: Well, if it's required... 😄

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTestClient.cs
```diff
@@ -268,7 +268,7 @@ private HttpRequestMessage CreateRequestMessage(OpenIddictRequest request, HttpM
                     continue;
                 }
 
-                var values = (string[]) parameter.Value;
+                var values = (string[]) parameter.Value!;
```
kevinchalet: Should we declare `parameters` as `List<KeyValuePair<string, string?>>()` and use `string?[]?` instead here?

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTestClient.cs
kevinchalet: If we update `parameters` to make the "value" part of `KeyValuePair<>` nullable, I guess it will no longer be required.

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTests.cs
```diff
@@ -1523,7 +1523,7 @@ public async Task ProcessSignIn_ScopeDefaultsToOpenId()
                 options.AddEventHandler<ProcessSignInContext>(builder =>
                     builder.UseInlineHandler(context =>
                     {
-                        Assert.Equal(new[] { Scopes.OpenId }, context.Principal.GetScopes());
+                        Assert.Equal(new[] { Scopes.OpenId }, context.Principal?.GetScopes());
```
kevinchalet: I'd just use `!` here (and in the other tests too).

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTests.cs
```diff
@@ -4223,10 +4223,10 @@ protected virtual void ConfigureServices(IServiceCollection services)
                 });
         }
 
-        protected abstract ValueTask<OpenIddictServerIntegrationTestServer> CreateServerAsync(Action<OpenIddictServerBuilder> configuration = null);
+        protected abstract ValueTask<OpenIddictServerIntegrationTestServer> CreateServerAsync(Action<OpenIddictServerBuilder> configuration = null!);
```
kevinchalet: It would be better to make the `configuration` parameter nullable instead of using the null-forgiving operator.

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTests.cs
```diff
@@ -4223,10 +4223,10 @@ protected virtual void ConfigureServices(IServiceCollection services)
...
 
         protected OpenIddictApplicationManager<OpenIddictApplication> CreateApplicationManager(
-            Action<Mock<OpenIddictApplicationManager<OpenIddictApplication>>> configuration = null)
+            Action<Mock<OpenIddictApplicationManager<OpenIddictApplication>>> configuration = null!)
```
kevinchalet: Same remark.

test/OpenIddict.Server.Owin.IntegrationTests/OpenIddictServerOwinIntegrationTests.Exchange.cs
kevinchalet: Instead of the `?`, I'd add a `!` after `context.Transaction.GetOwinRequest()`.
If the `!` is required for `Headers` after this change, let's keep it.

kevinchalet (approved):
Thanks for your PR! Excellent work!

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTestClient.cs
kevinchalet: Regarding the null-forgiving operator used for `Value`, I opened https://github.com/openiddict/openiddict-core/issues/1118 to track future improvements that could help avoid it.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs
johanndev: The problem is that the first ctor (https://github.com/openiddict/openiddict-core/blob/ba85a4b618f5e6065ca6310881af6b6c0fff04e3/test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs#L20) does not initialize the `Host` property; therefore it must be marked as nullable.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.cs
johanndev: I took a second look and just and there is a much more reasonable solution: Align the types :)

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictRequestTests.cs
johanndev: The first `?` is actually unnecessary, the second on can reasonably be replaced by a `!`.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs
johanndev: As remarked above, the first ctor does not initialize the `Host` property.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.Exchange.cs
johanndev: I propose just adding a `!` to the `GetHttpRequest()` method and lose all other nullability-related operators on the next line.

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTests.cs
johanndev: Whoops, will do :)

test/OpenIddict.Server.AspNetCore.IntegrationTests/OpenIddictServerAspNetCoreIntegrationTestServer.cs
kevinchalet: Oh, then we should rework the definition a bit to only add this constructor when generic host is not supported. Something like that:

```csharp
#if SUPPORTS_GENERIC_HOST
        public OpenIddictServerAspNetCoreIntegrationTestServer(IHost host)
        {
            Host = host;
            Server = host.GetTestServer();
        }
#else
        public OpenIddictServerAspNetCoreIntegrationTestServer(TestServer server)
            => Server = server;
#endif
```

johanndev:
@kevinchalet Do you prefer that I squash all changes into a single commit?

kevinchalet:
You can, but it's not necessary (GitHub has an option to do so when "merging" the PR) 😃 

I'll take a look at your changes later today. Thanks again.

kevinchalet:
Merged, thanks a lot for your great contribution! 👏

PR #1149, opened 2020-10-26
Stop adding a jti claim by default

kevinchalet:


horseznguyen:
Hi @kevinchalet ,

Why you stop adding a jti claim, currently i need it to handle blacklist, so How to add or configure it.

PR #1260, opened 2021-05-07
Updated FormValueRequiredAttribute to include "application/x-www-form…

danlister:
…-urlencoded"

I've experienced different content types being used when sending POST requests. For example: `application/x-www-form-urlencoded`. Hopefully this change helps catch the additional variations :)


kevinchalet:
Thanks for your PR.

Do you have an actual example of a service that sends this value? AFAIK, this is not a standard variation and OpenIddict itself doesn't support it:

https://github.com/openiddict/openiddict-core/blob/422d8979adb8cdebc6c8c8e14faa1d736208271e/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.cs#L493-L504

Issue #1242, opened 2021-04-07
Combine MongoDB store with InMemory Store.


SebastianStehle:
### Is your feature request related to a problem?
I would like to use an in-memory store because I only have one application and one scope and I do not see that much value of persisting it to MongoDB. It also makes migration harder, when the URL changes.

It is very straight forward to implement such a store, but it is hard to combine it with the MongoDB implementation. The reason is that the TokenStore uses BsonId for references to the application and scope.

### Describe the solution you'd like
Just change the properties to String.

kevinchalet:
Using mixed stores is not really a supported scenario (and for in-memory scenarios, my recommendation is to use EF Core + the in-memory integration, or Sqlite with an in-memory database).

Changing the Id properties would be a breaking change so it's not an option at this point. That said, I don't think there's anything preventing you from using `ObjectId` in your custom implementation, is there?

SebastianStehle:
It could be implemented as non-breaking change with a custom serializer or so, but I see that this is an issue.

I am about to migrate 2 services from identity server to openiddict-core. 

In the first case it is very simple because I only have one client and one scope and I created in-memory implementations where I create ObjectId-compliant Ids.

But in another use case the applications are dependent on other services.

This is the current implementation wit: https://github.com/Squidex/squidex/blob/master/backend/src/Squidex/Areas/IdentityServer/Config/LazyClientStore.cs

So my hope was that I am able to port this store to openid-dict and when I want to support multiple databases like Mongo or EF it becomes very likely impossible.

Of course I can also write code where I hook into all the places where I create clients and apps and users in my system and then create applications in openid-dict but it is more complicated and because of the redirect URIs I also have to run a migration when a backup is restored into another machine.

kevinchalet:
> It could be implemented as non-breaking change with a custom serializer or so, but I see that this is an issue.

I don't think a custom serializer would be required as `[BsonId, BsonRepresentation(BsonType.ObjectId)]` could likely be used. But custom serializer or not, changing the type of a CLR property is a binary and source breaking change and not something we can easily do in a non-major version.

> So my hope was that I am able to port this store to openid-dict and when I want to support multiple databases like Mongo or EF it becomes very likely impossible.

[...]

PR #1265, opened 2021-05-15
Custom keys for MongoDB.

SebastianStehle:
Hi,

I created a feature request a while ago: https://github.com/openiddict/openiddict-core/issues/1242

I tried to solve it by making the key generic so that I can replace it with a string instead.

### DIFF
--- a/samples/Mvc.Server/Startup.cs
+++ b/samples/Mvc.Server/Startup.cs
@@ -3,9 +3,11 @@ using Microsoft.AspNetCore.Identity;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using MongoDB.Driver;
 using Mvc.Server.Models;
 using Mvc.Server.Services;
 using Quartz;
+using System;
 using static OpenIddict.Abstractions.OpenIddictConstants;
 
 namespace Mvc.Server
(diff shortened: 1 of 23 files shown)


SebastianStehle:
I was testing the sample, but the client cannot connect due to SSL issues. How do you handle this?

SebastianStehle:
Hi,

I am using this in production since 2 months now and it is working fine. Any chance we can get it through? Do you want me to do any improvements or refactorings?

kevinchalet:
Hi @SebastianStehle,

As part of the 4.0 effort, I'm actually evaluating a few other options, including the creation of "delegating" stores that would be able to wrap one or multiple stores and unify their results. I suspect this may suit your original need (without requiring to hack the MongoDB stores to share Ids with EntityFramework 😄)

I'll keep you informed.

kevinchalet:
Sorry for not getting back to you earlier.

> As part of the 4.0 effort, I'm actually evaluating a few other options, including the creation of "delegating" stores that would be able to wrap one or multiple stores and unify their results. I suspect this may suit your original need (without requiring to hack the MongoDB stores to share Ids with EntityFramework 😄)

I discussed this option with a few OpenIddict users that could have been potentially interested in using such an option and we came to the conclusion that it could never be as flexible as custom stores. As such, it's unlikely that investing resources/time to have that directly in OpenIddict would make a lot of sense.

Having custom stores that exactly match your needs - as you did - remains the best option.

Closing.

SebastianStehle:
Hi, can we talk about this again? I am still maintaining my own fork here and would like to prevent that.

kevinchalet:
> Hi, can we talk about this again?

Certainly, but AFAICT, you're still the only one asking for such a feature. If I could dedicate an infinite amount of time to my OSS projects, I'd of course implement all the cool things folks ask for, but the truth is my resources are limited and given the limited number of people that would use this feature, it's very hard to justify spending time implementing and maintaining it.

That said, I'll definitely keep an eye on this thread to see if there are more people interested 👍🏻

SebastianStehle:
I would like to care about the implementation. Therefore the PR, but I understand the point about maintaining...

PR #1266, opened 2021-05-16
Disable caching per entity.

SebastianStehle:
Because I replace only some stores I would like to disable caching for them. I have solved it with a custom ApplicationCache but I think this is more efficient.

### DIFF
--- a/src/OpenIddict.Core/Managers/OpenIddictApplicationManager.cs
+++ b/src/OpenIddict.Core/Managers/OpenIddictApplicationManager.cs
@@ -178,7 +178,7 @@ namespace OpenIddict.Core
 
             await Store.CreateAsync(application, cancellationToken);
 
-            if (!Options.CurrentValue.DisableEntityCaching)
+            if (!Options.CurrentValue.DisableApplicationCaching)
             {
                 await Cache.AddAsync(application, cancellationToken);
             }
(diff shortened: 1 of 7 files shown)


kevinchalet:
FWIW, I was tempted to remove the ability to disable entity caching in 3.0 as scenarios where you'd want to do so are honestly quite rare (when I added it, it served only purpose: allowing custom stores already implementing a L2-like caching mechanism to avoid double-caching).

Can you tell me more about your scenario?

SebastianStehle:
In my application I use Orleans as backbone for all my domain objects. So the applications/clients are stored as distributed actors in my cluster with a little bit of local caching for peaks (< 2 minutes cache time or so).  My custom application store is therefore read-only. Because of Orleans I have no distributed cache and if an business object gets deleted I would have to remove the entry from all local caches in my cluster. Getting an domain object through Orleans is usually < 1ms so I don't need a cache. I am porting my identity-server code to openiddict and I know that I am fine with the current performance.

kevinchalet:
Interesting scenario! Is the source code public? 😃 

> I am porting my identity-server code to openiddict and I know that I am fine with the current performance.

OpenIddict works differently than IdentityServer: in 3.0, pretty much every aspect of the request processing is represented as a specialized event handler that may or may not need to access the stores (e.g a handler validates the grant type permissions and another one the response type permissions). All these handlers eventually form a coherent pipeline but are actually pretty much independent from each other. As such, methods like `FindByClientIdAsync` may be called multiple times per-request, which is why there's a per-scope/request L2-like cache in OpenIddict to ensure only the first call reaches the store implementation. Even if your store implementation is super fast, I would still recommend leaving the cache enabled.

SebastianStehle:
Oh, stupid me. I have not realized that it is a scoped cache. Then this PR is indeed useless.

Then it can be closed. The code is at https://github.com/squidex/squidex

kevinchalet:
Thanks, I'll take a look 👍🏻 

BTW, thanks for the sponsorship, much appreciated.

PR #1267, opened 2021-05-16
Extract client secret hasher.

SebastianStehle:
In my case the client secrets are already hashed, so I don't need this code. I can just create a custom application manager (and it works) but it is easier like this, I think.

Also makes unit testing easier.

### DIFF
--- a/src/OpenIddict.Abstractions/Managers/NullClientSecretHasher.cs
+++ b/src/OpenIddict.Abstractions/Managers/NullClientSecretHasher.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace OpenIddict.Abstractions
+{
+    /// <summary>
+    /// Implements the <see cref="IOpenIddictClientSecretHasher"/> with a null implementation tht does not hash the secret.
+    /// </summary>
+    /// <remarks>
+    /// Only use this class for custom implementations of the <see cref="IOpenIddictApplicationStore{TApplication}"/> interface when your secret is already hashed.
+    /// </remarks>
+    public class NullClientSecretHasher : IOpenIddictClientSecretHasher
+    {
+        /// <inheritdoc />
+        public ValueTask<string> ObfuscateClientSecretAsync(string secret, CancellationToken cancellationToken = default)
+        {
+            return new ValueTask<string>(secret);
+        }
+
+        /// <inheritdoc />
+        public ValueTask<bool> ValidateClientSecretAsync(string secret, string comparand, CancellationToken cancellationToken = default)
+        {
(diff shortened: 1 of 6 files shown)


src/OpenIddict.Abstractions/Managers/NullClientSecretHasher.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <inheritdoc />
+        public ValueTask<string> ObfuscateClientSecretAsync(string secret, CancellationToken cancellationToken = default)
+        {
+            return new ValueTask<string>(secret);
+        }
+
+        /// <inheritdoc />
+        public ValueTask<bool> ValidateClientSecretAsync(string secret, string comparand, CancellationToken cancellationToken = default)
+        {
+            var equals = string.Equals(secret, comparand);
```
kevinchalet, line 29: Such equality checks are optimized to return `false` as soon as the first non-matching char is encountered, which makes this implementation prone to statistical/timing attacks. It's definitely not something that should be provided by OpenIddict itself.

kevinchalet (commented):
Thanks for your PR 😃 

While it's not clear to me how it helps make things clearer compared to the current approach, I'm actually much more concerned by the `NullClientSecretHasher` you added for two reasons:
  - It doesn't use a time-constant comparison, which makes it vulnerable to timing attacks. The current OpenIddict implementation doesn't suffer from this issue as it uses a time-constant comparison.
  - Comparing the 2 client secret hashes makes your app vulnerable to pass-the-hash attacks (the hash becomes the secret). The OpenIddict implementation doesn't have this issue, as it requires sending the real secret and not a hash (also, it's not a simple hash but the output of a PBKDF, which slows down the secret validation and helps mitigate brute force attacks).

Closing this PR for now.

SebastianStehle:
I can just remove the null-implementation, but it would it make easier to override it. What do you mean with "timing attacks"?

It is just a minor improvements, I know, but I think that the current application manager makes more high-level stuff and should delegate this low-level things to other services anyway. Like password policies and so on the UserManager from ASP.NET Core Identity.

kevinchalet:
> but it would it make easier to override it

Not making it easy to override was actually deliberate: folks are often tempted to customize things and this one is hard to customize "safely", which is why it's only doable by subclassing the manager itself.

> What do you mean with "timing attacks"?

Here's a good overview of what timing attacks are: https://en.wikipedia.org/wiki/Timing_attack

SebastianStehle:
> Not making it easy to override was actually deliberate: folks are often tempted to customize things and this one is hard to customize "safely", which is why it's only doable by subclassing the manager itself.

I see. It is the only way that I found to be compatible with my current identity-server implementation, because the secrets are from another databases (or multiple databases actually). I guess another approach could be to obfuscate the secrets on the fly, when they are loaded from the database. But they I would access to the hasher algorithm as well and I am not sure what it means for performance.

PR #1291, opened 2021-07-01
Update the ID0112 and ID0166 messages to help debug invalid ASP.NET Core configurations

kevinchalet:


kevinchalet:
Note: this change will be backported to 3.1.

PR #1303, opened 2021-08-22
Avoid developer errors that changes options

pksorensen:
I argue that the options should be a `IOptionsSnapshot<OpenIddictServerOptions>` instead, which would give a clean `OpenIddictServerOptions` for each request. 

This would not allow the developer to misconfigure critical settings on the option by altering them thinking its per client change.

Like in this example: 
image

But also applies for token life times and such.

I am working on a UI where its possible to set alot of settings per client and would like to be able to alter settings `IOpenIddictServerHandler<HandleTokenRequestContext>` dynamically, but in a safe way.

### DIFF
--- a/src/OpenIddict.Server/OpenIddictServerFactory.cs
+++ b/src/OpenIddict.Server/OpenIddictServerFactory.cs
@@ -13,14 +13,14 @@ namespace OpenIddict.Server
     public class OpenIddictServerFactory : IOpenIddictServerFactory
     {
         private readonly ILogger _logger;
-        private readonly IOptionsMonitor<OpenIddictServerOptions> _options;
+        private readonly IOptionsSnapshot<OpenIddictServerOptions> _options;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


kevinchalet:
Thanks for your PR.

Unfortunately, `IOptionsSnapshot<T>` would force options to be re-computed for each request, which can be a very expensive operation and the reason why it's used almost nowhere in ASP.NET Core itself.

pksorensen:
Do you want an update where this is configurable by the options itself `UseOptionsPerRequest` or a better name? where the implementation resolve `IOptionsSnapshot` if set to true or otherwise keeps the Monitor?

Otherwise i will just keep the implementation in project only.

kevinchalet:
Your scenario will be supported in 4.0 thanks to a new `GenerateToken` event whose `GenerateTokenContext` will allow you to set the encryption/signing credentials dynamically. In the meantime, your scenario can be implemented using the events model, without using `IOptionsSnapshot<T>`.

PR #1304, opened 2021-08-22
Revamp the token authentication/generation mechanism and introduce new events

kevinchalet:
This PR revamps the authentication/generation mechanism by introducing two new sub-events: `ValidateToken` and `GenerateToken`, that are called by `ProcessAuthentication` and `ProcessSignIn`.

These events allow controlling what security token handler and what signing/encryption credentials are used to process tokens. They are designed to be called multiple times per request (such support will be needed for token exchange).


kevinchalet:
Note: this PR also moves the "token usage" validation made during introspection response handling higher in the stack. It's a change I plan to document in the migration guide.

Issue #1346, opened 2021-10-30
Update the code base to support assembly trimming and NativeAOT


kevinchalet:
https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming

kevinchalet:
OpenIddict doesn't currently work with NativeAOT due to the fact it relies on `System.Text.Json` but doesn't use source-generated serialization (which is required as reflection-based serialization isn't used on NativeAOT).

In OpenIddict 7.0, we'll want to support this scenario and make all the assemblies trimmer-friendly.

PR #1396, opened 2022-03-03
Introduce a web providers companion package

kevinchalet:
This PR contains the prototype of a companion package for the new OpenIddict client that could be used as a potential alternative to the existing aspnet-contrib OAuth 2.0 providers for ASP.NET Core, as announced in this blog post: https://kevinchalet.com/2022/02/25/introducing-the-openiddict-client/

While certainly not definitive (I'm open to all suggestions), the design of this companion package _tries_ to reduce the maintenance burden seen with the aspnet-contrib providers:

  -  Instead of having a package per provider, I opted for a single/unified package including all providers. This not only avoids the issues we saw with package validation (that makes adding new packages a bit painful) but also allows sharing event handlers required by multiple providers instead of having to duplicate them. For instance, here's how the scope reformatting logic for providers that use comma-separated scopes can be implemented for multiple providers at the same time:

```csharp
context.Request.Scope = context.Registration.GetProviderName() switch
{
    // The following providers are known to use comma-separated scopes instead of
    // the standard format (that requires using a space as the scope separator).
    Providers.Deezer or
    Providers.Facebook or
    Providers.Reddit
        when context.GrantType is GrantTypes.AuthorizationCode or GrantTypes.Implicit
        => string.Join(",", context.Scopes),

    _ => context.Request.Scope
};
```

  - The companion package doesn't reference ASP.NET Core or `OpenIddict.Client.AspNetCore` to ensure it remains fully host-agnostic. This is how we can guarantee that these providers will also be usable on OWIN/Blazor/MAUI/WPF/UWP (once the core OpenIddict client has integration packages for these environments, of course 😄).
 
  - The package references `OpenIddict.Client.SystemNetHttp` to be able to customize HTTP requests as they are sent over the wire. This is not a problem as all the web providers use HTTP anyway.

  - No endpoint URL constant is deliberately exposed: the OpenIddict client is compatible with both dynamic (via discovery) and static configurations and not exposing these addresses allows migrating providers that use a static configuration to a dynamic one without introducing a breaking change.

  - There's no claims mapping at all: all the claims resolved from the backchannel/frontchannel identity tokens and userinfo responses/tokens are returned as-is in a merged principal by `AuthenticateAsync()` and it's up to the developer to preserve the needed claims and/or implement a custom mapping. This differs from the aspnet-contrib providers, that try to map a few claims to their `ClaimTypes` WS-Fed equivalent (which proved to be painful).

  - There's deliberately no additional API call to retrieve things that are not part of identity tokens and/or userinfo responses/tokens (e.g the GitHub integration doesn't make a second API call to retrieve the email if it's not returned by the userinfo endpoint).

  - I tested this PR with 3 different providers and they worked fine (the client identifiers/secrets are included if you'd like to test it):
    - GitHub: it uses a static configuration and doesn't require any "hack" to work.
    - Google: it uses a dynamic configuration and doesn't require any workaround.
    - Reddit: it uses a static configuration and requires a little hack to use comma-separated scopes. There's also a custom event handler to make the user agent specific to the application (something we already use in the aspnet-contrib provider).

Note: it's likely this PR won't be merged. It's not clear yet whether it's something we'd want to do or whether we'd do that in this repository (maybe an openiddict-contrib or openiddict-providers repo would be more appropriate?)


src/OpenIddict.Client.WebIntegration/OpenIddict.Client.WebIntegration.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net461;netcoreapp3.1;net5.0;net6.0;netstandard2.0;netstandard2.1</TargetFrameworks>
```
martincostello, line 4: Maybe not bother with 5 as it's out of support in 2 months?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -0,0 +1,123 @@
...
+    public static ImmutableArray<OpenIddictClientHandlerDescriptor> DefaultHandlers { get; } = ImmutableArray.Create(
+        /*
+         * Challenge processing:
+         */
+        FormatNonStandardScopeParameter.Descriptor)
+        .AddRange(Exchange.DefaultHandlers)
+        .AddRange(Userinfo.DefaultHandlers);
+
+    /// <summary>
+    /// Contains the logic responsible of overriding the standard "scope"
```
martincostello:
```suggestion
    /// Contains the logic responsible for overriding the standard "scope"
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -0,0 +1,123 @@
...
+
+                _ => context.Request.Scope
+            };
+
+            return default;
+        }
+    }
+
+    /// <summary>
+    /// Contains the logic responsible of overriding the user agent for providers
```
martincostello:
```suggestion
    /// Contains the logic responsible for overriding the user agent for providers
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -0,0 +1,123 @@
...
+            var request = context.Transaction.GetHttpRequestMessage();
+            if (request is null)
+            {
+                throw new InvalidOperationException(SR.GetResourceString(SR.ID0173));
+            }
+
+            // A few providers (like Reddit) are known to aggressively check user agents and encourage
+            // developers to use unique user agents. While OpenIddict itself always adds a user agent,
+            // the default value doesn't differ accross applications. To reduce the risks of seeing
+            // requests blocked by these providers, the user agent is replaced by a custom value
```
martincostello, line 236: Nit: it appends, not replaces.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -0,0 +1,123 @@
...
+            // containing the product name and version set by the user or by the client identifier.
+            if (context.Registration.GetProviderName() is Providers.Reddit)
```
martincostello, line 238: Personally, I'd just do this for every provider.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationSettings.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// <summary>
+    /// Gets the scopes requested to the authorization server.
+    /// </summary>
+    public virtual HashSet<string> Scopes { get; } = new(StringComparer.Ordinal);
+
+    /// <summary>
+    /// Provides various settings needed to configure the GitHub integration.
+    /// </summary>
+    public class GitHubSettings : OpenIddictClientWebIntegrationSettings
```
martincostello: I'd make the containing class `partial` and define these in a file each. Otherwise this file would get bloated AF as more providers were added.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationSettings.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        // Note: Reddit requires sending at least one scope element. If no scope parameter
+        // is set, a misleading "invalid client identifier" error is returned to the caller.
+        // To prevent that, the "identity" scope is always added by default.
+        public RedditSettings() => Scopes.Add("identity");
+
+        /// <summary>
+        /// Gets or sets the product name, used to build the user agent header.
+        /// </summary>
+        public string? ProductName { get; set; }
```
martincostello: As per previous comment, I'd just make this a general feature, not Reddit-specific.

martincostello:
I see the value in a lightweight _alternative_ to `OAuthHandler` et al, but I'm not yet convinced it's a _replacement_ for it for the _"I just want to do social login in my hobby app"_ use case (which I imagine a lot of the aspnet-contrib contributions come from).

Some specific thoughts:

* Only having one package would make deployment and adding new providers easier, but over time if it scaled to the size of aspnet-contrib OAuth's number of providers, the assembly might get quite chonky/bloated.
* What would the migration story for apps using the OAuthHandler-based providers that don't already use OpenIddict be?
* I think having roles baked-in in aspnet-contrib makes the providers quite easy to plug-in and get going in apps written by people without much/any experience of what's going on under the hood for OAuth. If the amount of custom code needed to be written for feature parity with the existing providers is "too much", then this new client might be less attractive for adoption by users who aren't experts in the authentication stack.

kevinchalet:
@martincostello thanks for the feedback, much appreciated!

> I see the value in a lightweight _alternative_ to `OAuthHandler` et al, but I'm not yet convinced it's a _replacement_ for it for the _"I just want to do social login in my hobby app"_ use case (which I imagine a lot of the aspnet-contrib contributions come from).

Fair point (which is why I indeed described it as an *alternative* in this PR rather than as a *replacement* 😃).

The approach used in this PR certainly differs from how the base OAuth 2.0 handler and the derived implementations we maintain in aspnet-contrib work, so yeah, it's likely not going to be an exact replacement.

> * Only having one package would make deployment and adding new providers easier, but over time if it scaled to the size of aspnet-contrib OAuth's number of providers, the assembly might get quite chonky/bloated.

It's a very legitimate concern. I see two counter-arguments to your remark:

  - Having multiple assemblies is generally less efficient than having a single assembly from a size perspective. It's specially true for the aspnet-contrib providers, that typically don't contain much code (compared to the size of the assembly metadata). As an example, the `OpenIddict.Client.WebIntegration` DLL produced by this PR for the `net6.0` TFM - that includes 3 providers - weighs 17KB while most of the aspnet-contrib binaries are between 15 and 17KB (with a few exceptions, like the Apple provider, which almost doubles that 😄). With 3 aspnet-contrib providers, the resulting size would be 15 x 3 = 45KB.

  - When the aspnet-contrib providers collection was created 7 years ago, things were actually quite different: we now have a good tree shaking/IL trimming story that could be leveraged to remove APIs corresponding to providers that are not actually used by devs in their apps. https://github.com/openiddict/openiddict-core/issues/1346 tracks making the entire OpenIddict codebase trimming-friendly.

> * What would the migration story for apps using the OAuthHandler-based providers that don't already use OpenIddict be?

The OpenIddict client is completely independent from the rest of the stack (just like you can use the validation stuff without the server feature, you can use the client without the server or validation features), so the story would be identical for those who already use OpenIddict and those who don't.

In their current form, the OpenIddict client and this companion package are not drop-in replacements. Technical things that differ:

  - The aspnet-contrib providers use an authentication scheme per provider, which means you can do `[Authorize(AuthenticationSchemes = "Facebook")]` to trigger an authentication dance. In contrast, the OpenIddict client uses a single authentication scheme and requires setting the issuer as an `AuthenticationProperties` item if multiple providers are registered. If we were interested in avoiding a breaking change here, we could use `IAuthenticationHandler` "proxies" to map the `Facebook` authentication scheme to the OpenIddict client handler and set the issuer based on that.

  - For the same reason, the providers registered via the OpenIddict client are currently not listed by Identity's `SignInManager.GetExternalAuthenticationSchemesAsync()` and so don't appear in the "external providers" list returned by the default Identity UI. I have mixed feelings about this thing: both the MSFT and aspnet-contrib providers use a hard-coded English name for `DisplayName`, which is fine for many apps, but some devs will want to localize the provider names, and it's hard to do this way. That said, if we wanted to avoid a breaking change, the same approach as in the previous point could be implemented.

  - The OpenIddict client doesn't do claims mapping so users would need to change their code to use the exact claims returned by the providers instead of the ones we chose in the aspnet-contrib providers (just like the `OAuthHandler`-based implementations maintained by MSFT, some of the claims are inconsistently mapped to their WS-Fed equivalent - i.e represented as URLs - while others tend to use the original names.

  - The OpenIddict client doesn't have the "delegate that `ClaimsPrincipal` instance to the cookie handler so it can create an authentication cookie based on it" logic you have in the built-in OWIN and ASP.NET Core handlers so users will need to add a tiny minimal API handler/MVC action to "handle" the callback part and store the claims they want in their cookies:

https://github.com/openiddict/openiddict-core/blob/cf0e49a3b85fdebff8e57edb86a30a701b9284e6/sandbox/OpenIddict.Sandbox.AspNetCore.Client/Controllers/AuthenticationController.cs#L30-L125

It requires custom code, but on the other hand, I _think_ it makes things less magic by encouraging users to take care of the external identity -> internal identity logic.

> * I think having roles baked-in in aspnet-contrib makes the providers quite easy to plug-in and get going in apps written by people without much/any experience of what's going on under the hood for OAuth. If the amount of custom code needed to be written for feature parity with the existing providers is "too much", then this new client might be less attractive for adoption by users who aren't experts in the authentication stack.

By roles, did you mean claims? If so, would having a static class with the claims returned by the providers make that better?

I agree the aspnet-contrib providers offer a great experience for non-experts... as long as you use ASP.NET Core Identity. If you don't, you need to register an instance of the cookies authentication handler and define it as the default sign-in scheme in the authentication options. You also need to either use automatic authentication in one of the providers to trigger challenges automatically or have an authentication controller that starts them (i.e the approach I used in the aspnet-contrib sample 7 years ago 😅). In that case, you need to at least understand how the ASP.NET Core authentication stack works.

I don't think the OpenIddict client makes that worse: these steps are implemented the same way and most - if not all - the details of the OAuth 2.0 protocol are managed for you by OpenIddict. The 2 entry points exposed to the users are the OpenIddict config in the Startup class and the callback minimal API handler/MVC action that retrieves the external claims and stores them in a cookie (or anywhere else).

src/OpenIddict.Client.WebIntegration/OpenIddict.Client.WebIntegration.csproj
kevinchalet: Makes sense, I opened https://github.com/openiddict/openiddict-core/issues/1397 to remove it globally.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: Woops, I suspect I made the same mistake in tons of places in the existing codebase. Time to make a global find-and-replace pass, I guess 😄 

Thanks.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: The default user agent is added by OpenIddict to `HttpClient.DefaultHeaders` via `HttpClientFactoryOptions.HttpClientActions`. The default user agent to configured in the default headers is always discarded if you explicitly attach a user agent to `HttpRequestMessage.Headers` so while we do call `.Add` here, the original user agent ends up being replaced by the custom one used here.

The aspnet-contrib Reddit provider doesn't use `HttpClientFactory` but I suspect it behaves the same way.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: Yeah, it makes sense. I actually did that to demonstrate how `HttpRequestMessage` can be customized per-provider so doing that "globally" would have kinda defeated that 😅

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationSettings.cs
kevinchalet: Makes total sense.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
martincostello: By default for the OAuth handlers you get [this behaviour](https://github.com/dotnet/aspnetcore/blob/84c8d0f93cc4b6fbce1367ffc366a10037ecee77/src/Security/Authentication/OAuth/src/OAuthPostConfigureOptions.cs#L39-L45) unless you set `Backchannel` yourself.

kevinchalet: > By default for the OAuth handlers you get [this behaviour](https://github.com/dotnet/aspnetcore/blob/84c8d0f93cc4b6fbce1367ffc366a10037ecee77/src/Security/Authentication/OAuth/src/OAuthPostConfigureOptions.cs#L39-L45) unless you set `Backchannel` yourself.

Yeah, it's more or less what the OpenIddict client/System.Net.Http integration does (except it uses `HttpClientFactory`).

I just tested with the aspnet-contrib provider and as I suspected, it behaves exactly the same way: if you configure a user-agent in the options, the default one (configured by the OAuth 2.0 base handler) is simply overwritten:

image

kevinchalet: If you want to preserve the two user agents, you basically need to import the headers attached to `HttpClient`:

```csharp
if (!string.IsNullOrEmpty(Options.UserAgent))
{
    foreach (var header in Backchannel.DefaultRequestHeaders.UserAgent)
    {
        request.Headers.UserAgent.Add(header);
    }

    request.Headers.UserAgent.ParseAdd(Options.UserAgent);
}
```

image

kevinchalet:
So, I started playing with Roslyn source generators and I'm feeling they could be a great fit for this companion package. The idea is that all the plumbing (builder methods, constants and settings classes) could be dynamically generated based on an editable providers file. Providers that are spec-compliant and support configuration discovery (e.g Google) could be supported by just adding a single line in that file. Providers that are spec-compliant but require a static configuration can be easily added too by using statically-defined endpoints.

In this prototype, I opted for XML (personal preference, but open to all suggestions). Here's what it could look like:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<Providers>

  <Provider Name="GitHub" Issuer="https://github.com/">
    <Configuration AuthorizationEndpoint="https://github.com/login/oauth/authorize"
                   TokenEndpoint="https://github.com/login/oauth/access_token"
                   UserinfoEndpoint="https://api.github.com/user" />
  </Provider>

  <Provider Name="Google" Issuer="https://accounts.google.com/" />

  <Provider Name="Reddit" Issuer="https://www.reddit.com/">
    <Configuration AuthorizationEndpoint="https://www.reddit.com/api/v1/authorize"
                   TokenEndpoint="https://www.reddit.com/api/v1/access_token"
                   UserinfoEndpoint="https://oauth.reddit.com/api/v1/me" />

    <Setting Type="string" Name="ProductName" Property=".product_name"
              Description="Gets or sets the product name, used to build the user agent header." />

    <Setting Type="string" Name="ProductVersion" Property=".product_version"
              Description="Gets or sets the product version, used to build the user agent header." />
  </Provider>

</Providers> 
```

kevinchalet:
Native OWIN support was added in the main branch so I updated this PR to demonstrate how this companion package is going to be used with ASP.NET 4.8.

kevinchalet:
I updated the generator to support multiple environments per provider (something currently used in the aspnet-contrib SuperOffice and Salesforce providers) and per-user userinfo endpoints that need to be resolved dynamically (also used in the Salesforce provider)

Note: I added Salesforce to the providers list but it's unlikely to stay in its current form as:
  - Salesforce now support OpenID Connect configuration discovery so we don't need to hardcode the endpoints.
  - It seems they now offer a native OpenID Connect userinfo endpoint that doesn't require using the dynamic userinfo thing.

kevinchalet:
Next step: updating the generator to support multitenant providers (that typically require setting a subdomain, a custom domain or a path base to be able to compute the actual endpoint addresses).

kevinchalet:
I updated the generator to support multitenant providers like Zendesk:

```xml
<Provider Name="Zendesk">
  <!--
    Note: Zendesk is a multitenant provider that relies on subdomains to identify instances.
    As such, the following URLs all include a {tenant} placeholder that will be dynamically
    replaced by OpenIddict at runtime by the tenant configured in the Zendesk settings.
  -->

  <Environment Name="Production" Issuer="https://{tenant}.zendesk.com/">
    <Configuration AuthorizationEndpoint="https://{tenant}.zendesk.com/oauth/authorizations/new"
                   TokenEndpoint="https://{tenant}.zendesk.com/oauth/tokens"
                   UserinfoEndpoint="https://{tenant}.zendesk.com/api/v2/users/me" />
  </Environment>

  <Setting Type="string" Name="Tenant" Property=".tenant"
           Description="Gets or sets the tenant used to identify the Zendesk instance." />
</Provider>
```

Under the hood, it uses the excellent SmartFormat library (released under the MIT license) to format endpoint URLs based on the settings configured by the user.

kevinchalet:
Provider generator updated to support manual configuration of PKCE and client authentication methods. E.g for Fitbit, that supports PKCE and requires sending the client credentials in the `Authorization` header:

```xml
<Provider Name="Fitbit">
  <Environment Issuer="https://www.fitbit.com/">
    <Configuration AuthorizationEndpoint="https://www.fitbit.com/oauth2/authorize"
                   TokenEndpoint="https://api.fitbit.com/oauth2/token"
                   UserinfoEndpoint="https://api.fitbit.com/1/user/-/profile.json">
      <CodeChallengeMethod Value="S256" />

      <TokenEndpointAuthMethod Value="client_secret_basic" />
    </Configuration>
  </Environment>
</Provider>
```

kevinchalet:
After playing extensively with source generation, I think it's the best approach for this project as it will allow generating tons of providers without having to maintain dozens/hundreds of plumbing classes, which will help keep the maintenance burden quite low (my biggest concern when I started working on this prototype 😄) 

So it's time for the next step.

Things to do:
  - Polish the provider generator (e.g to generate null checks for "required" settings).

  - Find the best home for `OpenIddict.Client.WebIntegration`. A dedicated repo would make sense but keeping it in `openiddict-core` has an advantage: we can reference the companion package from the `OpenIddict` metapackage very easily since they are built at the same time.

  - Make an exhaustive list of the existing aspnet-contrib providers and create categories based on their particularities:
      - Providers that support OpenID Connect. 
      - Providers that require `client_secret_basic`.
      - Providers that require sending a `scope` with one or multiple "default" values.
      - Providers that use multitenancy.
      - Providers that use dynamic endpoints.
      - Providers that use custom parameters similar to the standard ones (e.g `app_id` instead of `client_id`).
      - Providers that use POST for the userinfo endpoint.
      - Providers that use GET for the token endpoint.
      - Providers that implement "request signing".
      - Providers that require a special treatment for `client_secret` (e.g Apple or Visual Studio Online).
 
  - Based on that list, ensure the OpenIddict client implements all the hooks necessary to deal with these requirements.

  - Write documentation (or better, a flow chart?) indicating how to add a new provider, which should include the following steps:
      - Does the provider offer multiple environments? (e.g production, development, staging). If so, the provider MUST have multiple `<Environment />` nodes with the corresponding configuration.
      - Does the provider expose a configuration document for the supported environments? If so, the provider MUST use discovery instead of static configuration for all environments that support it.
      - If static configuration is used, does the provider support PKCE?
      - If static configuration is used, does the provider require using `client_secret_basic`?
      - Does the provider require custom code to accommodate to non-standard behaviors?

  - Create a GitHub issue template dedicated to adding new providers that will point to that documentation. Such requests should probably be opened to non-sponsors to increase contributions.

  - Post an announcement in the aspnet-contrib OAuth 2.0 repo with a link to this project and the differences between `OpenIddict.Client.WebIntegration` and the existing aspnet-contrib providers.

@martincostello assuming we'll still want to develop and maintain the aspnet-contrib providers in parallel, it's likely the information collected while adding new OpenIddict providers will be useful for aspnet-contrib. E.g:

  - Fitbit now supports PKCE but it's not enabled in the aspnet-contrib Fitbit provider by default.
  - Reddit now supports standard space-separated scopes, so we're no longer forced to use commas.

martincostello:
At first read-through (I'll take it in properly tomorrow) that sounds a reasonable plan. Cross-referencing back to aspnet-contrib sounds like it could feed into the v7 release where things can be improved (e.g. Reddit PKCE) without worrying about breaking changes.

When this is more ""ready"" I've a handful of personal repos using the aspnet-contrib OAuth providers (Amazon, Apple, GitHub) they could test it out with.

kevinchalet:
The latest commit added an XML schema to make the providers file much easier to work with:

image

kevinchalet:
The latest commit removed the event handler responsible for adding the default scopes and moved its logic to the provider-specific configuration classes, that will now support adding default scopes (that are automatically added if no other scope is set) and required scopes (that are always added even if a scope was set by the user).

image

kevinchalet:
Alright, I'm quite happy with the result so it's time to merge this PR. I'll open follow-up tickets to track the remaining tasks.

kevinchalet:
> I'll open follow-up tickets to track the remaining tasks.

- https://github.com/openiddict/openiddict-core/issues/1451
- https://github.com/openiddict/openiddict-core/issues/1452
- https://github.com/openiddict/openiddict-documentation/issues/42

PR #1404, opened 2022-03-29
Add status param to FindByAuthorizationId at TokenStore

jersiovic:
Trying to improve a bit performance I've added an override for FindByAutorizationIdAsync method at IOpenIddictTokenStore.
To help us to only load from db valid tokens.

### DIFF
--- a/src/OpenIddict.Abstractions/Caches/IOpenIddictTokenCache.cs
+++ b/src/OpenIddict.Abstractions/Caches/IOpenIddictTokenCache.cs
@@ -70,6 +70,15 @@ public interface IOpenIddictTokenCache<TToken> where TToken : class
     IAsyncEnumerable<TToken> FindByAuthorizationIdAsync(string identifier, CancellationToken cancellationToken);
 
     /// <summary>
+    /// Retrieves the list of tokens corresponding to the specified authorization identifier and token status.
+    /// </summary>
+    /// <param name="identifier">The authorization identifier associated with the tokens.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
+    /// /// <param name="status">The token status.</param>
+    /// <returns>The tokens corresponding to the specified authorization.</returns>
+    IAsyncEnumerable<TToken> FindByAuthorizationIdAsync(string identifier, string status, CancellationToken cancellationToken);
+
+    /// <summary>
     /// Retrieves a token using its unique identifier.
     /// </summary>
     /// <param name="identifier">The unique identifier associated with the token.</param>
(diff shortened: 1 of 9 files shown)


kevinchalet:
Thanks for your PR!

Not filtering based on the current status of tokens was actually deliberate to ensure all tokens - not just those with the `valid` status - can be marked as revoked (e.g so that clients get appropriate errors when trying to use them).

Can you tell me more about the perf' issues you were trying to solve?

jersiovic:
> Thanks for your PR!
> 
> Not filtering based on the current status of tokens was actually deliberate to ensure all tokens - not just those with the `valid` status - can be marked as revoked (e.g so that clients get appropriate errors when trying to use them).
> 
I understand, that's why I provide an alternate override for using only when it is useful.
For example here https://github.com/openiddict/openiddict-core/blob/4d33e0f3e8dd01039ea2a25027499fc7a84b4df3/src/OpenIddict.Server/OpenIddictServerHandlers.Protection.cs#L924 you don't need to retrieve from the database all the access token and id_token which are not in a valid status. In Orchard core that means to load not valid records from last 14 days that is the default threshold for the background task which cleans tokens.
In fact, in my app I reject all the tokens related to an authorization when a new request of password flow happen. So I would like to don't have to load all not valid tokens from db.

> Can you tell me more about the perf' issues you were trying to solve?

In my scenario I started to suffer timeouts using an azure sql server instance with a Basic service  tier when the tokens table started to grow. That's why I send this other pr to Orchard https://github.com/OrchardCMS/OrchardCore/pull/11382 . 
However I tried to find other reasons why this could happen and I found this extra load of not needed records like sth which could be improved

kevinchalet:
> you don't need to retrieve from the database all the access token and id_token which are not in a valid status.

The API you mentioned in used in **only one** place: in the file you linked. So when I mentioned that not filtering based on the status was deliberate, I had this exact place in mind: marking all tokens - not just the ones with a "valid" status - is 100% deliberate and is a security measure.

> So I would like to don't have to load all not valid tokens from db.

This should be rather rare: this API is only called if OpenIddict detects that a one-time token (e.g an authorization code or a refresh token) is replayed by a client. In this case, it needs to revoke the entire chain to ensure none of the already issued token can be used. Does that ring a bell?

> In Orchard core that means to load not valid records from last 14 days that is the default threshold for the background task which cleans tokens.

Maybe we should simply update OC to use a shorter interval?

jersiovic:
The API I mentioned is used in **only one** place: in your code, but it is also used **in my code**. Because when a new password flow or client credentials request happens, just before the new token is generated I reject all previous tokens for that user. 

I though I don't have any other way to improve performance for those queries, because I would like to use not only db access also I would like to use same internal cache openiddict uses

> Maybe we should simply update OC to use a shorter interval?

This sounds me sensible but complementary. One thing should not skip the other

PR #1476, opened 2022-06-24
Decode Basic Auth as UTF-8

Suchiman:
After upgrading from `OAuthAuthorizationServerMiddleware`, some clients failed to authenticate which, after some bruteforcing (would be handy if you could set an option not to `[redact]` secrets for debugging purposes), i've managed to track back to non ASCII characters in their ClientSecret, like `§`.
[OAuthAuthorizationServerMiddleware used UTF-8](https://github.com/vairam-svs/katana/blob/9c69a9d81f28b79e7f3aa008e98256f9e095fb70/src/Microsoft.Owin.Security.OAuth/Provider/OAuthValidateClientAuthenticationContext.cs#L62) and [modern browsers like Firefox and Chrome also encode basic auth using UTF-8](https://stackoverflow.com/a/7243567) so i propose to change to decoding it as UTF-8 which should not break compat as UTF-8 is a superset of ASCII.

Btw. thank you for this awesome project, very impressive and extensible architecture!

### DIFF
--- a/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.cs
+++ b/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.cs
@@ -752,7 +752,7 @@ public static partial class OpenIddictServerAspNetCoreHandlers
             try
             {
                 var value = header.Substring("Basic ".Length).Trim();
-                var data = Encoding.ASCII.GetString(Convert.FromBase64String(value));
+                var data = Encoding.UTF8.GetString(Convert.FromBase64String(value));
 
                 var index = data.IndexOf(':');
                 if (index < 0)


kevinchalet:
Thanks for your PR, @Suchiman!

> i've managed to track back to non ASCII characters in their ClientSecret, like §.
[OAuthAuthorizationServerMiddleware used UTF-8](https://github.com/vairam-svs/katana/blob/9c69a9d81f28b79e7f3aa008e98256f9e095fb70/src/Microsoft.Owin.Security.OAuth/Provider/OAuthValidateClientAuthenticationContext.cs#L62) and [modern browsers like Firefox and Chrome also encode basic auth using UTF-8](https://stackoverflow.com/a/7243567) so i propose to change to decoding it as UTF-8 which should not break compat as UTF-8 is a superset of ASCII.

It's important to note that unlike `OAuthAuthorizationServerMiddleware`, [OpenIddict implements the standard flavor of `client_secret_basic` since 2018](https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/515), that requires formurl-encoding the `client_id` and `client_secret` before base64-encoding them:

> The client identifier is encoded using the "application/x-www-form-urlencoded" encoding algorithm per [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B), and the encoded value is used as the username; the client password is encoded using the same algorithm and used as the password.

https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

When implementing the standard encoding logic, using UTF-8 instead of ASCII shouldn't make any difference, as all non-ASCII chars would be replaced using percent-encoding as defined in application/x-www-form-urlencoded before being base64-encoded.

If things broke for you when moving to OpenIddict, it's very likely because some of the clients targeting your OAuth 2.0 service are not compliant, which is an extremely common case and the reason why OpenIddict itself will try to favor `client_secret_post` when available:

https://github.com/openiddict/openiddict-core/blob/fdf99fee691baab5285f331c64be22f2159699c0/src/OpenIddict.Client.SystemNetHttp/OpenIddictClientSystemNetHttpHandlers.Exchange.cs#L83-L111

Based on that, I can see two options:
  - If you can, fix the non-standard clients to do the proper thing.
  - If you can't, create your own event handler as you did for this PR, remove the "official" one and register your custom handler at the same place in the pipeline.

kevinchalet:
> Btw. thank you for this awesome project, very impressive and extensible architecture!

Thanks for the kind words! 😃

Suchiman:
Mhm i see, but wouldn't it make sense to use UTF-8 nontheless to appease the broken clients, or even just plain OAuth2 clients, as this should not affect the correctly behaving OIDC clients? They would still potentially be broken if they use characters like `+` or `%` but they're... less broken in that case? One could detect any characters outside of the ASCII range and then avoid UrlDecoding but that can't fix the issue if you have an ASCII secret with unencoded `+%` characters...

> If you can, fix the non-standard clients to do the proper thing.

Not really an option for me unfortunately, also third party clients involved.
I've discovered that at least `swagger-ui` (from swashbuckle) does it wrong as does `Thinktecture.IdentityModel.Client`

> If you can't, create your own event handler as you did for this PR, remove the "official" one and register your custom handler at the same place in the pipeline.

Yep that's what i did, thanks to the very extensible architecture again :>

Anyway, if you do think this is not going in the right direction, feel free to close

kevinchalet:
> Mhm i see, but wouldn't it make sense to use UTF-8 nontheless to appease the broken clients, or even just plain OAuth2 clients, as this should not affect the correctly behaving OIDC clients?

It's something I indeed considered 4 years ago, but decided against it because:
  - OpenIddict tries to stay as close to the specification as possible.
  - Most vendors affected by this issue that have switched to OIDC haven't fixed it when migrating so discriminating between OAuth 2.0 and OIDC clients wouldn't really help.
  - Modern browsers tend to use UTF-8 but it's not true for most headless clients, that are the main targets for client authentication (it's extremely rare to use client secrets in browser apps, except when dealing with legacy clients or when implementing dynreg) so the audience for such a change would be actually quite limited.
  - As you said, it couldn't cover all cases 😢 
  - It basically hides a defect in the client implementation instead of encouraging users to fix it.

Since the change was made 4 years ago, it's the first report I get, so it's likely the issue is not widespread. I'd feel less reluctant to go against the spec' if there were more people impacted.

> I've discovered that at least `swagger-ui` (from swashbuckle) does it wrong as does `Thinktecture.IdentityModel.Client`

Are you still using a very old version of `Thinktecture.IdentityModel.Client`? They fixed their implementation in 2018 with a few other client vendors, which motivated me to adopt the standard encoding in ASOS/OpenIddict so it would be surprising if their recent bits were impacted.

Regarding swagger-ui, we should definitely let them know they don't use a standard encoding so they have a chance to fix it. Do you mind opening a ticket on their repo?

> Yep that's what i did, thanks to the very extensible architecture again :>

Haha yeah, dealing with non-standard stuff is basically the _raison d'être_ of the events model (that was originally inspired by `OAuthAuthorizationServerMiddleware`'s notification mechanism but got a massive overhaul in OpenIddict 3.0) 😃

Suchiman:
> Are you still using a very old version of Thinktecture.IdentityModel.Client? They fixed their implementation in 2018 with a few other client vendors, which motivated me to adopt the standard encoding in ASOS/OpenIddict so it would be surprising if their recent bits were impacted.

It's version `4.0.1` from 2015, which is the newest version available of that package. I think https://github.com/IdentityModel/IdentityModel.OidcClient is the successor to that library from the same people but the issue i ran into was, that it requires OIDC and doesn't work with plain OAuth2, so once i've upgraded all environments with openiddict, i might look into adopting that again (or switch directly to your recently announced new client lib 😄)

> Regarding swagger-ui, we should definitely let them know they don't use a standard encoding so they have a chance to fix it. Do you mind opening a ticket on their repo?

Opened https://github.com/swagger-api/swagger-ui/issues/8073

kevinchalet:
> It's version 4.0.1 from 2015, which is the newest version available of that package. I think https://github.com/IdentityModel/IdentityModel.OidcClient is the successor to that library from the same people but the issue i ran into was, that it requires OIDC and doesn't work with plain OAuth2

Ah yeah, the newer one indeed requires an OIDC server (Dominick received a few demands to make it compatible but he has always refused).

> or switch directly to your recently announced new client lib 😄

Out of curiosity, on what environment(s) are you using the IdentityModel lib? The OpenIddict client currently supports ASP.NET Core and Katana and I'm trying to evaluate what should come next 😄

Suchiman:
ASP.NET Core, which is why i had to migrate off `OAuthAuthorizationServerMiddleware`, one big boom bang migration from ASP.NET MVC 5, WebAPI 2, EF6, Identity and OAuth2 to ASP.NET Core 6, EFC6, Identity Core and Openiddict (and to appease satan, i've ported MachineKey to Core so i can read old school tokens on core, for which i've forked Openiddict.DataProtection and replaced its innards with MachineKey). That and we need to SSO with a third party next that only supports OIDC for SSO purposes.

kevinchalet:
Sounds fun! 😄 

> and to appease satan, i've ported MachineKey to Core so i can read old school tokens on core, for which i've forked Openiddict.DataProtection and replaced its innards with MachineKey

FYI, the token generation/validation process got an overhaul in 4.0 with new `GenerateToken`/`ValidateToken` events that are used in the Data Protection integration so it's likely you'll need to make a few changes when migrating to 4.0.

PR #1488, opened 2022-07-19
Fix logging statement

Suchiman:
The logging statement is defined as
> The device request was rejected because the mandatory '{Parameter}' parameter was missing.

but did not pass what Parameter was missing


kevinchalet:
Thanks for your PR. Your fix will be included in the next preview 😃

PR #1493, opened 2022-08-04
Provide real IAsyncEnumerable implementation for EF6

Suchiman:
Provides a real streaming IAsyncEnumerable implementation for EF6.
Noticed it was calling ToListAsync by chance and i already had a proper implementation at hand :-)
The implementation was lifted from https://github.com/dotnet/ef6/blob/c4897b9cf82589b3fe46099a08c930015def0031/src/EntityFramework/Infrastructure/IDbAsyncEnumerableExtensions.cs#L18-L47 which is being used by `ToListAsync()` internally.

### DIFF
--- a/src/OpenIddict.EntityFramework/OpenIddictEntityFrameworkHelpers.cs
+++ b/src/OpenIddict.EntityFramework/OpenIddictEntityFrameworkHelpers.cs
@@ -4,6 +4,7 @@
  * the license and the contributors participating to this project.
  */
 
+using System.Data.Entity.Infrastructure;
 using System.Runtime.CompilerServices;
 using Microsoft.Extensions.DependencyInjection;
 using OpenIddict.EntityFramework;
@@ -60,12 +61,12 @@ public static class OpenIddictEntityFrameworkHelpers
     }
 
     /// <summary>
-    /// Executes the query and returns the results as a non-streamed async enumeration.
+    /// Executes the query and returns the results as a streamed async enumeration.
     /// </summary>
     /// <typeparam name="T">The type of the returned entities.</typeparam>
     /// <param name="source">The query source.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
-    /// <returns>The non-streamed async enumeration containing the results.</returns>
+    /// <returns>The streamed async enumeration containing the results.</returns>
     internal static IAsyncEnumerable<T> AsAsyncEnumerable<T>(this IQueryable<T> source, CancellationToken cancellationToken)
     {
         if (source is null)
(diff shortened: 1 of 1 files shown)


src/OpenIddict.EntityFramework/OpenIddictEntityFrameworkHelpers.cs
```diff
@@ -77,9 +78,24 @@ internal static IAsyncEnumerable<T> AsAsyncEnumerable<T>(this IQueryable<T> sour
...
+                cancellationToken.ThrowIfCancellationRequested();
+
+                if (await enumerator.MoveNextAsync(cancellationToken))
+                {
+                    Task<bool> moveNextTask;
+                    do
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        var current = enumerator.Current;
+                        moveNextTask = enumerator.MoveNextAsync(cancellationToken);
```
kevinchalet: That's an interesting construct (likely used to work around limitations in the old C# versions), but I'm wondering if something like this wouldn't be simpler and more readable:

```csharp
static async IAsyncEnumerable<T> ExecuteAsync(IQueryable<T> source, [EnumeratorCancellation] CancellationToken cancellationToken)
{
    using var enumerator = ((IDbAsyncEnumerable<T>) source).GetAsyncEnumerator();

    while (await enumerator.MoveNextAsync(cancellationToken))
    {
        yield return enumerator.Current;
    }
}
```

kevinchalet (commented):
Thanks for your PR 😃

src/OpenIddict.EntityFramework/OpenIddictEntityFrameworkHelpers.cs
Suchiman: I think this is a performance tweak where the async operation to fetch the next item is already kicked off before returning the current item.

kevinchalet: Makes sense, tho' I'm not sure it's worth it (accessing `enumerator.Current` and yield-returning it should be a super cheap operation) 😄

Suchiman: It is but eliminating the temp by doing
```c#
moveNextTask = enumerator.MoveNextAsync(cancellationToken);
yield return enumerator.Current;
```
is not safe, as `MoveNextAsync` might complete synchronously and would otherwise be susceptible to races.
And in the current way, this increases the chance that it might return `enumerator.Current` synchronously if the MoveNextAsync Task completed before the IAsyncEnumerable has been asked to yield the next item.

Suchiman: You're right ofc that the method looks a bit dated, i cleaned it up a bit

kevinchalet: > It is but eliminating the temp by doing
> 
> ```cs
> moveNextTask = enumerator.MoveNextAsync(cancellationToken);
> yield return enumerator.Current;
> ```

Ah yeah, I would definitely not suggest doing that for the reasons you suggested 😄 

What you ended with is not a bad idea, but I'm not convinced there's any perf' gain to expect by starting the `enumerator.MoveNextAsync(cancellationToken)` operation before doing the `yield return` dance.

To keep things simpler, let's opt for the approach I suggested in my OP, as it's the closest to what a real `await foreach` would look like (https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8#a-tour-through-async-enumerables).

kevinchalet:
BTW, I'm surprised no one at dotnet/ef6 has suggested that `IAsyncEnumerable` (the BCL type) be supported OOTB. I guess the chances are low, but we never know? 😄

Suchiman:
@kevinchalet yeah IIRC dotnet/ef6 recently announced that they'll not take any contribution that isn't a critical bug fix anymore.

kevinchalet:
Erf, too bad. It's sad because EF6 is still massively used and could benefit from some love... 😭

kevinchalet:
Merged, thanks again for your contribution 🎉

PR #1508, opened 2022-08-25
Add Trigger Identity to avoid duplicating the Trigger

Suchiman:
As there was no identity given, a Trigger with a random guid was created everytime the application started.
While the Job has the `[DisallowConcurrentExecution]` attribute, that does not seem to work when there are multiple triggers and the application is scaled out, in my case i found 3 concurrent executions (equal to the number of scale out instances we have) of the token pruning query running in parallel and running every minute or so, completely bogging down production (there were 173 trigger in the table when we've noticed).

I'm currently checking if i can optimize the token pruning query, the current issue is that there are no indexes supporting that query but the current shape is non-sargable too (inequality for status, OR predicate across join). The result is a clustered index scan at repeateable read isolation, essentially locking the entire table for modifications.
image
image

### DIFF
--- a/src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
+++ b/src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
@@ -34,6 +34,7 @@ public class OpenIddictQuartzConfiguration : IConfigureOptions<QuartzOptions>
             // reached if the application is shut down or recycled. As such, this trigger is set up to fire
             // between 1 and 10 minutes after the application starts to ensure the job is executed at least once.
             builder.ForJob(OpenIddictQuartzJob.Identity)
+                   .WithIdentity(OpenIddictQuartzJob.Identity.Name, OpenIddictQuartzJob.Identity.Group)
                    .WithSimpleSchedule(options => options.WithIntervalInHours(1).RepeatForever())
                    .WithDescription(SR.GetResourceString(SR.ID8002))
                    .StartAt(DateBuilder.FutureDate(new Random().Next(1, 10), IntervalUnit.Minute));
--- a/src/OpenIddict.Quartz/OpenIddictQuartzJob.cs
+++ b/src/OpenIddict.Quartz/OpenIddictQuartzJob.cs
@@ -38,7 +38,7 @@ public class OpenIddictQuartzJob : IJob
     /// Gets the default identity assigned to this job.
     /// </summary>
     public static JobKey Identity { get; } = new JobKey(
-        name: typeof(OpenIddictQuartzJob).Name,
+        name: nameof(OpenIddictQuartzJob),
         group: typeof(OpenIddictQuartzJob).Assembly.GetName().Name!);
 
     /// <inheritdoc/>


src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
```diff
@@ -34,6 +34,7 @@ public void Configure(QuartzOptions options)
             // reached if the application is shut down or recycled. As such, this trigger is set up to fire
             // between 1 and 10 minutes after the application starts to ensure the job is executed at least once.
             builder.ForJob(OpenIddictQuartzJob.Identity)
+                   .WithIdentity(OpenIddictQuartzJob.Identity.Name, OpenIddictQuartzJob.Identity.Group)
```
kevinchalet, line 37: @lahma hey Marko 👏🏻 

I have a question for you, if you don't mind: instead of using "programmatic names" for the names of the jobs/triggers, would it make sense to use better-looking names so they are more... "human-readable"? Something like:
  - `JobKey.Name`: OpenIddict job
  - `JobKey.Group`: OpenIddict/Quartz.NET integration
  - `TriggerKey.Name`: OpenIddict built-in trigger
  - `TriggerKey.Group`: OpenIddict/Quartz.NET integration

https://www.quartz-scheduler.net/documentation/best-practices.html#job-tips doesn't have any specific recommendation regarding names, but I'm wondering if it wouldn't be a good idea for users that configure their Quartz job via an UI like Quartzmin, where the job/trigger key/group will be displayed. Is it something you'd recommend/encourage?

Thanks!

kevinchalet:
@Suchiman thanks for your PR. I'll make sure it's included in 4.0 (whose RTM will ship next month).

> While the Job has the [DisallowConcurrentExecution] attribute, that does not seem to work when there are multiple triggers and the application is scaled out, in my case i found 3 concurrent executions (equal to the number of scale out instances we have)

@lahma can you please confirm it's the expected behavior and that using a persistent trigger name will fix that?

src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
lahma: Well I don't have and solid advice, but things to consider:

* group is very useful concept that you can also to do batch operations like pause all and resume all in group, so sharing a group name like you have in you configuration example probably is the most useful thing, the actual contents don't matter that much - there's also "starts with" kind of queries but probably don't apply here
* triggers and jobs also have `Description` which might also be usable if shown somewhere

lahma:
> @Suchiman thanks for your PR. I'll make sure it's included in 4.0 (whose RTM will ship next month).
> 
> > While the Job has the [DisallowConcurrentExecution] attribute, that does not seem to work when there are multiple triggers and the application is scaled out, in my case i found 3 concurrent executions (equal to the number of scale out instances we have)
> 
> @lahma can you please confirm it's the expected behavior and that using a persistent trigger name will fix that?

`[DisallowConcurrentExecution] ` should block job from running simultaneously if scheduler is configured using database and `clustered = true`. Also bear in mind that it's not CLR type that is blocked - it's the job key. Given two jobs sharing the type and different job keys, they might run at the same time as concrete lock is based on job's name and group.

Suchiman:
@lahma
> `[DisallowConcurrentExecution] ` should block job from running simultaneously if scheduler is configured using database and `clustered = true`.

It's been a few months and I'm currently out of office but from what I've remember, that's what we did and what we ended up with, were three servers, running the same job key but each triggered by a different, over due, trigger.

src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
kevinchalet: Great, thanks a lot for the insights!

We already assigned the job a group and a description, but the trigger itself didn't have a static `TriggerKey` :+1:

kevinchalet:
@Suchiman I integrated your work in https://github.com/openiddict/openiddict-core/pull/1573.

Thanks for your contribution 😃

PR #1531, opened 2022-10-13
Add StackExchange support

kevinchalet:
This PR adds StackExchange to the list of supported providers. For reference, the aspnet-contrib provider can be found here: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/dev/src/AspNet.Security.OAuth.StackExchange

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
@@ -58,6 +58,20 @@
     </Environment>
   </Provider>
 
+  <Provider Name="StackExchange" Documentation="https://api.stackexchange.com/docs/authentication">
+    <Environment Issuer="https://api.stackexchange.com/">
+      <Configuration AuthorizationEndpoint="https://stackoverflow.com/oauth"
+                     TokenEndpoint="https://stackoverflow.com/oauth/access_token/json"
+                     UserinfoEndpoint="https://api.stackexchange.com/2.3/me" />
+    </Environment>
+
+    <Setting PropertyName="ApplicationKey" ParameterName="key" Type="String" Required="true"
+             Description="The application key used to communicate with the StackExchange API" />
+
+    <Setting PropertyName="Site" ParameterName="site" Type="String" Required="true" DefaultValue="stackoverflow"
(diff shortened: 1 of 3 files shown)


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -58,6 +58,20 @@
     </Environment>
   </Provider>
 
+  <Provider Name="StackExchange" Documentation="https://api.stackexchange.com/docs/authentication">
+    <Environment Issuer="https://api.stackexchange.com/">
+      <Configuration AuthorizationEndpoint="https://stackoverflow.com/oauth"
+                     TokenEndpoint="https://stackoverflow.com/oauth/access_token/json"
```
kevinchalet, line 64: @martincostello StackExchange supports returning a JSON token response if you use `https://stackoverflow.com/oauth/access_token/json` instead of `https://stackoverflow.com/oauth/access_token`. Should we update the aspnet-contrib provider to use it and avoid having to deal with formurl-encoded responses?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -58,6 +58,20 @@
...
+                     UserinfoEndpoint="https://api.stackexchange.com/2.3/me" />
+    </Environment>
+
+    <Setting PropertyName="ApplicationKey" ParameterName="key" Type="String" Required="true"
```
kevinchalet, line 68: @martincostello this parameter is required when communicating with APIs that take an access token (which is the case for the userinfo endpoint), but for some reasons, we don't throw an exception if it's not set in the aspnet-contrib provider. Should we change that?

```json
{"error_id":405,"error_message":"`key` is required when `access_token` is passed.","error_name":"key_required"}
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
martincostello: Cool - I'll take a look at that and the other comment to see if we can make the change as part of the v7 release.

PR #1552, opened 2022-10-26
Return true from IsEmptyJsonElement for Undefined and Null

trejjam:
Hello there,
thank you for a really nice library. I really like to work with it.

I was playing with it and I had a small issue with `OpenIddictServerDataProtectionFormatter`. For some reason serialized principal for AuthorizationCode flow does not contain `OpenIddictConstants.Claims.Private.Audience`, which is probably fine, because it's populated in token flow. But the absence of the audience claim/property causes the deserialization of payload to fail while trying to call `SetClaims` for `Claims.Private.Audience` in OpenIddictServerDataProtectionFormatter.

This PR should solve it since `SetClaims` check the existence of a value with `IsEmptyJsonElement`.

### DIFF
--- a/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
@@ -3097,6 +3097,9 @@ public static class OpenIddictExtensions
     {
         switch (element.ValueKind)
         {
+            case JsonValueKind.Undefined or JsonValueKind.Null:
+                return true;
+
             case JsonValueKind.String:
                 return string.IsNullOrEmpty(element.GetString());
 
--- a/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs
+++ b/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs
@@ -3016,7 +3016,7 @@ public class OpenIddictExtensionsTests
         var identity = new ClaimsIdentity();
 
         // Act
-        identity.SetClaims("type", default(JsonElement));
+        identity.SetClaim("type", default(JsonElement));
 
         // Assert
         Assert.Null(identity.GetClaim("type"));
@@ -3043,7 +3043,7 @@ public class OpenIddictExtensionsTests
         var identity = new ClaimsIdentity();
 
         // Act
-        identity.SetClaims("type", JsonSerializer.Deserialize<JsonElement>("null"));
+        identity.SetClaim("type", JsonSerializer.Deserialize<JsonElement>("null"));
 
         // Assert
         Assert.Null(identity.GetClaim("type"));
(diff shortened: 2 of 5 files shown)


src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
```diff
@@ -3097,6 +3097,12 @@ private static bool IsEmptyJsonElement(JsonElement element)
     {
         switch (element.ValueKind)
         {
+            case JsonValueKind.Undefined:
+                return true;
+
+            case JsonValueKind.Null:
+                return true;
```
kevinchalet: For bonus points, we can merge the two cases 😄 

```suggestion
            case JsonValueKind.Null or JsonValueKind.Undefined:
                return true;
```

test/OpenIddict.Server.DataProtection.Tests/OpenIddict.Server.DataProtection.Tests.csproj
```diff
@@ -0,0 +1,26 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\OpenIddict.Server.DataProtection\OpenIddict.Server.DataProtection.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
+    <Reference Include="System.Net.Http" />
+  </ItemGroup>
+
```
kevinchalet:
```suggestion
```

kevinchalet (commented):
Hey,

> Hello there,
thank you for a really nice library. I really like to work with it.

Nice catch. Thanks for your kind words and your PR! ❤️ 

There are tests for these extensions (but not for null/undefined `JsonElement` instances it seems 😅) in `OpenIddict.Abstractions.Tests`. Could you add a test or two there to ensure we don't regress in the future?

Thanks again!

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs
```diff
@@ -3010,6 +3009,60 @@ public void ClaimsPrincipal_SetClaimWithJsonElement_RemovesEmptyClaim()
...
     }
 
+    [Fact]
+    public void ClaimsIdentity_SetClaimWithJsonElement_Undefined()
+    {
+        // Arrange
+        var identity = new ClaimsIdentity();
+
+        // Act
+        identity.SetClaims("type", default(JsonElement));
```
kevinchalet: Nit: the method used here doesn't match the name of the test: `SetClaims` vs `SetClaim`.
AFAICT, the other test methods are affected too.

kevinchalet: Actually, we should probably have similar tests for both `SetClaim` and `SetClaims` so both are covered 😃

trejjam: Make sense

trejjam:
Thanks for reviewing. It should be complete now :)

kevinchalet:
Merged. Congrats for your first contribution to OpenIddict, Jan! 🎉

PR #1559, opened 2022-11-04
Introduce IOpenIddictApplicationManager

trejjam:
Hi, I understand why there is a non-generic `IOpenIddictApplicationManager`. At the same time, it's pretty comfortable for a final library consumer to have a generic version also.

I prepared PoC on ApplicationManager, if you like the idea I can prepare a similar one for other managers.

### DIFF
--- a/src/OpenIddict.Abstractions/Managers/IOpenIddictApplicationManager.cs
+++ b/src/OpenIddict.Abstractions/Managers/IOpenIddictApplicationManager.cs
@@ -14,6 +14,495 @@ namespace OpenIddict.Abstractions;
 
 /// <summary>
 /// Provides methods allowing to manage the applications stored in the store.
+/// </summary>
+public interface IOpenIddictApplicationManager<TApplication> where TApplication : class
+{
+    /// <summary>
+    /// Determines the number of applications that exist in the database.
+    /// </summary>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> that can be used to abort the operation.</param>
+    /// <returns>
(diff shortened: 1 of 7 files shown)


kevinchalet:
Hey. Thanks for your PR!

What's the scenario you have in mind? So far, not having a generic interface was deliberate and meant to encourage folks who need to override certain methods of the manager to derive their implementations from the concrete `OpenIddictApplicationManager<T>` instead of trying to implement an interface (it's the same exact approach used by ASP.NET Core Identity for the user and role managers).

It's worth noting that if you prefer working with the strongly-typed manager and don't mind depending on a specific entity, it's already 100% supported: you can simply resolve `OpenIddictApplicationManager<T>` from the DI container.

trejjam:
I do not mind about extending `OpenIddictApplicationManager<T>` rather than implementing my own.

I did not realize that it was possible to get implementation from DI until I prepared this PR. It's not possible in general. I tend to depend on interfaces in my code and exclusively when I am using an external library. For me is comfortable to depend on `IService<MyEntity>` rather than `ThirdPartyImplementation<MyEntity>`. Also, it's easy/clear when you need to mock services using an interface rather than implementation (I know that managers have everything virtual, therefore it's possible).

kevinchalet:
Your reasoning makes sense. That said, the demand for that is currently very low (AFAIK, you're the only one asking for that) so I'll close it for now. Should we get more demand for a generic interface, it's of course something we'll be able to reconsider in the future 😃 

Thanks again.

trejjam:
Sure, no pressure :)

PR #1563, opened 2022-11-07
Fix global.json to reference Arcade 7.0.0-beta.22513.4

kevinchalet:


kevinchalet:
Note: Arcade 7 references a recent version of xUnit that introduced nullability annotations changes and required tweaking a few tests.

PR #1624, opened 2022-12-13
PopulateClaims complex JSON response throws exceptions

danbluhmhansen:
Hello! We are using an OpenID provider that returns JSON arrays and objects within parameters returned from the `UserInfo` endpoint. The returned parameters are populated [here](https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Client/OpenIddictClientHandlers.Userinfo.cs#L179) which uses [this](https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs#L1414) `JsonElement.GetString()` call to convert the JSON to strings, which fails since the enumerated `JsonElement`s are not primitives.

Below is a suggestion that should not fail when calling `JsonElement.GetString()`.

Here is the exception we experience:
```
System.InvalidOperationException: The requested operation requires an element of type 'String', but the target element has type 'Object'.
   at System.Text.Json.ThrowHelper.ThrowJsonElementWrongTypeException(JsonTokenType expectedType, JsonTokenType actualType)
   at System.Text.Json.JsonDocument.GetString(Int32 index, JsonTokenType expectedType)
   at OpenIddict.Abstractions.OpenIddictExtensions.AddClaims(ClaimsIdentity identity, String type, JsonElement value, String issuer)
   at OpenIddict.Client.OpenIddictClientHandlers.Userinfo.PopulateClaims.HandleAsync(HandleUserinfoResponseContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientService.<>c__DisplayClass14_0.<<SendUserinfoRequestAsync>g__HandleUserinfoResponseAsync|3>d.MoveNext()
--- End of stack trace from previous location ---
   at OpenIddict.Client.OpenIddictClientService.SendUserinfoRequestAsync(OpenIddictClientRegistration registration, OpenIddictRequest request, Uri uri, CancellationToken cancellationToken)
   at OpenIddict.Client.OpenIddictClientService.SendUserinfoRequestAsync(OpenIddictClientRegistration registration, OpenIddictRequest request, Uri uri, CancellationToken cancellationToken)
   at OpenIddict.Client.OpenIddictClientHandlers.SendUserinfoRequest.HandleAsync(ProcessAuthenticationContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientHandlers.Authentication.ValidateTokens.HandleAsync(ValidateRedirectionRequestContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientHandlers.Authentication.ValidateRedirectionRequest.HandleAsync(ProcessRequestContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.OpenIddictClientDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Client.AspNetCore.OpenIddictClientAspNetCoreHandler.HandleRequestAsync()
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Foo.Server.Extensions.TenantResolverExtensions.<>c.<<UseTenantResolveFromLoginCallbackPath>b__2_2>d.MoveNext() in C:\Projects\FooProject\Source\Presentation\Foo.Server\Extensions\TenantResolverExtensions.cs:line 71
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddlewareImpl.Invoke(HttpContext context)
```

### DIFF
--- a/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
@@ -1409,10 +1409,9 @@ public static class OpenIddictExtensions
 
         var set = new HashSet<string>(StringComparer.Ordinal);
 
-        foreach (var element in value.EnumerateArray().Where(e =>
-            e.ValueKind is JsonValueKind.False or JsonValueKind.Number or JsonValueKind.String or JsonValueKind.True))
+        foreach (var element in value.EnumerateArray())
         {
-            var item = element.GetString()!;
+            var item = element.ToString()!;
             if (set.Add(item))
             {
                 identity.AddClaim(new Claim(
--- a/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
+++ b/src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
@@ -1409,7 +1409,8 @@ public static class OpenIddictExtensions
 
         var set = new HashSet<string>(StringComparer.Ordinal);
 
-        foreach (var element in value.EnumerateArray())
+        foreach (var element in value.EnumerateArray().Where(e =>
+            e.ValueKind is JsonValueKind.False or JsonValueKind.Number or JsonValueKind.String or JsonValueKind.True))
         {
             var item = element.GetString()!;
             if (set.Add(item))
--- a/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs
+++ b/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs
@@ -2312,7 +2312,10 @@ public class OpenIddictExtensionsTests
         var identity = new ClaimsIdentity();
 
         // Act
-        identity.AddClaims("type", JsonSerializer.Deserialize<JsonElement>(@"[""Fabrikam"",""Contoso""]"), "issuer");
+        identity.AddClaims(
+            "type",
+            JsonSerializer.Deserialize<JsonElement>(@"[""Fabrikam"", ""Contoso"", { ""Foo"": ""Bar"" }]"),
+            "issuer");
 
         // Assert
         var claims = identity.FindAll("type").ToArray();
@@ -2322,6 +2325,9 @@ public class OpenIddictExtensionsTests
         Assert.Equal("Contoso", claims[1].Value);
         Assert.Equal(ClaimValueTypes.String, claims[1].ValueType);
         Assert.Equal("issuer", claims[1].Issuer);
+        Assert.Equal("{ \"Foo\": \"Bar\" }", claims[2].Value);
+        Assert.Equal("JSON", claims[2].ValueType);
+        Assert.Equal("issuer", claims[2].Issuer);
     }
 
     [Fact]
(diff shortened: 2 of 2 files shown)


src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
```diff
@@ -1409,7 +1409,8 @@ public static ClaimsIdentity AddClaims(this ClaimsIdentity identity, string type
 
         var set = new HashSet<string>(StringComparer.Ordinal);
 
-        foreach (var element in value.EnumerateArray())
+        foreach (var element in value.EnumerateArray().Where(e =>
```
kevinchalet: Instead of doing that, we should probably just fix the line 1414/1415 to use `element.ToString()` instead of `element.GetString()` so that `System.Text.Json` automatically stringifies non-JSON-string elements. This way, even complex objects will be added to the resulting `ClaimsPrincipal` instead of being ignored 😃

kevinchalet (commented):
Thanks for your PR! I added a comment suggesting a different approach that should preserve complex claims instead of silently ignoring them.

(I checked the other places where `GetString()` and it seems it's the only problematic usage: all other calls only work on JSON strings).

kevinchalet:
> Hello! We are using an OpenID provider that returns JSON arrays and objects within parameters returned from the `UserInfo` endpoint.

Hey! I'm curious, may I ask what provider you're using?

kevinchalet:
Note: this also affects the validation handler when using introspection as it uses the same code path.

danbluhmhansen:
> > Hello! We are using an OpenID provider that returns JSON arrays and objects within parameters returned from the `UserInfo` endpoint.
> 
> Hey! I'm curious, may I ask what provider you're using?

It's a Danish company https://zenegy.com/

src/OpenIddict.Abstractions/Primitives/OpenIddictExtensions.cs
danbluhmhansen: Done!

kevinchalet:
There are tests for this extension. Can you please update them to use a nested non-string element so that things don't regress in the future? 😃 

https://github.com/openiddict/openiddict-core/blob/6e0c97262dcee693932e35d4a70decdf6163289b/test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs#L2309-L2345

kevinchalet:
> It's a Danish company https://zenegy.com/

Nice! Is it a popular service? If so, it may make sense to support it in `OpenIddict.Client.WebIntegration` 😃 
Would you be interested in adding it?

danbluhmhansen:
> Nice! Is it a popular service? If so, it may make sense to support it in `OpenIddict.Client.WebIntegration` 😃 Would you be interested in adding it?

I'm not sure it's that popular. I could contact their developer team and suggest they added that.

test/OpenIddict.Abstractions.Tests/Primitives/OpenIddictExtensionsTests.cs
```diff
@@ -2322,6 +2325,9 @@ public void ClaimsIdentity_AddClaimsWithJsonElement_AddsExpectedClaims()
         Assert.Equal("Contoso", claims[1].Value);
         Assert.Equal(ClaimValueTypes.String, claims[1].ValueType);
         Assert.Equal("issuer", claims[1].Issuer);
+        Assert.Equal("{ \"Foo\": \"Bar\" }", claims[2].Value);
+        Assert.Equal("JSON", claims[2].ValueType);
```
danbluhmhansen, line 2329: I am unsure how to handle the `ValueType`. There is no `JSON` type in `ClaimValueTypes`, what do you think?

kevinchalet: These values come from `JsonClaimValueTypes`, a type defined in `Microsoft.IdentityModel.JsonWebTokens` (which is not a package referenced by `OpenIddict.Abstractions`).

Let's just keep the hardcoded string for now 😃

kevinchalet:
> I could contact their developer team and suggest they added that.

Please don't hesitate. You can point them to https://documentation.openiddict.com/guides/contributing-a-new-web-provider.html (it seems they support OpenID Connect discovery so it should be fairly easy to add: https://auth.zenegy.com/.well-known/openid-configuration) 😃

kevinchalet:
Merged, thanks! And congrats for your first contribution to OpenIddict 🎉

PR #1661, opened 2023-01-28
Introduce a faster way to revoke all the tokens associated with an authorization and use bulk operations when available

### LINKED ISSUE #1217: Removing historic token entries generates exception when using an executing strategy.
### Describe the bug
Removing historic token entries generates exception when using an executing strategy using SQL Server.

### To reproduce
I'm using the Quartz plugin as well as EF Core on a SQL Server with the EnableRetryOnFailure strategy set on the DbContext (options.UseSqlServer(storageSettings.ConnectionString, sqlServerOptions => sqlServerOptions.EnableRetryOnFailure());)

### Exceptions (if any)
On job execution: One or more errors occurred. (The configured execution strategy 'SqlServerRetryingExecutionStrategy' does not support user-initiated transactions. Use the execution strategy returned by 'DbContext.Database.CreateExecutionStrategy()' to execute all the operations in the transaction as a retriable unit.)

### My analysis
Enabling a executing strategy prevents you from initiating your own transactions (https://docs.microsoft.com/en-us/ef/core/miscellaneous/connection-resiliency#execution-strategies-and-transactions).

In OpenIddictEntityFrameworkCoreTokenStore.PruneAsync transactions are initiated to remove batches of tokens. If I remove EnableRetryOnFailure, pruning seems to work just fine.

I'm looking for a solution for this. I would like to keep EnableRetryOnFailure.

For me, using a explicit transaction in OpenIddictEntityFrameworkCoreTokenStore.PruneAsync seems not needed. I expect historic tokens are never modified and are only deleted by PruneAsync. So my solution would be removing the explicit transaction, but I probably overlook something.

Same for OpenIddictEntityFrameworkCoreAuthorizationStore.

Thanks for developing this package.

### LINKED ISSUE #1660: Leverage EF Core 7.0's set-based/bulk delete support in `PruneAsync()`

EF Core 7.0 now supports bulk delete and update operations, which could be used to significantly improve the performance of the `PruneAsync()` methods:

https://github.com/openiddict/openiddict-core/blob/5e54e9be73261eb8741a404f6b6ee1f77a4ff226/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs#L662-L750

https://github.com/openiddict/openiddict-core/blob/5e54e9be73261eb8741a404f6b6ee1f77a4ff226/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreTokenStore.cs#L626-L711

### Additional context

_No response_

### LINKED ISSUE #1856: Introduce a more efficient way to revoke all the tokens associated with an authorization

When a token replay is detected, all the tokens attached to the same authorization are automatically revoked. For that, the tokens are individually retrieved from the DB and marked as revoked. While this allows revoking the tokens individually without being affected by concurrent operations, this is a slow process.

In the next major version, we should consider adding new store methods to make that process much more efficient.

### Additional context

https://github.com/openiddict/openiddict-core/issues/1819#issuecomment-1618396124

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/1217.
Fixes https://github.com/openiddict/openiddict-core/issues/1660.
Fixes https://github.com/openiddict/openiddict-core/issues/1856.

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs
@@ -150,71 +150,93 @@ public class OpenIddictEntityFrameworkCoreAuthorizationStore<TAuthorization, TAp
             throw new ArgumentNullException(nameof(authorization));
         }
 
-        async ValueTask<IDbContextTransaction?> CreateTransactionAsync()
+#if SUPPORTS_BULK_DBSET_OPERATIONS
+        if (!Options.CurrentValue.DisableBulkOperations)
         {
-            // Note: transactions that specify an explicit isolation level are only supported by
-            // relational providers and trying to use them with a different provider results in
-            // an invalid operation exception being thrown at runtime. To prevent that, a manual
-            // check is made to ensure the underlying transaction manager is relational.
-            var manager = Context.Database.GetService<IDbContextTransactionManager>();
-            if (manager is IRelationalTransactionManager)
+            var strategy = Context.Database.CreateExecutionStrategy();
+            await strategy.ExecuteAsync(async () =>
             {
+                // To prevent an SQL exception from being thrown if a new associated entity is
+                // created after the existing entries have been listed, the following logic is
+                // executed in a serializable transaction, that will lock the affected tables.
+                using var transaction = await Context.CreateTransactionAsync(IsolationLevel.Serializable, cancellationToken);
+
+                // Remove all the tokens associated with the authorization.
+                await (from token in Tokens.AsTracking()
+                       where token.Authorization!.Id!.Equals(authorization.Id)
+                       select token).ExecuteDeleteAsync(cancellationToken);
+
+                // Note: calling DbContext.SaveChangesAsync() is not necessary
+                // with bulk delete operations as they are executed immediately.
+
+                Context.Remove(authorization);
+
                 try
                 {
-                    return await Context.Database.BeginTransactionAsync(IsolationLevel.Serializable, cancellationToken);
+                    await Context.SaveChangesAsync(cancellationToken);
+                    transaction?.Commit();
                 }
 
-                catch (Exception exception) when (!OpenIddictHelpers.IsFatal(exception))
+                catch (DbUpdateConcurrencyException exception)
                 {
-                    return null;
+                    // Reset the state of the entity to prevents future calls to SaveChangesAsync() from failing.
+                    Context.Entry(authorization).State = EntityState.Unchanged;
+
+                    throw new ConcurrencyException(SR.GetResourceString(SR.ID0241), exception);
                 }
-            }
-
-            return null;
+            });
         }
 
-        // Note: due to a bug in Entity Framework Core's query visitor, the tokens can't be
-        // filtered using token.Application.Id.Equals(key). To work around this issue,
-        // this local method uses an explicit join before applying the equality check.
-        // See https://github.com/openiddict/openiddict-core/issues/499 for more information.
-
-        Task<List<TToken>> ListTokensAsync()
-            => (from token in Tokens.AsTracking()
-                join element in Authorizations.AsTracking() on token.Authorization!.Id equals element.Id
-                where element.Id!.Equals(authorization.Id)
-                select token).ToListAsync(cancellationToken);
-
-        // To prevent an SQL exception from being thrown if a new associated entity is
-        // created after the existing entries have been listed, the following logic is
-        // executed in a serializable transaction, that will lock the affected tables.
-        using var transaction = await CreateTransactionAsync();
-
-        // Remove all the tokens associated with the authorization.
-        var tokens = await ListTokensAsync();
-        foreach (var token in tokens)
+        else
+#endif
         {
-            Context.Remove(token);
-        }
+            // Note: due to a bug in Entity Framework Core's query visitor, the tokens can't be
+            // filtered using token.Application.Id.Equals(key). To work around this issue,
+            // this local method uses an explicit join before applying the equality check.
+            // See https://github.com/openiddict/openiddict-core/issues/499 for more information.
 
-        Context.Remove(authorization);
+            Task<List<TToken>> ListTokensAsync()
+                => (from token in Tokens.AsTracking()
+                    join element in Authorizations.AsTracking() on token.Authorization!.Id equals element.Id
+                    where element.Id!.Equals(authorization.Id)
+                    select token).ToListAsync(cancellationToken);
 
-        try
-        {
-            await Context.SaveChangesAsync(cancellationToken);
-            transaction?.Commit();
-        }
-
-        catch (DbUpdateConcurrencyException exception)
-        {
-            // Reset the state of the entity to prevents future calls to SaveChangesAsync() from failing.
-            Context.Entry(authorization).State = EntityState.Unchanged;
-
-            foreach (var token in tokens)
+            var strategy = Context.Database.CreateExecutionStrategy();
+            await strategy.ExecuteAsync(async () =>
             {
-                Context.Entry(token).State = EntityState.Unchanged;
-            }
+                // To prevent an SQL exception from being thrown if a new associated entity is
+                // created after the existing entries have been listed, the following logic is
+                // executed in a serializable transaction, that will lock the affected tables.
+                using var transaction = await Context.CreateTransactionAsync(IsolationLevel.Serializable, cancellationToken);
 
-            throw new ConcurrencyException(SR.GetResourceString(SR.ID0241), exception);
+                // Remove all the tokens associated with the authorization.
+                var tokens = await ListTokensAsync();
+                foreach (var token in tokens)
+                {
+                    Context.Remove(token);
+                }
+
+                Context.Remove(authorization);
+
+                try
+                {
+                    await Context.SaveChangesAsync(cancellationToken);
+                    transaction?.Commit();
+                }
+
+                catch (DbUpdateConcurrencyException exception)
+                {
+                    // Reset the state of the entity to prevents future calls to SaveChangesAsync() from failing.
+                    Context.Entry(authorization).State = EntityState.Unchanged;
+
+                    foreach (var token in tokens)
+                    {
+                        Context.Entry(token).State = EntityState.Unchanged;
+                    }
+
+                    throw new ConcurrencyException(SR.GetResourceString(SR.ID0241), exception);
+                }
+            });
         }
     }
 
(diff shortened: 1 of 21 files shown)


src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs
```diff
@@ -661,38 +638,44 @@ public virtual IAsyncEnumerable<TAuthorization> ListAsync(int? count, int? offse
     /// <inheritdoc/>
     public virtual async ValueTask PruneAsync(DateTimeOffset threshold, CancellationToken cancellationToken)
     {
-        // Note: Entity Framework Core doesn't support set-based deletes, which prevents removing
-        // entities in a single command without having to retrieve and materialize them first.
+#if SUPPORTS_BULK_DBSET_OPERATIONS
+        // To prevent an SQL exception from being thrown if a new associated token is
+        // created after the existing tokens have been removed, the following logic is
+        // executed in a serializable transaction, that will lock the affected tables.
```
Suchiman: Ah cool, you've got around to do it before me 😁 
I assume you mean that deletions of authorizations would fail if a new token entry to an authorization were added between deleting Tokens and Authorizations? Does `!authorization.Tokens.Any()` not sufficiently guard against that?

Bumping isolation all the way up to Serializable might guard against correctness issues such as trying to delete parents who've just acquired new childrens but at the cost of drastically increased chance of running into deadlocks and inducing massive blocking on those tables.

Especially since the two deletes have no Take on them, which means that once i would turn that task back on in my system (haven't turned the task back on since #1508), it would attempt to delete millions of rows in a single serializable transaction which... is bad xD I suppose under normal operation the task runs frequently enough to not delete too many rows at once but it's something to think about.

matthieume: Could we consider a .Take(1000).ExecuteDeleteAsync() ?

kevinchalet: > Bumping isolation all the way up to Serializable might guard against correctness issues such as trying to delete parents who've just acquired new childrens but at the cost of drastically increased chance of running into deadlocks and inducing massive blocking on those tables.

I agree it's not a perfect solution so I explored other options. One of them - implemented in the latest commit - uses `DeleteBehavior.SetNull` to let the database automatically set FKs to `null` without having to remove children entities on our side, which simplifies things a lot.

Of course, this leaves orphaned authorizations and tokens when deleting an application or orphaned tokens when deleting an authorization, but I don't think it's a problem in this case (tho' it's certainly an observable behavior change).

Now, I'm not sure if `DeleteBehavior.SetNull` is compatible with all DBs: unlike `DeleteBehavior.Cascade` that doesn't work in this case with SQL Server (as there are cyclic references), `DeleteBehavior.SetNull` seems to work with at least both SQL Server and SQLite.

> Could we consider a .Take(1000).ExecuteDeleteAsync() ?

Certainly. I opted for 50K in my latest commit.

Let me know what you think 😄

kevinchalet: > One of them - implemented in the latest commit - uses `DeleteBehavior.SetNull` to let the database automatically set FKs to `null` without having to remove children entities on our side, which simplifies things a lot.

Another option I considered was to remove the constraints on foreign keys (after all, OpenIddict itself doesn't care much if FK columns point to inexistant/invalid entries), but it's not something EF Core seems to allow.

kevinchalet: @Suchiman @matthieume don't hesitate to share your thoughts 😃

Suchiman: Sorry, last week's notifications drowned in my postbox.

> unlike DeleteBehavior.Cascade that doesn't work in this case with SQL Server (as there are cyclic references)

If you set all FKs to Cascade, you'll run into multiple cascade path's but there are no cycles as far as i can see.
There are two cascade paths:
`Application -> Authorization -> Token`
`Application -> Token`
The thing that SQL Server is unhappy about is that deleting `Application` would try to delete `Token` twice in different paths. However we can see that the second path is unnecessary since the first path would take care of it.
So if you configure `Authorization.Application` and `Token.Authorization` to be `Cascade` and `Token.Application` to `NoAction`, it should work.

> Another option I considered was to remove the constraints on foreign keys (after all, OpenIddict itself doesn't care much if FK columns point to inexistant/invalid entries), but it's not something EF Core seems to allow.

It is possible, if you delete the ForeignKey constraint from the generated migration manually, EFC does not seem to complain about the FK values that do not exist in the database. However that is a very hard to sell user experience.

kevinchalet: > If you set all FKs to Cascade, you'll run into multiple cascade path's but there are no cycles as far as i can see.
> There are two cascade paths:
> `Application -> Authorization -> Token`
> `Application -> Token`
> The thing that SQL Server is unhappy about is that deleting `Application` would try to delete `Token` twice in different paths. However we can see that the second path is unnecessary since the first path would take care of it.
> So if you configure `Authorization.Application` and `Token.Authorization` to be `Cascade` and `Token.Application` to `NoAction`, it should work.

Interestingly, it's an approach I considered but it doesn't work when no intermediate `Authorization` entity is created... which is typically the case when the developer doesn't explicitly create a permanent authorization using the `IOpenIddictAuthorizationManager.CreateAsync()` API and prevents OpenIddict from creating ad-hoc authorizations using `options.DisableAuthorizationStorage()`. If you use `options.DisableAuthorizationStorage()`, the first cascade path doesn't kick in and you get an error since the tokens are not removed/detached:

```
SqlException : The DELETE statement conflicted with the REFERENCE constraint "FK_OpenIddictTokens_OpenIddictApplications_ApplicationId". The conflict occurred in database "openiddict-sandbox-aspnetcore-server", table "dbo.OpenIddictTokens", column 'ApplicationId'.
```

If we wanted to go this route, we'd have to remove `options.DisableAuthorizationStorage()` and always force the creation of an authorization in the DB. It's definitely going to be a major breaking change and it removes an important scenario, so not sure it's our best option, sadly.

> It is possible, if you delete the ForeignKey constraint from the generated migration manually, EFC does not seem to complain about the FK values that do not exist in the database. However that is a very hard to sell user experience.

Ah yeah, when I said it wasn't allowed, I meant via an entity configuration: as you say, forcing users to manually remove constraints from the migrations would be a terrible experience 😄

Suchiman: ~I'm not at a computer right now so I can't test but is it possible to set `Token.Application` to SetNull while keeping the rest Cascade?~ Thinking about it though, do we have the same problems of people trying to delete applications while new tokens are being added? Can we workaround it by setting the application to disabled or trashing its client ID to prevent new tokens being created while running a retry loop that does { delete all application tokens; delete application }? That way we can keep it Cascade for the majority of cases and handle it special in application deletes which should be rare for most users, unlike the cleanup tasks.

edmacdonald: > I opted for 50K in my latest commit.

Can this be made configurable?  50k will lock the table and that can't happen in our deployment.

src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreTokenStore.cs
```diff
@@ -626,55 +626,50 @@ public virtual IAsyncEnumerable<TToken> ListAsync(int? count, int? offset, Cance
...
+#if SUPPORTS_BULK_DBSET_OPERATIONS
+            try
+            {
+                await (from token in Tokens
+                       where token.CreationDate < date
+                       where (token.Status != Statuses.Inactive && token.Status != Statuses.Valid) ||
+                             (token.Authorization != null && token.Authorization.Status != Statuses.Valid) ||
+                              token.ExpirationDate < DateTime.UtcNow
+                       orderby token.Id
+                       select token).Take(50_000).ExecuteDeleteAsync(cancellationToken);
```
edmacdonald: This will result in a full table scan.  The token Id (clustered index) is a Guid and a new one is just as likely to end up at the beginning of the table as the end.  This will halt a large system like ours (200 million tokens).  The 50k batch size will result in a table lock which will be a non-starter for some (including me) so can this be made configurable?

edmacdonald (changes requested):

src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreTokenStore.cs
kevinchalet: > The 50k batch size will result in a table lock which will be a non-starter for some (including me) so can this be made configurable?

If we reduce the batch size, we need to make more passes to eliminate the same number of entries. Or we need to accept there may be many remaining entries after a single call to `PruneAsync()`.

I'm open to all suggestions: what would be a more appropriate value for you?

>  This will halt a large system like ours (200 million tokens).

For such demanding deployments, I'd recommend going DB-first and writing your own stores - e.g based on Dapper or ADO.NET directly - instead of using the EF Core stores and the default generated schema.

edmacdonald: Batches of 1_000 are generally comfortable, at 5_000 (in my experience) you'll start to experience lock escalation.  For sites that can't tolerate the downtime, the alternative is turning off pruning so leaving behind some tokens should be fine.

I rolled my own clean up script, but added an index on, and only looked at `ExpirationDate`.  You might want to consider that here as well.  With `CreationDate` and `Status` in the predicate, SqlServer will just scan the CX, and the oldest, most expired token might be the very last one.  That will create seizures in even a modest size system.

I'll look closer at stores in the new year once we're through our busy season, but our biggest challenge has been the number of tokens stored (3 every time we use a refresh token to get an access token) and the storage size, rather that EF performance.  I'm getting off an a tangent here, so I'll put together observations, suggestions and maybe even a PR or two in the new year.  Thanks for this project.

matthieume: As we have a datetime threshold, it could be great to have a limit to the batch size.

kevinchalet: > Batches of 1_000 are generally comfortable, at 5_000 (in my experience) you'll start to experience lock escalation.

> As we have a datetime threshold, it could be great to have a limit to the batch size.


Sold. I updated my PR to use 1K batches, which is the same value used when bulk operations are not supported or enabled.

> You might want to consider that here as well.

The problem with this approach is that you might remove tokens that are just expired, so the threshold is not respected and you lose track of the tokens history. I wish there was a perfect option but it doesn't seem to exist 😕

src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs
kevinchalet: I like the idea: I reverted the cascade changes I previously tested for now, but If I get reports indicating conflicts are a problem, I may end up implement the retry logic you suggest 😃

kevinchalet:
PR updated to also fix https://github.com/openiddict/openiddict-core/issues/1217 and https://github.com/openiddict/openiddict-core/issues/1856, that are closely related.

I can't say I'm fully happy with the result, but it seems to be a reasonable compromise 😅 

Hope I addressed everyone's concerns.

kevinchalet:
If there's no remaining feedback/remarks to address, I'll merge this PR and release a new 5.0 preview, so everyone has a chance to test it before next month's release 😃

PR #1662, opened 2023-01-31
Add native support for Windows Runtime

### LINKED ISSUE #1653: Make the Windows client integration compatible with sandboxed applications

The new Windows integration currently relies on `Process.Start()` to launch the system browser:

https://github.com/openiddict/openiddict-core/blob/5f64d67c92f9f031296b2f672ae606fec32a8ed2/src/OpenIddict.Client.Windows/OpenIddictClientWindowsHandlers.Authentication.cs#L57-L79

While this works in regular Win32 applications and packaged applications that are granted the `runFullTrust` capability (which is added by default for WinUI 3 apps), this doesn't work in partial trust applications and sandboxed applications that run in an `AppContainer`.

To fix that, we'll need to detect whether we're running on an environment that supports `Launcher.LaunchUriAsync()` and use it instead of `Process.Start()`.

### Additional context

_No response_

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/1653.

Note: with this change, the OpenIddict Windows client integration now works natively in UWP applications, but there are important limitations to keep in mind:
  - Windows 8's Metro-style/"universal applications" are not supported.
  - Windows 10 1809 is the minimum supported version: earlier versions of Windows 10 are not supported for UWP applications (full-trust desktop applications using the OpenIddict Windows client integration are supported on any 7/8/10 version).
  - The callback handling mechanism (that uses an `IHostedService` invoked via the entry point) will only work OOTB with UWP applications that opt in multi-instancing (`desktop4:SupportsMultipleInstances="true"`/`iot2:SupportsMultipleInstances="true"`).
  - AFAIK, there's no easy-to-use UWP/WinUI2 integration for the .NET generic host, so it's not as well integrated as with WinForms or WPF.
  - EF Core 3.x doesn't fully work on UWP so you need to use different stores (MongoDB worked but it's not clear whether it would pass Store validation 😅). More recent/less bugged versions of EF Core cannot be used as they no longer target .NET Standard 2.0...

I tested with the latest Windows App Cert Kit (WACK) version and was able to confirm no banned API is used.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -86,6 +86,7 @@
     <DefineConstants>$(DefineConstants);SUPPORTS_HTTP_CLIENT_DEFAULT_REQUEST_VERSION_POLICY</DefineConstants>
     <DefineConstants>$(DefineConstants);SUPPORTS_MULTIPLE_VALUES_IN_QUERYHELPERS</DefineConstants>
     <DefineConstants>$(DefineConstants);SUPPORTS_ONE_SHOT_HASHING_METHODS</DefineConstants>
+    <DefineConstants>$(DefineConstants);SUPPORTS_OPERATING_SYSTEM_VERSIONS_COMPARISON</DefineConstants>
     <DefineConstants>$(DefineConstants);SUPPORTS_PEM_ENCODED_KEY_IMPORT</DefineConstants>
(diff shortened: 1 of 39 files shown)


kevinchalet:
Note: Windows App SDK is deliberately not used, as the deployment story for non-packaged applications is still quite painful.

PR #1700, opened 2023-02-28
Remove dependency on EnablePreviewFeatures

virzak:
With that property it is required in the consuming dll.


kevinchalet:
Thanks for your PR.

Having to set `EnablePreviewFeatures` to `true` in your project is deliberate, as mentioned in https://kevinchalet.com/2023/02/27/introducing-system-integration-support-for-the-openiddict-client/.

It will eventually go away in a future version, but for now, it's still experimental.

PR #1731, opened 2023-03-10
Replace SmartFormat by C# string interpolation and add Battle.net to the list of supported providers

kevinchalet:
~~Note: `<PackageScribanIncludeSource>true</PackageScribanIncludeSource>` is deliberately used to "internalize" Scriban and avoid potential build/runtime issues in applications that would indirectly or directly reference both the signed and non-signed Scriban packages.~~


kevinchalet:
Change of plan: embedding Scriban in the `OpenIddict.Client.WebIntegration` was causing issues with PDB validation. It is now replaced by native C# string interpolation.

PR #1745, opened 2023-04-06
Do not resolve configuration if access token was not provided

2d1a0ec3:
## Purpose of the Pull-Request

This PR fixes a regression introduced in 4.* where the discovery endpoint is always called first, reglardless if access token was provided in a call or not. In the previous version 3.* the OpenIdConnect configuration was resolved only after verifying that the token is not empty. 

## Why this is an issue

If the authorization server is offline, then it is not possible to reach any of the API endpoints, including those that are publicly available. The validation middleware will try to read a OpenIdConnect configuration from the discovery endpoint and fail every call, even if access token was not provided and the API endpoint is public.

For example, let's consider the following API:

```csharp
builder.Services.AddOpenIddict()
    .AddValidation(options =>
    {
        //...
    });
//..
var app = builder.Build();
app.MapGet("/private", (ClaimsPrincipal user) => $"{user.Identity!.Name} is allowed to access the API");
app.MapGet("/public", () => "This is a public endpoint. Authorization is not required.");
```

If the auth. server is offline then trying to call the `/public` endpoint will result in an internal (500) error response.

This shouldn't happen and calls to public APIs should be possible even if the auth. server is offline.

## Discussion

To solve the problem I reordered some validation handlers. I chose `Order` values that made sense to me, but I would be happy to use different values if there is a specific convention that you chose or if it is in suboptimal in any other way. Normally, I would create an issue with the proposal first, but I am neither a sponsor or a contributor yet.

Please note that this PR only fixes handling of calls made with no access tokens. If access token is provided and a public API is called while the auth. server is offline, it will still produce a 500 error. I think this issue existed in v3.* too and probably should be addressed at some point.


kevinchalet:
Thanks for your PR!

> Please note that this PR only fixes handling of calls made with no access tokens. If access token is provided and a public API is called while the auth. server is offline, it will still produce a 500 error. I think this issue existed in v3.* too and probably should be addressed at some point.

It's worth noting that this behavior exists in both the ASP.NET Core OIDC and JWT bearer handlers too: if the configuration cannot be retrieved (for any reason), it will throw an exception that will be surfaced as a 500 error.

That said, I agree the UX is not ideal: it would certainly be a good idea to catch such exceptions and call `context.Reject()` so that the ASP.NET Core authentication service can receive an `AuthenticateResult` indicating that no authentication could be performed due to the missing configuration. At least, it would allow the authorization stack to determine whether the request should be rejected or just considered unauthenticated.

I'll open a ticket to track doing that in the next release.

> This PR fixes a regression introduced in 4.* where the discovery endpoint is always called first, reglardless if access token was provided in a call or not. In the previous version 3.* the OpenIdConnect configuration was resolved only after verifying that the token is not empty.

I wouldn't call that a regression, as calling `AuthenticateAsync("OpenIddict.Validation.AspNetCore")` (which is what happens automatically in .NET 7 if there's no other authentication handler registered) without the server being reachable/configuration being available has never been a supported scenario. As you said, it happened to work because it was delayed in 3.x, but the configuration is now retrieved and exposed earlier in the `ProcessAuthentication(Context)` event.

I'm not opposed to adopting the change you suggested, but only as an optimization (tho' it may potentially break user-defined event handlers that need the server configuration to perform custom actions during the `ProcessAuthentication(Context)` handling): the real fix here is to catch configuration retrieval errors and call `context.Reject()` 😃

2d1a0ec3:
Thanks for a quick reply! I am OK with calling it an optimization of validation order rather than a regression fix :).

In the end, I think that a mix of both solutions could be optimal.

1. Optimizing the the validation order so that existence of a token is verified before resolving configuration
2. Calling `context.Reject()` to catch configuration retrieval errors if token is present, but configuration cannot be resolved

I think the main reason for the importance of the 1. is that the http client used for resolving configuration, by default, will follow a retry policy and will try to reach the configuration endpoint a couple of times before it gives up, which would be sub-optimal, because it would unnecessarily delay the API responses.

Right now, when we call a public API while the auth. server is offline it will return a 500 (after a relatively long delay caused by the retry policy). Implementing the 1. change, suggested in this PR, will deal with calls made with no access tokens present and return a valid reply immediately. Then if you decide to implement the 2., it would handle calls made with access tokens present and auth. server offline. There still would be a delay then, but at least the users would not get a 500.

The ideal solution would be if we could somehow establish that a specific API endpoint is public and skip the entire validation, but I guess this wouldn't be easy?

kevinchalet:
> The ideal solution would be if we could somehow establish that a specific API endpoint is public and skip the entire validation, but I guess this wouldn't be easy?

Sadly, it's not possible: the authentication stack is completely decoupled from the authorization stack (which is invoked after authentication has taken place anyway), so OpenIddict never knows whether authentication will be required or not eventually.

src/OpenIddict.Validation/OpenIddictValidationHandlers.cs
```diff
@@ -17,9 +17,9 @@ public static partial class OpenIddictValidationHandlers
         /*
          * Authentication processing:
          */
-        ResolveServerConfiguration.Descriptor,
         EvaluateValidatedTokens.Descriptor,
         ValidateRequiredTokens.Descriptor,
+        ResolveServerConfiguration.Descriptor,
```
kevinchalet, line 22: Can you please also move the handler code so that it appears at the same place in the file as defined here?

src/OpenIddict.Validation/OpenIddictValidationHandlers.cs
```diff
@@ -79,7 +79,7 @@ public sealed class EvaluateValidatedTokens : IOpenIddictValidationHandler<Proce
         public static OpenIddictValidationHandlerDescriptor Descriptor { get; }
             = OpenIddictValidationHandlerDescriptor.CreateBuilder<ProcessAuthenticationContext>()
                 .UseSingletonHandler<EvaluateValidatedTokens>()
-                .SetOrder(ResolveServerConfiguration.Descriptor.Order + 1_000)
+                .SetOrder(int.MinValue + 80_000)
```
kevinchalet: Let's use `100_000` here.

src/OpenIddict.Validation/OpenIddictValidationHandlers.cs
```diff
@@ -49,7 +49,7 @@ public sealed class ResolveServerConfiguration : IOpenIddictValidationHandler<Pr
         public static OpenIddictValidationHandlerDescriptor Descriptor { get; }
             = OpenIddictValidationHandlerDescriptor.CreateBuilder<ProcessAuthenticationContext>()
                 .UseSingletonHandler<ResolveServerConfiguration>()
-                .SetOrder(int.MinValue + 100_000)
+                .SetOrder(ValidateRequiredTokens.Descriptor.Order + 5_000)
```
kevinchalet:
```suggestion
                .SetOrder(ValidateRequiredTokens.Descriptor.Order + 1_000)
```

src/OpenIddict.Validation/OpenIddictValidationHandlers.cs
```diff
@@ -125,7 +125,7 @@ public sealed class ValidateRequiredTokens : IOpenIddictValidationHandler<Proces
                 .UseSingletonHandler<ValidateRequiredTokens>()
                 // Note: this handler is registered with a high gap to allow handlers
                 // that do token extraction to be executed before this handler runs.
-                .SetOrder(EvaluateValidatedTokens.Descriptor.Order + 50_000)
+                .SetOrder(EvaluateValidatedTokens.Descriptor.Order + 5_000)
```
kevinchalet: `50_000` was an appropriate value, let's keep it 😃

test/OpenIddict.Validation.IntegrationTests/OpenIddictValidationIntegrationTests.cs
```diff
@@ -63,6 +63,34 @@ public async Task ProcessAuthentication_EvalutesCorrectValidatedTokens()
...
+    {
+        // Arrange
+        await using var server = await CreateServerAsync(options =>
+        {
+            options.AddEventHandler<ProcessAuthenticationContext>(builder =>
+            {
+                builder.UseInlineHandler(context =>
+                {
+                    // Assert
+                    Assert.True(context.IsRejected);
```
kevinchalet: Are we sure this inline handler is invoked/this assert is reached? When a handler rejects the request, the next handler in the pipeline is not invoked, so this shouldn't be possible do such a check here.

kevinchalet (commented):
Thanks for your PR!

kevinchalet:
> I think the main reason for the importance of the 1. is that the http client used for resolving configuration, by default, will follow a retry policy and will try to reach the configuration endpoint a couple of times before it gives up, which would be sub-optimal, because it would unnecessarily delay the API responses.

Note: if you think the retry policy results in delays that are too long, I'm not opposed to tweaking it to achieve a better compromise between resilience to transient errors and UX 😃

test/OpenIddict.Validation.IntegrationTests/OpenIddictValidationIntegrationTests.cs
2d1a0ec3: You are right, I refactored the test to verify that the demand is already rejected before resolving server configuration when access token is missing. I hope this one makes more sense and if not, I would be happy to remove the test altogether or follow a more specific suggestion.

kevinchalet: I don't think it will work (just like the test under the one you wrote, `ProcessAuthentication_RejectsDemandWhenAccessTokenIsMissing`, is flaky and has an inline handler that is not executed: since the previous handler marks the request as rejected, it's never invoked).

Let's remove it for now: I'll add a test that creates a mocked `IConfigurationManager<T>` that throws when calling `GetConfigurationAsync()` in a separate PR (unless you want to do it, of course 😃)

kevinchalet: Something like this should do the trick 😃 

```csharp
[Fact]
public async Task ProcessAuthentication_RejectsDemandWithoutResolvingServerConfigurationWhenAccessTokenIsMissing()
{
    // Arrange
    var manager = Mock.Of<IConfigurationManager<OpenIddictConfiguration>>();

    await using var server = await CreateServerAsync(options =>
    {
        options.Configure(options => options.ConfigurationManager = manager);
    });

    await using var client = await server.CreateClientAsync();

    // Act
    var response = await client.PostAsync("/authenticate", new OpenIddictRequest());

    // Assert
    Assert.Equal(0, response.Count);
    Mock.Get(manager).Verify(manager => manager.GetConfigurationAsync(It.IsAny<CancellationToken>()), Times.Never());
}
```

2d1a0ec3: Thanks Kevin! :) You are of course right again, the updated test also wasn't good. I removed the test completely and if it is all right with you, I would gladly accept your offer and leave creating the proper test to you :)

kevinchalet:
PR merged: congrats for your first contribution to OpenIddict 🎉

PR #1774, opened 2023-05-20
Update the LinkedIn provider to use "Sign In with LinkedIn V2"

kevinchalet:
The userinfo endpoint used by the LinkedIn provider points to the unversioned LinkedIn Marketing API that will be removed soon. While it would be easy to use the versioned alternative, this PR goes further by moving from the OAuth 2.0-only "Sign In with LinkedIn" service to the newer OpenID Connect implementation called "Sign In with LinkedIn V2" that offers a standard userinfo endpoint.

~Note: this change is binary breaking (given that the `Fields` option is being removed).~ It's also important to note that different claims are returned by the standard OIDC userinfo endpoint.


kevinchalet:
To avoid the binary breaking change, I decided to keep the `Fields` option and update the generator to support marking it as obsolete.

PR #1780, opened 2023-06-02
Update OpenIddictParameter.Count to use the correct nullable annotations

kevinchalet:


kevinchalet:
Integrated into https://github.com/openiddict/openiddict-core/pull/1781.

PR #1786, opened 2023-06-05
Fix compatibility issue with ef 8

ismayil-musayev:
In ef 8 preview a new parameter has been added to the [LoadAsync](https://github.com/dotnet/efcore/blob/f7b80507a6904f764e178e99a9926658af62320b/src/EFCore/ChangeTracking/ReferenceEntry.cs#L115) method
This pr fixes this compatibility issue

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreAuthorizationStore.cs
@@ -446,7 +446,7 @@ public class OpenIddictEntityFrameworkCoreAuthorizationStore<TAuthorization, TAp
                 return null;
             }
 
-            await reference.LoadAsync(cancellationToken);
+            await reference.LoadAsync(cancellationToken: cancellationToken);
         }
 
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR.

Unfortunately, this won't fix the issue you're seeing. For that, we'll need to compile against EF Core 8.0 for the compiler to target the new `LoadAsync()` signature taking a `LoadOptions` parameter, which is a much bigger fix that requires targeting .NET 8.0 and referencing the EF Core 8.0 packages.

I opened https://github.com/openiddict/openiddict-core/issues/1789 to track that and informed the EF team this breaking change is not documented: https://github.com/dotnet/efcore/issues/30306#issuecomment-1577009150.

ismayil-musayev:
My fault.

Will you accept pr if I change this line
`await reference.LoadAsync(cancellationToken)`
to this
`await reference.Query().ToListAsync(cancellationToken);`
as a workaround?

It will send the exact same query to a database.

ismayil-musayev:
@kevinchalet any comments regarding new approach?

kevinchalet:
It's likely a good workaround (thanks!), but I'd like to hear back from the EF folks before changing anything.

PR #1836, opened 2023-07-19
Disable userinfo retrieval for the ADFS provider

### LINKED ISSUE #1835: ADFS provider doesn't work with OpenIdDict client because of accessing the userinfo endpoint
### Personal contribution

- [ ] I'm not interested in submitting a pull request and understand that this provider may not be fixed in the near future without my contribution.

### Version

4.x

### Provider name

ADFS 2022

### Describe the bug

When connecting and using ADFS, I am able to get all the id and access tokens.  But, after acquiring the tokens, OpenIdDict client attempts to retrieve extra info from the userinfo endpoint. ADFS doesn't return any useful information from this endpoint and I couldn't find a way to ignore this and think that it should have an option to skip this extra step. e.g. like IdentityModel.OidcClient allows.  LoadProfile = false

I think the right thing to do with ADFS is avoid calling the userinfo endpoint all together, I would say that when using response_type=code where you get both id token and access token, it should be an option with any provider to go back and query the userinfo endpoint.

I was trying to follow the pipeline in the openiddict client to see if I could implement the LoadProfile = false, but I wasn't able to get the right spot.  

Do you think this is something useful?

### To reproduce

Tried adding the event callback to skip by setting HandleRequest or SkipRequest but it just hangs and never returns.  
Added Provider
                       .AddActiveDirectoryFederationServices(options =>
                       {
                           options.SetProviderName("ADFSTest")
                               .SetIssuer("https://xxx.com/adfs")
                               .SetClientId("c8cf672f-xxxxx-eee306376ce3")
                               .SetRedirectUri("http://localhost:7891/callback")
                               .AddScopes(Scopes.OpenId)
                               //.SetResource("urn:microsoft:userinfo");
                           .SetResource("c8cf672f-xxxxx-eee306376ce3");//Required for ADFS with AD Relying Party
                       });
....
                options.AddEventHandler<OpenIddictClientEvents.ApplyUserinfoRequestContext>(builder =>
                {
                    builder.UseInlineHandler(context =>
                    {
                        if (context != null)
                        {
                            AnsiConsole.MarkupLine("[green]ApplyUserinfoRequestContext[/]:{0}", context.Request);
                            //context.HandleRequest();//Don't send to ADFS endpoints
                            //context.SkipRequest();
                        }
                        else
                        {
                            AnsiConsole.MarkupLine("[red]No Token in Context [/]");
                        }
                        return default;
                    });
                    builder.SetOrder(OpenIddictClientHandlers.Protection.ValidateIdentityModelToken.Descriptor.Order - 500);
                });

### Exceptions (if any)

```shell
Error: invalid_token
 Message: An error occurred while authenticating the user.
 Description: The userinfo request was rejected by the remote server.

The ADFS error is UserInfoListener.ValidateAccessToken: The access token in the request doesn't have required audience 'urn:microsoft:userinfo'. Actual audience 'microsoft:identityserver:c8cf672f-xxxx-eee306376ce3'

This is because the access token doesn't audience will no longer match the resource sent as ADFS prefixes microsoft:identityserver: and as in your examples, it expects "urn:microsoft:userinfo" to be in the resource and aud.  so changing the resource to urn:microsoft:userinfo doesn't help.
"aud": "microsoft:identityserver:c8cf672f-xxxxx-eee306376ce3"

Each application will use a different clientid, so setting the client id to urn:microsoft.userinfo is not the right thing to do.  Setting the resource to urn:microsoft.userinfo and using a different client id will deny the request to the access token and id token because they don't match.
```

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/1835.

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
@@ -677,6 +677,14 @@ public static partial class OpenIddictClientWebIntegrationHandlers
 
             context.SendUserinfoRequest = context.Registration.ProviderType switch
             {
+                // Note: ADFS has severe restrictions affecting the ability to access the userinfo endpoint
+                // (e.g the "resource" parameter MUST be null or the "urn:microsoft:userinfo" value MUST be
+                // used, which prevents specifying any other resource as only one value is allowed by ADFS).
+                //
+                // Since the userinfo endpoint returns very limited information anyway,
+                // userinfo retrieval is always disabled for the ADFS provider.
+                ProviderTypes.ActiveDirectoryFederationServices => false,
+
                 // Note: the frontchannel or backchannel access tokens returned by Azure AD when a
                 // Xbox scope is requested cannot be used with the userinfo endpoint as they use a
                 // legacy format that is not supported by the Azure AD userinfo implementation.


eholyat:
Tried it, it works well.  Thanks for this.  Until it is rolled out, I added the following to my client and it works fine.  I'm not really sure how to properly gauge what the SetOrder is supposed to be, -1 or -500.  Is there a guide to this?
                options.AddEventHandler<OpenIddictClientEvents.ProcessAuthenticationContext>(builder =>
                {
                    builder.UseInlineHandler(context =>
                    {
                        context.SendUserinfoRequest = false;  
                        return default;
                    });
                    // make sure this is before userinfo access
                    builder.SetOrder(OpenIddictClientHandlers.AttachUserinfoRequestParameters.Descriptor.Order - 500);
                });

kevinchalet:
The best option is to use the 4.7.0 nightly builds you can find on MyGet.org.

> I'm not really sure how to properly gauge what the SetOrder is supposed to be, -1 or -500.

I'd go with simply `EvaluateUserinfoRequest.Descriptor.Order + 1` here. Note that your handler will basically disable userinfo retrieval unconditionally, for all providers, for all client registrations.

PR #1850, opened 2023-08-01
Add MusicBrainz to the list of supported providers

steveberdy:


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -633,6 +633,24 @@
...
+                     UserinfoEndpoint="https://musicbrainz.org/oauth2/userinfo">
+        <GrantType Value="authorization_code" />
+        <GrantType Value="refresh_token" />
+      </Configuration>
+
+      <!--
+        Note: MusicBrainz requires specifying a scope on authorization, with "profile" being the most generic scope and thus the default
+      -->
+
+      <Scope Name="profile" Default="true" Required="true" />
```
kevinchalet, line 650: `Required="true"` is typically used when the userinfo endpoint cannot be accessed without the scope: do you confirm MusicBrainz' userinfo API returns an error if you don't request `profile`? (e.g `submit_isrc`)

kevinchalet (commented):
Thanks for your PR! 👍🏻 
Do you confirm the provider works fine?

It's worth noting OpenIddict 4.7 will include a built-in claims mapping feature to map well-known claims like the user ID or the username.

I updated the web provider contribution guide to explain what's required to make a provider integration compatible: https://documentation.openiddict.com/guides/contributing-a-new-web-provider.html#if-the-provider-doesnt-support-standard-openid-connect-userinfo-map-the-provider-specific-claims-to-their-claimtypes-equivalent. Since MusicBrainz doesn't support OIDC, it's likely they use custom parameters: that would be cool if you could update the built-in handler to support MusicBrainz.

kevinchalet:
I see in MusicBrainz' docs that they support 2 custom parameters that would likely be useful to have:

```
access_type (optional)
    Indicates if your application needs to access the API when the user is not present at the browser. This parameter defaults to online. If your application needs to refresh access tokens when the user is not present at the browser, then use offline. This will result in your application obtaining a refresh token the first time your application exchanges an authorization code for a user.

approval_prompt (optional)
    Indicates if the user should be re-prompted for consent. The default is auto, so a given user should only see the consent page for a given set of scopes the first time through the sequence. If the value is force, then the user sees a consent page even if they have previously given consent to your application for a given set of scopes.
```

Supporting such settings shouldn't be too hard. Here's how we do that for Google:

https://github.com/openiddict/openiddict-core/blob/a812ca46711d511d1b2dd3b9c237ed85fe4cfbbe/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml#L415-L416

https://github.com/openiddict/openiddict-core/blob/a812ca46711d511d1b2dd3b9c237ed85fe4cfbbe/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs#L1422-L1429

kevinchalet:
Closing due to inactivity.

PR #1851, opened 2023-08-01
Add Naver to the list of supported providers 

steveberdy:


kevinchalet:
Thanks! Same question as for the other PR: did you have a chance to confirm the provider works fine with the real service?

When you have a moment, please update the claims mapping handler as for the other providers.

kevinchalet:
Closing due to inactivity.

PR #1853, opened 2023-08-01
Suppress indentation on json response

cllio:
I noticed that in my token reply I have indentation with spaces and line breaks.


kevinchalet:
Thanks for your PR!

Returning indented JSON responses is actually deliberate as it makes token and configuration responses easier to read by human beings (at least for those who don't use Firefox and its native beautifier 😄).

Apple and Google do that too for similar reasons:
  - https://appleid.apple.com/.well-known/openid-configuration
  - https://accounts.google.com/.well-known/openid-configuration

That said, we should probably make that configurable for whose who want to save a few bytes (note: response caching and compression is generally way more rewarding).
I'll open a ticket to track that.

cllio:
yes I understand better the code with the interest to keep it for the .well-known but it's a pity to share this configuration with in my case the answer to the client_creditential even if indeed it's only a few bytes gained. Then it's not a big problem either
I thought it was a simple omission ;-)

kevinchalet:
I opened https://github.com/openiddict/openiddict-core/issues/1855 😃

Issue #1876, opened 2023-09-09
I want to make OpeniddictDbContext without inject it for other reasons


wisamidris7:
### Confirm you've already contributed to this project or that you sponsor it

- [X] I confirm I'm a sponsor or a contributor

### Describe the solution you'd like

# I want to make OpeniddictDbContext it new instance but without inject it

Cause There Some Other Issues And Things That Require To Not Inject
I Want To Make Like This Cause This Is Important
```c#
builder.Services.AddOpenIddict()

    // Register the OpenIddict core components.
    .AddCore(options =>
    {
        // Configure OpenIddict to use the Entity Framework Core stores and models.
        // Note: call ReplaceDefaultEntities() to replace the default OpenIddict entities.
        options.UseEntityFrameworkCore()
           .UseDbContext(() => new OpeniddictDbContext()); // Required Design

        // Enable Quartz.NET integration.
        options.UseQuartz();
    })
    ....
```

kevinchalet:
Hi,

Support is reserved to sponsors and contributors. For more information on how to sponsor the project on GitHub, visit https://github.com/sponsors/kevinchalet.

Hope to see you on board soon!

PR #1884, opened 2023-09-19
I Maked EntityFrameworkCore Flexable For Use

cursor-background-agent-01:
Here You Can Use These Ways

Way 1

```c#
builder.UseEntityFrameworkCore()
       .UseDbContext<DbContext>();
```

Way 2 Here What I Do

```c#
public class CustomOpeniddictDbContextFactory : IOpeniddictEntityFrameworkCoreContextFactory 
{
     public DbContext CreateDbContext()
     {
            return ...;
     }
}

builder.UseEntityFrameworkCore()
       .UseDbContextFactory<CustomOpeniddictDbContextFactory>();
```

Way 3 Also This

```c#
builder.UseEntityFrameworkCore()
       // This
       .UseDbContext(e => e.GetService<...>())
       // Or
       .UseDbContext(() => new DbContext());
```

This Is Me @wisamidris7
After I Think You Right It's Hard To Work For Free If
I Was In Your Place I Was Gonna Do The Same

https://github.com/openiddict/openiddict-core/issues/1876

And Thanks

Don't Forgot To Unblock Me

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/Factory/IOpeniddictEntityFrameworkCoreContextFactory.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Factory/IOpeniddictEntityFrameworkCoreContextFactory.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OpenIddict.EntityFrameworkCore.Factory;
+
+public interface IOpeniddictEntityFrameworkCoreContextFactory
+{
+    DbContext CreateDbContext();
+}
--- a/src/OpenIddict.EntityFrameworkCore/Factory/OpeniddictEntityFrameworkCoreContextFactory.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Factory/OpeniddictEntityFrameworkCoreContextFactory.cs
@@ -1,0 +1,29 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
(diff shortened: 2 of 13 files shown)


hishamqashan77:
This Is Me wisamidris7

and i maked what i want in this issue
https://github.com/openiddict/openiddict-core/issues/1876

And Thanks

Don't Forgot To Unblock Me

cursor-background-agent-01:
Way 2 Here What I Do
```
public class CustomOpeniddictDbContextFactory : IOpeniddictEntityFrameworkCoreContextFactory 
{
     public DbContext CreateDbContext()
     {
            return ...;
     }
}

// Forgot This
builder.Services.AddScoped<CustomOpeniddictDbContextFactory>();

builder.UseEntityFrameworkCore()
       .UseDbContextFactory<CustomOpeniddictDbContextFactory>();
```

kevinchalet:
It's not clear at all why you'd need such a hook ("I Want To Make Like This Cause This Is Important" is definitely not a good justification 🤣).

You can already achieve a similar thing with the built-in DI extensions:

```csharp
services.AddScoped(provider => new ApplicationDbContext(...));

services.AddOpenIddict()
    .AddCore(options =>
    {
        options.UseEntityFrameworkCore()
               .UseDbContext<ApplicationDbContext>();
    });
```

The design you suggest is also problematic because it allows creating `DbContext` objects outside the DI container but the lifetime of these instances is not tracked and they are never disposed (when objects are created for you by the DI container, they are automatically disposed when the service scope/container itself is disposed).

This approach could lead to memory leaks, so it's definitely not going to meet the bar to be included in OpenIddict.

PR #1886, opened 2023-09-20
The Story Is That I'm Making Multiency System Like Abp

cursor-background-agent-01:
And I Maked DbContext Named

AppDbContext

```c#
public class AppDbContext : DbContext
{
      public Tenant CurrentTenant { get; }
      public AppDbContext(Session session)
      {
            CurrentTenant = session.Tenant; // Here I Get Tenant From Claims And I Have TenantId In Claims
      }
      public void OnModelCreating(ModelBuilder builder)
      {
           // some business about connection string
      }
}
```

```c#
builder.AddDbContext<AppDbContext>(e =>
{
         e.UseSqlServer("none");
});
```

Here When Going To Openiddict The Host Db Work
But When Login From Tenant It Says Authorization Not Found And Some Times Application Not Found

So The ConnectionString Of The Host Is
"Server=.;Database=hostDb" And I Seed Data Into It

But In Tenant The Connection String Is Not The Same
"Server=.;Database=tenant1Db" Here Authorization Not Same
Cause Authorization Added Into Host

And After It He Checks The In tenant1Db

So What I Do 

```c#
// This Is In AddOpeniddict
builder.UseEntityFrameowrk()
     .UseDbContext(() => new AppDbContext(new Session(null))); // Here I Get The TenantId
     // When CurrentTenant Is Null I Make Host ConnectionString So This Is The Only Way
```

About The Additional Things I Added Is To Make It More And More Flexible For Other Developer

## I Don't Care To Be Contributor I Just Want To Complete My Website

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/Factory/IOpeniddictEntityFrameworkCoreContextFactory.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Factory/IOpeniddictEntityFrameworkCoreContextFactory.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OpenIddict.EntityFrameworkCore.Factory;
+
+public interface IOpeniddictEntityFrameworkCoreContextFactory
+{
+    DbContext CreateDbContext();
+}
--- a/src/OpenIddict.EntityFrameworkCore/Factory/OpeniddictEntityFrameworkCoreContextFactory.cs
+++ b/src/OpenIddict.EntityFrameworkCore/Factory/OpeniddictEntityFrameworkCoreContextFactory.cs
@@ -1,0 +1,29 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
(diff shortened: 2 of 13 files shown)


kevinchalet:
The scenario you describe is very common and you don't need any extra hook to implement it: https://github.com/openiddict/openiddict-core/pull/1884#issuecomment-1727404758.

cursor-background-agent-01:
I Don't Understand

cursor-background-agent-01:
It's Because The AppDbContext Is Injected By TenantId From Claims

kevinchalet:
https://learn.microsoft.com/en-us/ef/core/miscellaneous/multitenancy

cursor-background-agent-01:
This Because I Maked Each Database Can Contains Many Tenants
And You Can Make For Each Tenant A Database

cursor-background-agent-01:
I Has Big App Thats Been Like Abp

Editions
Tenants
Users
Roles
And Others

kevinchalet:
Again, you don't need a new hook to support that: you can resolve any service you need from your `DbContext` constructor, including a tenant service that can retrieve the tenant details from any place you want.

Anyway, this repository is not the right place for generic questions about EF Core: consider posting your question on StackOverflow if you still need help.

PR #1890, opened 2023-09-26
Fix AuthenticateWithDeviceAsync() to flow the scopes attached to the request model

kevinchalet:


kevinchalet:
Note: this PR will be backported to 4.9.0.

Issue #1797, opened 2023-06-17
Replace the default IMemory cache with IDistributedCache 


jacojvv-dev:
### Confirm you've already contributed to this project or that you sponsor it

- [X] I confirm I'm a sponsor or a contributor

### Version

4.x

### Question

Hi.

I am just curious about using IDistributedCache instead of IMemory cache for caching entities. Specifically I am looking to replace the in memory caching for the application manager to use Redis instead.

Currently I can't see a way of doing it without implementing my own manager, resolver and possibly cache.

Any guidance in this regard would be greatly appreciated!

Thanks again for all the work you put into this project!

kevinchalet:
Hey,

> I am just curious about using IDistributedCache instead of IMemory cache for caching entities. Specifically I am looking to replace the in memory caching for the application manager to use Redis instead.

It's important to note that the in-memory cache you're referring to is actually quite specific and likely different from what you're expecting as it's actually always created per-scope: cached entities are not shared between multiple ASP.NET Core requests.

The reason it exists is due to how the OpenIddict client, server and validation stacks are designed: while they form a chained pipeline, all the event handlers that are responsible for the request processing are actually independent from each other. As a result, multiple handlers will always need to retrieve the same entity - e.g the client application - from the DB at different places in the pipeline: if we hit the DB each time a handler needed to resolve the same entity, that would be terrible from a performance perspective. The built-in scoped cache mitigates that by storing these entities in an in-memory cache so that all the handlers get the same instance.

(in a way, you can see that as an equivalent of NHibernate's first-level cache concept 😃)

If you wanted to implement a distributed cache working across multiple requests, custom stores would be the right place for that (no need for custom managers or resolvers).

> Thanks again for all the work you put into this project!

Thanks for your kind words! ❤️

jacojvv-dev:
Thanks. That explanation makes a lot of sense to me. I was under the impression that the cache is shared between requests but come to think of it it makes sense that it is scoped to ensure the behavior above. (I was also wondering how that 250 default limit is not being hit across all the different requests - but - turns out I am just a little bit of an idiot and should spend more time reading the source before making assumptions 😄)

The issue I am trying to resolve is caching across several requests as you outlined above. Specifically the system where I introduced this where there are not many applications. Caching in a distributed manner will save several requests to the database between different .Net requests.

[...]

PR #1898, opened 2023-10-02
Add absolute expiration of caches

yfakariya:
It saves the case that create/update/delete logics and retrieval logics exist in seperate processes and "eventual consistency" of the change is acceptable. In such case, `CancellationTokenSource` based expiration does not work because the database will be changed without any notification in the "reader" process(es).

I've just implement absolute-to-now expiration because sliding expiration does not rescue above case.


kevinchalet:
Thanks for your PR!

It's worth noting these caches are not meant to be shared across requests or processes. See https://github.com/openiddict/openiddict-core/issues/1797#issuecomment-1595724230 for an explanation of why these caches exist.

yfakariya:
Thank you for quick and nice response! 
I read the your notice, I understand that the caches are "scoped" and they just prevent busy access for the datastore in single request scope. So, I will close this PR because it is useless.

PR #1902, opened 2023-10-18
feat: Add BackgroundService pruning job

davhdavh:
Adds an alternative to using Quartz for pruning.

### DIFF
--- a/OpenIddict.sln
+++ b/OpenIddict.sln
@@ -155,6 +155,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenIddict.Sandbox.WinForms
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenIddict.Sandbox.Console.Client", "sandbox\OpenIddict.Sandbox.Console.Client\OpenIddict.Sandbox.Console.Client.csproj", "{819CD7AA-01FD-4369-BABF-5DFCB7E94068}"
 EndProject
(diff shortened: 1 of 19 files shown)


kevinchalet:
Thanks for your PR!

It's worth mentioning that we considered implementing an `IHostedService`/`BackgroundService` for that, but eventually rejected this option. See https://github.com/openiddict/openiddict-core/issues/361 for more information.

davhdavh:
Yeah. Ofcourse the documentation needs to mention that this method is for a low number of servers. But considering all the quartz demo just use in-memory (and thus per-server) setups anyway, I'm guessing most people don't actually run into any issues

kevinchalet:
I'd really like to avoid having two implementations that end up doing the same thing (except the custom `IHostedService`-based would be way less flexible than the Quartz module).

Is there any particular reason for not using Quartz?

davhdavh:
We already use another lib that does task scheduling. There is a ton of work to make a task scheduler work nicely in a canary deployment model, so not really looking to add another.

Nor would I want to add a huge task scheduling module, basically just to run 2 lines of Prune code once an hour.

With a simple `IHostedService`-based solution, it is quite trivial to include in the pod that deals with other periodic cleanup.

kevinchalet:
> With a simple IHostedService-based solution, it is quite trivial to include in the pod that deals with other periodic cleanup.

Adding an `IHostedService` in your own code is also trivial and doesn't require any change in OpenIddict itself.
Before integrating such a thing in OI itself, I'd like to wait and see if there's more demand from the community.

Closing for now 😃

davhdavh:
I published it as https://www.nuget.org/packages/Catglobe.Openiddict.Contrib.Pruning.BackgroundService

kevinchalet:
Great initiative 👍🏻

PR #1913, opened 2023-11-13
feat: add constants to match spec

davhdavh:
Some constants are missing from the spec in this section: https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter

### DIFF
--- a/src/OpenIddict.Abstractions/OpenIddictConstants.cs
+++ b/src/OpenIddict.Abstractions/OpenIddictConstants.cs
@@ -4,6 +4,8 @@
  * the license and the contributors participating to this project.
  */
 
+using static System.Net.WebRequestMethods;
+
 namespace OpenIddict.Abstractions;
 
 public static class OpenIddictConstants
@@ -156,13 +158,26 @@ public static class OpenIddictConstants
         }
     }
  
-    public static class ClaimsParameters
+    public static class ClaimRequestMembers
     {
-       public const string Essential = "essential";
-       public const string IdToken = "id_token";
-       public const string Userinfo = "userinfo";
-       public const string Value = "value";
-       public const string Values = "values";
+        /// <summary>
+        /// Individual claims requests members, as per <see href="https://openid.net/specs/openid-connect-core-1_0.html#IndividualClaimsRequests"/>.
+        /// </summary>
+        public static class Individual
+        {
+            public const string Essential = "essential";
+            public const string Value = "value";
+            public const string Values = "values";
+        }
+        
+        /// <summary>
+        /// Top level request members, as per <see href="https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter"/>.
+        /// </summary>
+        public static class TopLevel
+        {
+            public const string Userinfo = "userinfo";
+            public const string IdToken = "id_token";
+        }
     }
 
     public static class ClientAssertionTypes
--- a/src/OpenIddict.Abstractions/OpenIddictConstants.cs
+++ b/src/OpenIddict.Abstractions/OpenIddictConstants.cs
@@ -155,6 +155,15 @@ public static class OpenIddictConstants
             public const string UserCodeLifetime = "oi_usrc_lft";
         }
     }
+ 
+    public static class ClaimsParameters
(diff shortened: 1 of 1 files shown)


src/OpenIddict.Abstractions/OpenIddictConstants.cs
```diff
@@ -365,6 +365,15 @@ public static class Parameters
         public const string VerificationUriComplete = "verification_uri_complete";
     }
 
+    public static class ClaimsParameters
```
kevinchalet: `ClaimRequestMembers` may be a better name.

src/OpenIddict.Abstractions/OpenIddictConstants.cs
```diff
@@ -365,6 +365,15 @@ public static class Parameters
...
+    {
+       public const string UserInfo = "userinfo";
```
kevinchalet: Let's name the constant `Userinfo` without a capital I.

kevinchalet (commented):
Thanks for your PR.

Please move the new class to the right place so the alphabetical order is respected (same thing for the constants inside the class).

src/OpenIddict.Abstractions/OpenIddictConstants.cs
davhdavh: Considering they are called `ClaimParameter` in the spec, I went with that name.
I don't think `ClaimRequestMembers` is a good match though, perhaps `ClaimRequestParameters` or `RequestClaimParameters` or `ClaimRequestParametersMembers`? I still think the `ClaimsParameters` is the best though.

davhdavh: alternatively, a split design:
empty `ClaimRequest` and `ClaimRequest.RequestTypes` with `userinfo` and `id_token`, and `ClaimRequest.Members` with the rest.

kevinchalet: > Considering they are called `ClaimParameter` in the spec, I went with that name.

I'm quite familiar with the spec but I re-read it just to be sure and I don't think it's true. The only parameter mentioned in the 5.5 section is named `claims`: it's the JSON object that is UTF8+formurl-encoded and contains two top-level `id_token` and `userinfo` nodes. These nodes present in the claims request are simply called "members":

image

davhdavh: The url is `openid-connect-core-1_0.html#ClaimsParameter`.
In any case, what is your final preferred naming scheme and I shall update :)

kevinchalet: > The url is `openid-connect-core-1_0.html#ClaimsParameter`. In any case, what is your final preferred naming scheme and I shall update :)

Well, the anchor points to the section title, that clearly says "`claims` Request Parameter": only `claims` is technically defined as a parameter, not the JSON nodes defined in the section itself.

Let's go with `ClaimRequestMembers`. It's not a great name, but it's the most accurate one.

davhdavh: Ok. I felt if you wanted to very strictly keep to the nomenclature used in the actual text of the spec, that we should also split it into `Individual` and `TopLevel`

src/OpenIddict.Abstractions/OpenIddictConstants.cs
```diff
@@ -4,6 +4,8 @@
  * the license and the contributors participating to this project.
  */
 
+using static System.Net.WebRequestMethods;
```
kevinchalet: Looks like an unwanted change (VS can be very annoying with that...)

src/OpenIddict.Abstractions/OpenIddictConstants.cs
kevinchalet: Well, it's not wrong, but I'm not so sure it's worth being that specific (it makes names longer). Let's just flatten the class (please also remove the comments, the other classes don't have them).

kevinchalet:
Merged, thanks for your contribution 👏🏻

PR #1945, opened 2024-01-09
Add a MAUI (WinUI-only) client sandbox project

kevinchalet:
Let's see if it builds 😅


kevinchalet:
> Let's see if it builds 😅

No... discarding.

PR #1958, opened 2024-01-17
Abort sign-in demands whose principal contains a standard claim with an invalid claim value type

### LINKED ISSUE #1956: Consider enforcing strict `ClaimValueTypes` checks for standard claims

Claims like `sub` are required to be formatted as JSON `string`s. While OpenIddict requires that a `sub` claim be present, it doesn't prevent a different `ClaimValueType` from being set, which results in tokens being issued with a non-standard JSON type without the user necessarily realizing it. It later causes issues in IdentityModel when validating these tokens as seen in these issues:
  - https://github.com/openiddict/openiddict-core/issues/1952
  - https://github.com/openiddict/openiddict-samples/issues/283

### Additional context

_No response_

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/1956.

While it's technically a breaking change, claims added with an incorrect `ClaimValueTypes` resulted in non-standard tokens and errors that were sometimes hard to debug by users.

### DIFF
--- a/src/OpenIddict.Abstractions/OpenIddictResources.resx
+++ b/src/OpenIddict.Abstractions/OpenIddictResources.resx
@@ -1578,6 +1578,15 @@ To apply post-logout redirection responses, create a class implementing 'IOpenId
   <data name="ID0423" xml:space="preserve">
     <value>Multiple claims of the same type are present in the identity or principal.</value>
   </data>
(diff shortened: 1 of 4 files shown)


kevinchalet:
Note: sign-in demands are also aborted when specifying well-known claims with an incorrect cardinality (e.g multiple `sub` claims).

PR #1971, opened 2024-02-03
Update the EF Core stores to resolve the navigation properties via the change tracker when possible

kevinchalet:
Doing that allows eliminating 2 queries per token creation, which saves 6 roundtrips for a typical OpenID Connect `grant_type=refresh_token` request where an access token, an identity token and a refresh token are created.


kevinchalet:
Note: the EF 6 stores are unaffected as they are using `FindAsync()` to load these navigation properties.

PR #1973, opened 2024-02-06
Use handlers to trim and rehydrate the refresh token

AlbertSchreuder-0:
This code removes all the claims from the refresh token except for a handful. and then rehydrates it when a new grant_type=refresh_token call is made.

We are also generating an unencrypted refresh token (it is signed).

```
eyJhbGciOiJSUzI1NiIsImtpZCI6ImJaZkJSLVRFMC1TRExKb3hqSHhHZml4NkozcFNUNGlaTTJzUlZ4OHVoMXMiLCJ0eXAiOiJvaV9yZWZ0K2p3dCJ9.eyJleHAiOjE3MDgxMTg1ODAsImlhdCI6MTcwNzI1NDU4MCwib2lfYXVfaWQiOiI2YjJhMDU2Yy0xMzM0LTQ0ZGMtODYwZC03MGI1Y2Q2MzI5ZTQiLCJvaV90a25faWQiOiIyMGIxMGIwZC04MzYwLTRlZGQtOTI4OC01M2MyOGY0YzQzZjcifQ.ZnMTrSw9Oqek9bIZdL3-BYpiNxnz4hvhDqO-qvP9LmwImOrEwZjVwZQU8nxv9nLH5bpaYoFazh4sV3Tp1fJMRykBhmGgrmx0-slUtoC0QFEdYK7efOcRqWyPLw7e1EzRtW_3abPEn8ItNRGE6eixH3-pNFvelyHiNeLokuLVWI1Oj0zC1hX4roS0Mdp_VrzC8TNgFhJSYivSp7vxYdbonfNGXP1fllmRFVcLKkrpquEUGrwrPcdQGDhpWAcJUrdRvpJAhqymJa9MpP_iHXQU9fFD32vtcPO_ebQb0WFflFAeK0b6sXqPXREn6zl4eCmvIDcdHVAnmfj4p2FqtZP6UA
```
With the contents:
```
{
  "exp": 1708118580,
  "iat": 1707254580,
  "oi_au_id": "6b2a056c-1334-44dc-860d-70b5cd6329e4",
  "oi_tkn_id": "20b10b0d-8360-4edd-9288-53c28f4c43f7"
}
```
When we rebuild the token:
We are consuming the full token from the http request, and and decoding it back to claims. We then re-hydrate the claims with claims we removed.

We then have two options.
-   We either generate a principal from the claims we already have.  And thereby short circuit the next call in the OIDT chain(ValidateIdentityModelToken)
- Or we use the fully built claims to rebuild a new token payload, and we continue down the OIDT that will decode it and build the identity model.
(both options work for me)

### DIFF
--- a/sandbox/OpenIddict.Sandbox.AspNetCore.Server/CustomServerHandlers.cs
+++ b/sandbox/OpenIddict.Sandbox.AspNetCore.Server/CustomServerHandlers.cs
@@ -22,9 +22,6 @@ namespace OpenIddict.Sandbox.AspNetCore.Server
             {
             }
 
-            /// <summary>
-            /// Gets the default descriptor definition assigned to this handler.
-            /// </summary>
             public static OpenIddictServerHandlerDescriptor Descriptor { get; }
                 = OpenIddictServerHandlerDescriptor.CreateBuilder<ProcessSignInContext>()
                     .AddFilter<RequireRefreshTokenGenerated>()
@@ -37,7 +34,7 @@ namespace OpenIddict.Sandbox.AspNetCore.Server
                     .Build();
 
 
-            public async ValueTask HandleAsync(OpenIddictServerEvents.ProcessSignInContext context)
+            public ValueTask HandleAsync(OpenIddictServerEvents.ProcessSignInContext context)
             {
                 if (context is null)
                 {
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR. That said, it looks quite specific to your scenario and isn't suitable for general use so I'm not going to merge it.

> We are also generating an unencrypted refresh token (it is signed).

It's a terrible idea: authorization codes and refresh tokens are always encrypted because they can contain very sensitive data (e.g authorization codes contain the `code_challenge` used for PKCE and while refresh tokens currently don't include claims that NEED to be kept private in most scenarios, it's not a guarantee and it may change at any point in the future, if new features require it).

PR #1975, opened 2024-02-09
Fix the ASP.NET Core Data Protection formatters to use the correct value type for the token lifetime claims

### LINKED ISSUE #1974: Data Protection causes claim value type to default to string

### Version

5.1.0

### Describe the bug

Hi,

we are currently in the process of migrating to OpenIddict 5 and have noticed an issue with the value type of the claims in combination with Data Protection.

It seems that when using Data Protection, the value type will always default to string, causing issues with the strict value type checks added in https://github.com/openiddict/openiddict-core/pull/1958

If you for example look at the Velusia.Server sample, it will work fine even with data protection because all of the claims used already seem have the value type string. But when adding a claim with a different value type, like for example `oi_act_lft`, the value type will change from integer to string when the principal is retrieved on the `/connect/token` endpoint, resulting in an exception.

### To reproduce

https://github.com/schneini/openiddict-samples/commit/865b872e404ef63e72543d935d3669a6b6c025b6

### Exceptions (if any)

```shell
System.InvalidOperationException: The 'oi_act_lft' claim present in the specified principal is malformed or isn't of the expected type.
```

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/1974.

### DIFF
--- a/src/OpenIddict.Validation.DataProtection/OpenIddictValidationDataProtectionFormatter.cs
+++ b/src/OpenIddict.Validation.DataProtection/OpenIddictValidationDataProtectionFormatter.cs
@@ -35,21 +35,10 @@ public sealed class OpenIddictValidationDataProtectionFormatter : IOpenIddictVal
 
             .SetClaim(Claims.Private.HostProperties, GetJsonProperty(properties, Properties.HostProperties))
 
-            .SetClaim(Claims.Private.AccessTokenLifetime,       GetProperty(properties, Properties.AccessTokenLifetime))
-            .SetClaim(Claims.Private.AuthorizationCodeLifetime, GetProperty(properties, Properties.AuthorizationCodeLifetime))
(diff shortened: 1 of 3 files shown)


src/OpenIddict.Validation.DataProtection/OpenIddictValidationDataProtectionFormatter.cs
```diff
@@ -35,21 +35,10 @@ public ClaimsPrincipal ReadToken(BinaryReader reader)
 
             .SetClaim(Claims.Private.HostProperties, GetJsonProperty(properties, Properties.HostProperties))
 
-            .SetClaim(Claims.Private.AccessTokenLifetime,       GetProperty(properties, Properties.AccessTokenLifetime))
```
kevinchalet, line 38: Note: most of these mappings were useless as the validation stack only processes DP access tokens and these properties/claims are not present in access tokens.

PR #1998, opened 2024-02-21
Bump the .NET SDK to 8.0.200 and stop using collection expressions with ImmutableArray<T>

kevinchalet:
Related ticket: https://github.com/dotnet/roslyn/issues/72170.


teo-tsirpanis:
I wonder why you didn't bump the `System.Collections.Immutable` package. 1.7.1 is very old.

kevinchalet:
@teo-tsirpanis for most projects, that would indeed be the easiest option but OpenIddict enforces strict rules regarding dependencies. One of them is that each TFM is only allowed to reference packages that have the same support policy as the TFM itself. For instance, the `net7.0` version can only reference the 7.0.x version of the `Microsoft.Extensions.*` packages to ensure we don't accidentally end up depending on NuGet packages that have a shorter support lifecycle than the runtime itself.

`System.Collections.Immutable` `1.x` is indeed not recent, but it's still supported and is covered by the super-LTS "ASP.NET Core 2.1 on .NET Framework 4.x" policy, which is why it's used by our .NET Framework 4.6.1+ and .NET Standard 2.0 TFMs here.

Hope it was clear 😃

kevinchalet:
Here's the `Directory.Packages.props` if you're interested in taking a look to see how it's done: https://github.com/openiddict/openiddict-core/blob/dev/Directory.Packages.props 😄

PR #2030, opened 2024-03-11
Allow specifying the domain for the correlation cookie

meywd:
Allow specifying the domain for the correlation cookie to [allow subdomain communication](https://stackoverflow.com/questions/18492576/share-cookies-between-subdomain-and-domain).


kevinchalet:
Thanks for your PR!

Since this is a very specific scenario (that I don't really want to encourage), I don't think it would make a ton of sense to add a builder method for that.

Note: you can already configure that yourself by calling `options.Configure(options => options.CookieBuilder.Domain = "example.com")` without needing a new API.

meywd:
sure thanks for pointing that out.

PR #2039, opened 2024-03-14
Add Clever to the list of supported providers

anarian:
Add Clever SSO to supported OpenID providers.

- It looks like they do not return the `id_token` with the response similar to Paypal, let me know if this is the correct way to set the configuration
- Scopes are handled by Clever with the configuration in their Partner portal

image

Let me know if you need any information for any additional testing or verification 😃

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -479,11 +479,11 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                     context.Response[Claims.Subject] = (string?) context.Response[Claims.Subject];
                 }
                 
-                // Note: Clever returns a non-standard "name" claim formatted as a nested JSON string
+                // Note: Clever returns a non-standard "name" claim formatted as a JSON object.
                 else if (context.Registration.ProviderType is ProviderTypes.Clever)
                 {
                     var name = context.Response[Claims.Name]?.GetNamedParameters();
-                    if (name != null)
+                    if (name is not null)
                     {
                         context.Response[Claims.Name] = $"{name["first"]} {name["last"]}";
                         context.Response[Claims.FamilyName] = name["last"];
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -478,6 +478,18 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                 {
                     context.Response[Claims.Subject] = (string?) context.Response[Claims.Subject];
                 }
+                
+                // Note: Clever returns a non-standard "name" claim formatted as a nested JSON string
+                else if (context.Registration.ProviderType is ProviderTypes.Clever)
+                {
+                    var name = context.Response[Claims.Name]?.GetNamedParameters();
+                    if (name != null)
+                    {
+                        context.Response[Claims.Name] = $"{name["first"]} {name["last"]}";
+                        context.Response[Claims.FamilyName] = name["last"];
+                        context.Response[Claims.GivenName] = name["first"];
+                    }
+                }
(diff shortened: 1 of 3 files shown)


kevinchalet:
Hey @anarian,

Thanks for your PR!

> It looks like they do not return the id_token with the response similar to Paypal, let me know if this is the correct way to set the configuration

Hum, it's weird, their documentation explicitly mentions that an `id_token` should be returned:

> The response looks like (id_token only present for apps configured for OIDC):

Could that be a configuration issue?

It also looks like their userinfo endpoint returns a non-standard `name` claim: the specification explicitly says it's a string and they return a JSON object. We'll probably need to quirk it in this handler:

https://github.com/openiddict/openiddict-core/blob/3bc72160a443fb5ce24189ad55d064a53b07e431/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs#L450-L483

Can anyone create a Clever test account or is it application/invite-only?

anarian:
Yes it looks like I missed a [section in their documentation](https://dev.clever.com/docs/oidc-implementation), I have to email them to enable OIDC responses.

I'll make that change for the name claim and test it as soon as they respond.

For a test account, they ask that you contact them to set up a developer account and sandbox district.
https://www.clever.com/developer-signup
I can send you a user account invite to our sandbox district if you need one.

anarian:
21722fd6_VsDebugConsole_jbst7RTEwg

After contacting Clever support to enable OIDC scopes, the `id_token` is now working. 
I've also updated userinfo to flatten the nested JSON to a first+last name, as well as separate them out to the given and family name claims.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
```diff
@@ -478,6 +478,18 @@ public ValueTask HandleAsync(ExtractUserinfoResponseContext context)
                 {
                     context.Response[Claims.Subject] = (string?) context.Response[Claims.Subject];
                 }
+                
+                // Note: Clever returns a non-standard "name" claim formatted as a nested JSON string
```
kevinchalet:
```suggestion
                // Note: Clever returns a non-standard "name" claim formatted as a JSON object.
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
```diff
@@ -478,6 +478,18 @@ public ValueTask HandleAsync(ExtractUserinfoResponseContext context)
...
+                else if (context.Registration.ProviderType is ProviderTypes.Clever)
+                {
+                    var name = context.Response[Claims.Name]?.GetNamedParameters();
+                    if (name != null)
```
kevinchalet:
```suggestion
                    if (name is not null)
```

kevinchalet:
> After contacting Clever support to enable OIDC scopes, the id_token is now working.
I've also updated userinfo to flatten the nested JSON to a first+last name, as well as separate them out to the given and family name claims.

Excellent, thanks! ❤️

kevinchalet:
One last thing: when starting the authorization flow, does Clever offer a way to deny the authorization demand? If so, could you please test it and see if an `access_denied` error is returned with the `state` attached?

anarian:
They do support denying the authorization demand for Teachers, but they do not return an `access_denied` error.

c3951973_firefox_L064FJCUYz

I had to use the sandbox ASP.NET Core Client to test this, the sandbox console client simply hangs waiting for the request because their script just closes the window without any response.

kevinchalet:
Merged. Congrats for your first contribution to OpenIddict! ❤️ 

> They do support denying the authorization demand for Teachers, but they do not return an access_denied error.

Thanks! That's a bit unfortunate but it's not like it would be the first provider to do that 😅 

> I had to use the sandbox ASP.NET Core Client to test this, the sandbox console client simply hangs waiting for the request because their script just closes the window without any response.

Note: you can hit CTRL+C to abort the authentication process. In a real world app, you'd also want to add a `CancellationTokenSource` that auto-cancels after a bit of time automatically but we're not doing that in the sandbox 'cause it wouldn't be very practical 😃

PR #2040, opened 2024-03-15
Add Classlink to the list of supported providers

anarian:
Add Classlink SSO to the list of supported providers

Quirks: 
- Their openid configuration is returning `"code"` rather than `"authorization_code"`, append the correct value at runtime in `AmendGrantTypes`
- The userinfo endpoint is returning `address` as JSON, separate it to `locality` and `region` claims.

d8b6a43a_VsDebugConsole_1jEAts1I8O

Similar to [Clever](https://github.com/openiddict/openiddict-core/pull/2039), they do require [contacting them](https://www.classlink.com/company/prospective-partners) to set up a partner account.


kevinchalet:
> * Their openid configuration is returning `"code"` rather than `"authorization_code"`, append the correct value at runtime in `AmendGrantTypes`

Returning `code` in `response_types_supported` is standard-compliant. I just checked and they actually don't return a `grant_types_supported` node (in which using `code` instead of `authorization_code` would be illegal), so OpenIddict assumes `authorization_code` and `implicit` are supported (as per the spec).

> * The userinfo endpoint is returning `address` as JSON, separate it to `locality` and `region` claims.

The OIDC standard requires formatting the `address` claim as a JSON object so they're doing the right thing 👍🏻

anarian:
Ah I see, I was looking at the wrong field. I think the interactive service doesn't have that assumption, which is why I thought it wasn't working. [`GetSelectedGrantTypeAsync`](https://github.com/openiddict/openiddict-core/blob/dev/sandbox/OpenIddict.Sandbox.Console.Client/InteractiveService.cs#L353C22-L353C47) doesn't add the default options when there's no results:
image

I've removed the changes to `AmendGrantTypes` and `MapNonStandardResponseParameters`. I've also updated the sandbox's `GetSelectedGrantTypeAsync` to include the `authorization_code` grant if no choices are available.

kevinchalet:
Ah yeah, nice catch, thanks for fixing that! 👏🏻 

I pushed a commit against your branch to fix a few things:
  - I changed the `Classlink` provider name to `ClassLink` to use the official casing.
  - I updated the ID of the Clever and Classlink providers to be lowercase and updated the schema to catch that (sorry, I missed it in the other PR 😅)
  - I replaced the tab that made things look weird in the provider definition.

kevinchalet:
Merged, thanks for your contribution 👍🏻

PR #2044, opened 2024-03-20
Remove `Uri.IsWellFormedOriginalString()`/`Uri.IsWellFormedUriString()`

### LINKED ISSUE #2043: Remove `Uri.IsWellFormedOriginalString()`/`Uri.IsWellFormedUriString()`

Many code paths in OpenIddict use `Uri.IsWellFormedOriginalString()` or `Uri.IsWellFormedUriString()` to check whether a URI is correctly formatted and reject implicit file URIs that are treated as absolute URIs on non-Windows platforms.

Unfortunately, these APIs produce false negatives that prevent returning user codes containing some non-ASCII characters - like emojis - in `verification_uri_complete` even if they are properly percent-encoded. Since https://github.com/openiddict/openiddict-core/pull/2042 will allow using such characters, we'll need to remove these APIs from OpenIddict's codebase.

### Additional context

_No response_

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/2043.

### DIFF
--- a/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
+++ b/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
@@ -419,7 +419,7 @@ public sealed partial class OpenIddictClientWebIntegrationBuilder
                 throw new ArgumentNullException(nameof({{ setting.parameter_name }}));
             }
 
-            if (!{{ setting.parameter_name }}.IsAbsoluteUri || !{{ setting.parameter_name }}.IsWellFormedOriginalString())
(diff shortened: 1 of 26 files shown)


kevinchalet:
Even if it's likely not something I'd recommend (due to the poor usability), here's an example of a device authorization response including a `verification_uri_complete` URI containing percent-encoded emojis in the query string:

```
https://localhost:44395/connect/verify?user_code=%F0%9F%90%B4%F0%9F%8D%A6%20%F0%9F%8F%A0%F0%9F%8F%A0%20%F0%9F%90%B6%F0%9F%8F%80
```

Without this change, such a URI would be rejected, despite being properly encoded. With the `Uri.IsWellFormedOriginalString()`/`Uri.IsWellFormedUriString()` checks removed, it's now properly handled:

image

PR #2063, opened 2024-05-02
Update Clever provider to not require backchannel ID token

anarian:
Hello, from my last PR #2039, while testing different use cases with our application, I came across an issue where they do not always return the ID token even if the application is registered to use OIDC scopes, and confirmed this with their support team.

image

Their application allows districts/teachers to use Clever authentication in two ways:
1. Students/teachers/family can use Clever Library as an authentication method from the external application
2. School Districts can add external applications for their teachers and students to use

Both use OAuth, but only when a School District adds and approves the application, Clever uses OpenID. This causes OpenIddict to return an ID2000 error because of the missing `id_token` if a student/teacher/family uses the Library integration.

image

image

With this update, the id_token isn't required and the request succeeds

image

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
@@ -641,9 +641,10 @@ public static partial class OpenIddictClientWebIntegrationHandlers
              context.RequireBackchannelIdentityToken,
              context.ValidateBackchannelIdentityToken) = context.Registration.ProviderType switch
             {
-                // Clever claims the OpenID Connect flavor of the code flow is supported,
+                // Clever claims the OpenID Connect flavor of the code flow is supported but
                 // their implementation doesn't always return an id_token from the token endpoint.
-                 ProviderTypes.Clever => (true, false, true),
+                ProviderTypes.Clever => (true, false, true),
+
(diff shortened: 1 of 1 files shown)


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -641,6 +641,9 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
              context.RequireBackchannelIdentityToken,
              context.ValidateBackchannelIdentityToken) = context.Registration.ProviderType switch
             {
+                // Clever claims the OpenID Connect flavor of the code flow is supported,
+                // their implementation doesn't always return an id_token from the token endpoint.
+                 ProviderTypes.Clever => (true, false, true),
```
kevinchalet:
```suggestion
                // Clever claims the OpenID Connect flavor of the code flow is supported but
                // their implementation doesn't always return an id_token from the token endpoint.
                ProviderTypes.Clever => (true, false, true),

```

kevinchalet (commented):
> Both use OAuth, but only when a School District adds and approves the application, Clever uses OpenID.

Ouch... their implementation is definitely not that clever 🤣 

Thanks for your PR.

kevinchalet:
Merged, thanks for your contribution.

PR #2071, opened 2024-05-10
Use TimeProvider

### LINKED ISSUE #2069: Use TimeProvider

Hello, I want to replace `DateTimeOffset.UtcNow` with `TimeProvider.GetUtcNow()`. `TimeProvider` will be passed from the service container.

Why?
I am running integration tests in which I use `TimeProvider` to control what `Now` means. Unfortunately, Openiddict does not use it, and it fails this attempt.

Does it sound like something you would like?

___
See https://www.nuget.org/packages/Microsoft.Bcl.TimeProvider/8.0.1#readme-body-tab

### Additional context

_No response_

trejjam:
Fixes #2069


src/OpenIddict.Abstractions/IOptionWithTimeProvider.cs
```diff
@@ -0,0 +1,17 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/openiddict/openiddict-core for more information concerning
+ * the license and the contributors participating to this project.
+ */
+
+namespace OpenIddict.Abstractions;
+
+public interface IOptionWithTimeProvider
```
kevinchalet: That's a nice trick but I'm not sure we should add a public interface just for that. Let's remove it and use `#if SUPPORTS_TIME_PROVIDER ... #else ... #endif` where this interface was called.

src/OpenIddict.Client/OpenIddictClientBuilder.cs
```diff
@@ -192,14 +192,18 @@ public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate()
     /// Registers (and generates if necessary) a user-specific development encryption certificate.
     /// </summary>
     /// <param name="subject">The subject name associated with the certificate.</param>
+    /// <param name="notBefore">The NotBefore of the certificate.</param>
     /// <returns>The <see cref="OpenIddictClientBuilder"/> instance.</returns>
-    public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate(X500DistinguishedName subject)
+    public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate(X500DistinguishedName subject, DateTimeOffset? notBefore = null)
```
kevinchalet: It's technically a binary breaking change, so we can't do that in 5.x. Let's revert these changes for now.

src/OpenIddict.Client/OpenIddictClientConfiguration.cs
```diff
@@ -44,6 +44,13 @@ public void PostConfigure(string? name, OpenIddictClientOptions options)
             throw new InvalidOperationException(SR.GetResourceString(SR.ID0075));
         }
 
+#if SUPPORTS_TIME_PROVIDER
+        if (options.TimeProvider is null)
+        {
+            options.TimeProvider = TimeProvider.System;
```
kevinchalet: We'll want to support resolving the instance from DI instead of directly falling back to the static instance. For that, you'll need to obsolete the existing constructor, add a new one that is also taking an `IServiceProvider` and do `options.TimeProvider ??= _provider.GetService<TimeProvider>() ?? TimeProvider.System`.

src/OpenIddict.Client/OpenIddictClientConfiguration.cs
```diff
@@ -245,8 +254,8 @@ public void PostConfigure(string? name, OpenIddictClientOptions options)
             (SecurityKey, SymmetricSecurityKey) => 1,
 
             // If one of the keys is backed by a X.509 certificate, don't prefer it if it's not valid yet.
-            (X509SecurityKey first, SecurityKey)  when first.Certificate.NotBefore  > DateTime.Now => 1,
-            (SecurityKey, X509SecurityKey second) when second.Certificate.NotBefore > DateTime.Now => -1,
+            (X509SecurityKey first, SecurityKey) when first.Certificate.NotBefore > now => 1,
```
kevinchalet: Please revert the code style changes.

src/OpenIddict.Client/OpenIddictClientOptions.cs
```diff
@@ -157,4 +157,11 @@ public sealed class OpenIddictClientOptions
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Advanced)]
     public HashSet<string> ResponseTypes { get; } = new(StringComparer.Ordinal);
+
+#if SUPPORTS_TIME_PROVIDER
+    /// <summary>
+    /// Gets the time provider
```
kevinchalet:
```suggestion
    /// Gets or sets the time provider.
```

src/OpenIddict.Client/OpenIddictClientOptions.cs
```diff
@@ -157,4 +157,11 @@ public sealed class OpenIddictClientOptions
...
+    /// </summary>
+    public TimeProvider? TimeProvider { get; set; }
```
kevinchalet, line 165: Since a default value is always registered by the post-configuration initializer, it's better to remove the `?` here (otherwise, you have to use the null forgiving operator everywhere 😄)

```suggestion
    public TimeProvider TimeProvider { get; set; } = default!;
```

src/OpenIddict.Core/OpenIddictCoreConfiguration.cs
```diff
@@ -0,0 +1,22 @@
...
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/openiddict/openiddict-core for more information concerning
+ * the license and the contributors participating to this project.
+ */
+
+using Microsoft.Extensions.Options;
+
+namespace OpenIddict.Core;
+
+public class OpenIddictCoreConfiguration : IPostConfigureOptions<OpenIddictCoreOptions>
```
kevinchalet, line 15: Please add the XML docs when you're done 👍🏻

src/OpenIddict.Core/OpenIddictCoreConfiguration.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public void PostConfigure(string? name, OpenIddictCoreOptions options)
+    {
+#if SUPPORTS_TIME_PROVIDER
+        if (options.TimeProvider is null)
+        {
+            options.TimeProvider = TimeProvider.System;
```
kevinchalet: Same remark as earlier.

src/OpenIddict.Core/OpenIddictCoreOptions.cs
```diff
@@ -59,4 +59,11 @@ public sealed class OpenIddictCoreOptions
     /// This property is not used when <see cref="DisableEntityCaching"/> is <see langword="true"/>.
     /// </summary>
     public int EntityCacheLimit { get; set; } = 250;
+
+#if SUPPORTS_TIME_PROVIDER
+    /// <summary>
+    /// Gets the time provider
+    /// </summary>
+    public TimeProvider? TimeProvider { get; set; }
```
kevinchalet, line 67: Same.

src/OpenIddict.Server/OpenIddictServerBuilder.cs
```diff
@@ -201,14 +201,18 @@ public OpenIddictServerBuilder AddDevelopmentEncryptionCertificate()
     /// Registers (and generates if necessary) a user-specific development encryption certificate.
     /// </summary>
     /// <param name="subject">The subject name associated with the certificate.</param>
+    /// <param name="notBefore">The NotBefore of the certificate.</param>
     /// <returns>The <see cref="OpenIddictServerBuilder"/> instance.</returns>
-    public OpenIddictServerBuilder AddDevelopmentEncryptionCertificate(X500DistinguishedName subject)
+    public OpenIddictServerBuilder AddDevelopmentEncryptionCertificate(X500DistinguishedName subject, DateTimeOffset? notBefore = null)
```
kevinchalet: Same.

src/OpenIddict.Server/OpenIddictServerConfiguration.cs
```diff
@@ -400,8 +409,8 @@ public void PostConfigure(string? name, OpenIddictServerOptions options)
             (SecurityKey, SymmetricSecurityKey)          => 1,
 
             // If one of the keys is backed by a X.509 certificate, don't prefer it if it's not valid yet.
-            (X509SecurityKey first, SecurityKey)  when first.Certificate.NotBefore  > DateTime.Now => 1,
-            (SecurityKey, X509SecurityKey second) when second.Certificate.NotBefore > DateTime.Now => -1,
+            (X509SecurityKey first, SecurityKey) when first.Certificate.NotBefore > now => 1,
```
kevinchalet: Revert the code style changes, please.

src/OpenIddict.Server/OpenIddictServerHandlers.Protection.cs
```diff
@@ -1404,7 +1404,7 @@ public ValueTask HandleAsync(GenerateTokenContext context)
                 // entry to the token) that can be used by the resource servers to determine
                 // whether an access token has already been used or blacklist them if necessary.
                 //
-                // Note: scopes are deliberately formatted as a single space-separated 
+                // Note: scopes are deliberately formatted as a single space-separated
```
kevinchalet, line 1415: 👍🏻

src/OpenIddict.Server/OpenIddictServerHandlers.Protection.cs
```diff
@@ -1421,7 +1421,7 @@ public ValueTask HandleAsync(GenerateTokenContext context)
 
                 // For authorization/device/user codes and refresh tokens,
                 // attach claims destinations to the JWT claims collection.
-                if (context.TokenType is TokenTypeHints.AuthorizationCode or TokenTypeHints.DeviceCode or 
+                if (context.TokenType is TokenTypeHints.AuthorizationCode or TokenTypeHints.DeviceCode or
```
kevinchalet, line 1432: 👍🏻

src/OpenIddict.Server/OpenIddictServerHandlers.cs
```diff
@@ -104,7 +104,7 @@ public static partial class OpenIddictServerHandlers
          */
         ValidateSignOutDemand.Descriptor,
         AttachCustomSignOutParameters.Descriptor,
-        
+
```
kevinchalet, line 107: 👍🏻

src/OpenIddict.Validation/OpenIddictValidationConfiguration.cs
```diff
@@ -158,8 +167,8 @@ public void PostConfigure(string? name, OpenIddictValidationOptions options)
             (SecurityKey, SymmetricSecurityKey) => 1,
 
             // If one of the keys is backed by a X.509 certificate, don't prefer it if it's not valid yet.
-            (X509SecurityKey first, SecurityKey)  when first.Certificate.NotBefore  > DateTime.Now => 1,
-            (SecurityKey, X509SecurityKey second) when second.Certificate.NotBefore > DateTime.Now => -1,
+            (X509SecurityKey first, SecurityKey) when first.Certificate.NotBefore > now => 1,
```
kevinchalet: Same.

test/OpenIddict.Server.IntegrationTests/OpenIddictServerIntegrationTests.Exchange.cs
```diff
@@ -450,7 +450,7 @@ public async Task ValidateTokenRequest_ExpiredAuthorizationCodeCausesAnError()
 
                     context.Principal = new ClaimsPrincipal(new ClaimsIdentity("Bearer"))
                         .SetTokenType(TokenTypeHints.AuthorizationCode)
-                        .SetExpirationDate(DateTimeOffset.UtcNow - TimeSpan.FromDays(1))
+                        .SetExpirationDate(context.Options.GetUtcNow() - TimeSpan.FromDays(1))
```
kevinchalet: No need to be super clever in the unit tests, using `DateTimeOffset.UtcNow` should be fine.

src/OpenIddict.Validation/OpenIddictValidationOptions.cs
```diff
@@ -181,4 +181,11 @@ public sealed class OpenIddictValidationOptions
         ValidateAudience = false,
         ValidateLifetime = false
     };
+
+#if SUPPORTS_TIME_PROVIDER
+    /// <summary>
+    /// Gets the time provider
```
kevinchalet:
```suggestion
    /// Gets or sets the time provider.
```

src/OpenIddict.Validation/OpenIddictValidationOptions.cs
```diff
@@ -181,4 +181,11 @@ public sealed class OpenIddictValidationOptions
...
+    /// </summary>
+    public TimeProvider? TimeProvider { get; set; }
```
kevinchalet, line 189:
```suggestion
    public TimeProvider TimeProvider { get; set; } = default!;
```

src/OpenIddict.Quartz/OpenIddictQuartzOptions.cs
```diff
@@ -38,4 +38,11 @@ public sealed class OpenIddictQuartzOptions
     /// By default, this value is set to 14 days and cannot be less than 10 minutes.
     /// </summary>
     public TimeSpan MinimumTokenLifespan { get; set; } = TimeSpan.FromDays(14);
+
+#if SUPPORTS_TIME_PROVIDER
+    /// <summary>
+    /// Gets the time provider
+    /// </summary>
+    public TimeProvider? TimeProvider { get; set; }
```
kevinchalet, line 46: Same.

src/OpenIddict.Server/OpenIddictServerOptions.cs
```diff
@@ -453,4 +453,11 @@ public sealed class OpenIddictServerOptions
     /// that provides additional protection against token leakage.
     /// </summary>
     public bool UseReferenceRefreshTokens { get; set; }
+
+#if SUPPORTS_TIME_PROVIDER
+    /// <summary>
+    /// Gets the time provider
+    /// </summary>
+    public TimeProvider? TimeProvider { get; set; }
```
kevinchalet, line 461: Same.

kevinchalet (commented):
Thanks for your PR!

src/OpenIddict.Client/OpenIddictClientOptions.cs
kevinchalet: Or maybe we should leave it nullable and do `TimeProvider?.GetUtcNow() ?? DateTimeOffset.UtcNow` where it's used? It would address the issue you had in the unit tests.

src/OpenIddict.Client/OpenIddictClientBuilder.cs
trejjam: Can it be an overload?

```cs
public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate(
    X500DistinguishedName subject
) => AddDevelopmentEncryptionCertificate(subject, DateTimeOffset.UtcNow);

public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate(X500DistinguishedName subject, DateTimeOffset notBefore)
{
...
}
```

kevinchalet: Could be, but it honestly looks very weird.

Another approach: move the entire method body inside the `return Configure(options => ...)` delegate. From there, you'd be able to access the `TimeProvider` via `options.TimeProvider`, tho' that only works if the provider was set before the inline `IOptionsConfigure<T>` registered by this extension kicks in. Could be a reasonable compromise.

sandbox/OpenIddict.Sandbox.AspNet.Server/Controllers/AuthorizationController.cs
```diff
@@ -57,7 +62,7 @@ public async Task<ActionResult> Authorize()
             // If the user principal can't be extracted or the cookie is too old, redirect the user to the login page.
             var result = await context.Authentication.AuthenticateAsync(DefaultAuthenticationTypes.ApplicationCookie);
             if (result?.Identity == null || (request.MaxAge != null && result.Properties?.IssuedUtc != null &&
-                DateTimeOffset.UtcNow - result.Properties.IssuedUtc > TimeSpan.FromSeconds(request.MaxAge.Value)))
+                _openIddictCoreOptions.Value.GetUtcNow() - result.Properties.IssuedUtc > TimeSpan.FromSeconds(request.MaxAge.Value)))
```
kevinchalet: Let's keep `DateTimeOffset.UtcNow` in the samples, there's no need for extensibility there.

src/OpenIddict.Client/OpenIddictClientBuilder.cs
```diff
@@ -193,13 +193,24 @@ public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate()
...
+    /// <param name="notBefore">The NotBefore of the certificate.</param>
+    /// <returns>The <see cref="OpenIddictClientBuilder"/> instance.</returns>
+    public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate(X500DistinguishedName subject, DateTimeOffset notBefore)
     {
         if (subject is null)
         {
             throw new ArgumentNullException(nameof(subject));
         }
 
+        var notBeforeDateTime = notBefore.DateTime;
```
kevinchalet: `X509Certificate2.NotBefore` and `X509Certificate2.NotAfter` always return a "local time" value while `DateTimeOffset.DateTime` is always unspecified, so it's very likely comparisons won't do what we expect. We need a `DateTimeKind.Local` value here, which is why `DateTime.Now` is currently used instead of `DateTime.UtcNow` (IIRC, `TimeProvider` has a method that returns the "local time", that you could likely convert to `DateTime` using `DateTimeOffset.LocalDateTime`).

kevinchalet: Reference: https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.notbefore?view=net-8.0

src/OpenIddict.Client/OpenIddictClientBuilder.cs
trejjam: I moved it inside the Configure method, as you suggested. The issue is that we resolve `TimeProvider` in the `IPostConfigureOptions`, therefore it will never be populated before this `Configure` is executed.

To handle this I used this API:
```cs
Services.AddOptions<OpenIddictClientOptions>().Configure<IServiceProvider>((options, serviceProvider) => {
...
});
```

I used the `DateTimeOffsetUtcNow.LocalDateTime`. It should have a correct `Kind`:
https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUH1TATgAoAVASwFsBTABTAHsA3CuGsAOkM4HEbgAGUYBjAIYAbAHKMA7iQCUnYeIkARMcBqVanANIUocBQG4gA==

src/OpenIddict.Client/OpenIddictClientBuilder.cs
trejjam: Thank you for clarifying that. I will continue with this in the previous comment:
https://github.com/openiddict/openiddict-core/pull/2071#discussion_r1596864593

src/OpenIddict.Client/OpenIddictClientBuilder.cs
```diff
@@ -200,67 +200,84 @@ public OpenIddictClientBuilder AddDevelopmentEncryptionCertificate(X500Distingui
...
-            .ToList();
-
-        if (!certificates.Exists(static certificate => certificate.NotBefore < DateTime.Now && certificate.NotAfter > DateTime.Now))
+        Services.AddOptions<OpenIddictClientOptions>().Configure<IServiceProvider>((options, serviceProvider) =>
         {
-#if SUPPORTS_CERTIFICATE_GENERATION
-            using var algorithm = OpenIddictHelpers.CreateRsaKey(size: 2048);
+#if SUPPORTS_TIME_PROVIDER
+            var timeProvider = options.TimeProvider ?? serviceProvider.GetService<TimeProvider>();
+            var notBefore = timeProvider?.GetLocalNow() ?? DateTimeOffset.Now;
```
kevinchalet: Since this variable is used for both `NotAfter` and not `NotBefore` checks, maybe a less specific name - like `now`, that you used in other files - would be better?

src/OpenIddict.Client/OpenIddictClientConfiguration.cs
```diff
@@ -23,14 +24,27 @@ namespace OpenIddict.Client;
 public sealed class OpenIddictClientConfiguration : IPostConfigureOptions<OpenIddictClientOptions>
 {
     private readonly OpenIddictClientService _service;
+    private readonly IServiceProvider? _serviceProvider;
```
kevinchalet: Let's make this one non-nullable and update the existing constructor to throw (it's really meant for DI, no nobody should be using that 😄)

src/OpenIddict.Client/OpenIddictClientConfiguration.cs
```diff
@@ -212,9 +233,17 @@ public void PostConfigure(string? name, OpenIddictClientOptions options)
         // Sort the handlers collection using the order associated with each handler.
         options.Handlers.Sort((left, right) => left.Order.CompareTo(right.Order));
 
+        var now = (
+#if SUPPORTS_TIME_PROVIDER
+                options.TimeProvider?.GetUtcNow() ??
```
kevinchalet: We likely need a local date here too.

src/OpenIddict.Client/OpenIddictClientHandlers.cs
```diff
@@ -1494,8 +1494,11 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
             context.FrontchannelAccessTokenExpirationDate = context.EndpointType switch
             {
                 OpenIddictClientEndpointType.Redirection when context.ExtractFrontchannelAccessToken
-                    => ((long?) context.Request[Parameters.ExpiresIn]) is long value ?
-                        DateTimeOffset.UtcNow.AddSeconds(value) : null,
+                    => (long?)context.Request[Parameters.ExpiresIn] is long value? (
```
kevinchalet:
```suggestion
                    => (long?) context.Request[Parameters.ExpiresIn] is long value ? (
```

src/OpenIddict.Core/OpenIddictCoreConfiguration.cs
```diff
@@ -0,0 +1,35 @@
...
+    private readonly IServiceProvider _serviceProvider;
+
+    /// <summary>
+    /// Creates a new instance of the <see cref="OpenIddictCoreConfiguration"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">The ServiceProvider.</param>
+    public OpenIddictCoreConfiguration(IServiceProvider serviceProvider)
+        => _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+
+    public void PostConfigure(string? name, OpenIddictCoreOptions options)
```
kevinchalet, line 27:
```suggestion
    /// <inheritdoc/>
    public void PostConfigure(string? name, OpenIddictCoreOptions options)
```

src/OpenIddict.Core/OpenIddictCoreConfiguration.cs
```diff
@@ -0,0 +1,35 @@
...
+/// Contains the methods required to ensure that the OpenIddict core configuration is valid.
+/// </summary>
+public class OpenIddictCoreConfiguration : IPostConfigureOptions<OpenIddictCoreOptions>
+{
+    private readonly IServiceProvider _serviceProvider;
+
+    /// <summary>
+    /// Creates a new instance of the <see cref="OpenIddictCoreConfiguration"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">The ServiceProvider.</param>
```
kevinchalet:
```suggestion
    /// <param name="serviceProvider">The service provider.</param>
```

src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
```diff
@@ -13,8 +14,18 @@ namespace OpenIddict.Quartz;
...
-public sealed class OpenIddictQuartzConfiguration : IConfigureOptions<QuartzOptions>
+public sealed class OpenIddictQuartzConfiguration : IConfigureOptions<QuartzOptions>,
+    IPostConfigureOptions<OpenIddictQuartzOptions>
 {
+    private readonly IServiceProvider _serviceProvider;
+
+    /// <summary>
+    /// Creates a new instance of the <see cref="OpenIddictQuartzConfiguration"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">The ServiceProvider.</param>
```
kevinchalet:
```suggestion
    /// <param name="serviceProvider">The service provider.</param>
```

src/OpenIddict.Quartz/OpenIddictQuartzConfiguration.cs
```diff
@@ -42,4 +53,14 @@ public void Configure(QuartzOptions options)
                    .StartAt(DateBuilder.FutureDate(new Random().Next(1, 10), IntervalUnit.Minute));
         });
     }
+
+    public void PostConfigure(string? name, OpenIddictQuartzOptions options)
```
kevinchalet, line 63:
```suggestion
    /// <inheritdoc/>
    public void PostConfigure(string? name, OpenIddictQuartzOptions options)
```

src/OpenIddict.Server/OpenIddictServerConfiguration.cs
```diff
@@ -21,6 +22,15 @@ namespace OpenIddict.Server;
 [EditorBrowsable(EditorBrowsableState.Advanced)]
 public sealed class OpenIddictServerConfiguration : IPostConfigureOptions<OpenIddictServerOptions>
 {
+    private readonly IServiceProvider _serviceProvider;
+
+    /// <summary>
+    /// Creates a new instance of the <see cref="OpenIddictServerConfiguration"/> class.
+    /// </summary>
+    /// <param name="serviceProvider">The ServiceProvider.</param>
```
kevinchalet:
```suggestion
    /// <param name="serviceProvider">The service provider.</param>
```

src/OpenIddict.Server/OpenIddictServerConfiguration.cs
```diff
@@ -183,16 +200,24 @@ public void PostConfigure(string? name, OpenIddictServerOptions options)
             throw new InvalidOperationException(SR.GetResourceString(SR.ID0086));
         }
 
+        var now = (
+#if SUPPORTS_TIME_PROVIDER
+                options.TimeProvider?.GetUtcNow() ??
```
kevinchalet, line 208: Likely here too.

src/OpenIddict.Validation/OpenIddictValidationConfiguration.cs
```diff
@@ -17,14 +18,27 @@ namespace OpenIddict.Validation;
...
     /// <param name="service">The validation service.</param>
+    [Obsolete($"Use constructor with the {nameof(IServiceProvider)}", false)]
     public OpenIddictValidationConfiguration(OpenIddictValidationService service)
         => _service = service ?? throw new ArgumentNullException(nameof(service));
 
+    /// <summary>
+    /// Creates a new instance of the <see cref="OpenIddictValidationConfiguration"/> class.
+    /// </summary>
+    /// <param name="service">The validation service.</param>
+    /// <param name="serviceProvider">The ServiceProvider.</param>
```
kevinchalet:
```suggestion
    /// <param name="serviceProvider">The service provider.</param>
```

src/OpenIddict.Validation/OpenIddictValidationConfiguration.cs
```diff
@@ -90,10 +111,18 @@ public void PostConfigure(string? name, OpenIddictValidationOptions options)
             }
         }
 
+        var now = (
+#if SUPPORTS_TIME_PROVIDER
+                options.TimeProvider?.GetUtcNow() ??
```
kevinchalet, line 113: And here too, I guess 😄

kevinchalet (commented):
Looks great, thanks!

src/OpenIddict.Client/OpenIddictClientBuilder.cs
kevinchalet: Great, thanks! It's a very reasonable approach.

trejjam:
To keep things simple, I changed `TimeProvider.GetLocalNow().LocalDateTime` to `TimeProvider.GetUtcNow().LocalDateTime`. It produces the same value. The difference is that `DateTimeOffset` is always in UTC. Also, it's less weird (for me). I expect that `TimeProvider.GetLocalNow().DateTime` will return Kind=Local, which is an incorrect assumption.

kevinchalet:
Looks great! Thanks a lot for this great contribution ❤️

kevinchalet:
5.6.0 "nightly builds" including your PR are now available on the MyGet feed. Could you please give 'em a try to confirm this does what you wanted? 😃

trejjam:
I would love to, but the issue is that our solution is targeting net7. Migration to net8 is in process but still needs to be done. This PR (without BCL) was a gift for a future me.

I hit the time issue while using BCL package under net7.

PR #2080, opened 2024-05-28
Relax MongoDB.Bson and MongoDB.Driver version requirements

mchwalek:
I wanted to use the mongo integration, but I was unable to - I'm stuck with mongo 3.2 and the version specified in the dependencies is too new. I tested changes contained in this PR locally and it seems to be working ok.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -243,8 +243,8 @@
     <PackageVersion Include="Microsoft.IdentityModel.JsonWebTokens"                           Version="7.5.2"           />
     <PackageVersion Include="Microsoft.IdentityModel.Protocols"                               Version="7.5.2"           />
     <PackageVersion Include="Microsoft.IdentityModel.Tokens"                                  Version="7.5.2"           />
-    <PackageVersion Include="MongoDB.Bson"                                                    Version="2.20.0"          />
-    <PackageVersion Include="MongoDB.Driver"                                                  Version="2.20.0"          />
+    <PackageVersion Include="MongoDB.Bson"                                                    Version="2.13.3"          />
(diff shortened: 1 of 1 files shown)


kevinchalet:
Thanks for your PR.

To be honest, I'm not super comfortable merging it for two reasons:
  - MongoDB 3.2 is old. Very old. So old it's no longer supported for almost 6 years now 😅 
  - The version you used is flagged as vulnerable (see https://github.com/advisories/GHSA-7j9m-j397-g4wx). We deliberately reference an old version of the MongoDB driver for .NET Framework TFMs (as the newer versions reference inconsistent `System.*` dependencies), but it's not really something we should be doing for the .NET 6.0+ TFMs.

mchwalek:
Yes, I'm aware that I'm using a severely outdated version, but there's nothing I can do about it for now. My other alternative is to reimplement all the storage classes, which seems pointless, as it most likely will resort to just copying your source one to one.

My change will still allow to use the most recent driver, that's compatible with newest mongo and has no vulnerabilities

kevinchalet:
> My change will still allow to use the most recent driver, that's compatible with newest mongo and has no vulnerabilities

But unless users explicitly reference a newer driver (or have a dependency on a package that references a newer version), they'll get the vulnerable version by default, which isn't ideal at all and something I really want to avoid for "modern" TFMs.

> My other alternative is to reimplement all the storage classes, which seems pointless, as it most likely will resort to just copying your source one to one.

Copying the MongoDB stores in your own code for the time being - i.e until you have a chance to bump the MongoDB server version - sounds like a better option to me than downgrading the MongoDB dependency of the OpenIddict package, which would affect every other user.

I'm not against revising this decision if there's a stronger demand but for now, I prefer closing this PR.

Issue #1801, opened 2023-06-18
Port the remaining aspnet-contrib providers


kevinchalet:
### Confirm you've already contributed to this project or that you sponsor it

- [X] I confirm I'm a sponsor or a contributor

### Describe the solution you'd like

Here's the complete list of providers currently supported by aspnet-contrib (note: providers supported only by OpenIddict - like Cognito or ADFS - are deliberately not listed here):

| Provider name                | Still operating?   | Supported by OpenIddict? |
|------------------------------|--------------------|--------------------------|
| AdobeIO                      | :white_check_mark: | :white_check_mark:       |
| Alipay                       | :white_check_mark: | :x:                      |
| Amazon                       | :white_check_mark: | :white_check_mark:       |
| amoCRM                       | :white_check_mark: | :x:                      |
| Apple                        | :white_check_mark: | :white_check_mark:       |
| ArcGIS                       | :white_check_mark: | :white_check_mark:       |
| Asana                        | :white_check_mark: | :white_check_mark:       |
| Autodesk                     | :white_check_mark: | :white_check_mark:       |
| Baidu                        | :white_check_mark: | :x:                      |
| Basecamp                     | :white_check_mark: | :white_check_mark:       |
| BattleNet                    | :white_check_mark: | :white_check_mark:       |
| Beam (Mixer)                 | :x:                | :x:                      |
| Bitbucket                    | :white_check_mark: | :white_check_mark:       |
| Buffer                       | :warning:          | :x:                      |
| CiscoSpark (Webex Teams)     | :white_check_mark: | :white_check_mark:       |
| Coinbase                     | :white_check_mark: | :x:                      |
| DeviantArt                   | :white_check_mark: | :white_check_mark:       |
| Deezer                       | :white_check_mark: | :white_check_mark:       |
| DigitalOcean                 | :warning:          | :x:                      |
| Discord                      | :white_check_mark: | :white_check_mark:       |
| Dropbox                      | :white_check_mark: | :white_check_mark:       |
| eBay                         | :white_check_mark: | :x:                      |
| EVE Online                   | :warning:          | :white_check_mark:       |
| ExactOnline                  | :white_check_mark: | :white_check_mark:       |
| Feishu                       | :white_check_mark: | :white_check_mark:       |
| Fitbit                       | :white_check_mark: | :white_check_mark:       |
| Foursquare                   | :warning:          | :x:                      |
| Gitee                        | :white_check_mark: | :white_check_mark:       |
[...]

PR #2092, opened 2024-06-14
Add Gitee to the list of supported providers

gehongyan:
This pull request would like to add Gitee to the list of supported providers, which is also supported by [AspNet.Security.OAuth.Providers] - [Gitee].

[AspNet.Security.OAuth.Providers]: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/dev
[Gitee]: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/docs/gitee.md

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
@@ -1260,14 +1260,14 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                 ProviderTypes.Atlassian => (string?) context.UserinfoResponse?["account_id"],
 
                 // These providers return the user identifier as a custom "id" node:
-                ProviderTypes.Basecamp or ProviderTypes.Box           or ProviderTypes.Dailymotion or
-                ProviderTypes.Deezer   or ProviderTypes.Discord       or ProviderTypes.Disqus      or
-                ProviderTypes.Facebook or ProviderTypes.GitHub        or ProviderTypes.Harvest     or
-                ProviderTypes.Kook     or ProviderTypes.Kroger        or ProviderTypes.Lichess     or
-                ProviderTypes.Mastodon or ProviderTypes.Meetup        or ProviderTypes.Nextcloud   or
-                ProviderTypes.Patreon  or ProviderTypes.Reddit        or ProviderTypes.Smartsheet  or
-                ProviderTypes.Spotify  or ProviderTypes.SubscribeStar or ProviderTypes.Todoist     or
-                ProviderTypes.Twitter  or ProviderTypes.Zoom
+                ProviderTypes.Basecamp   or ProviderTypes.Box      or ProviderTypes.Dailymotion   or
+                ProviderTypes.Deezer     or ProviderTypes.Discord  or ProviderTypes.Disqus        or
+                ProviderTypes.Facebook   or ProviderTypes.Gitee    or ProviderTypes.GitHub        or
+                ProviderTypes.Harvest    or ProviderTypes.Kook     or ProviderTypes.Kroger        or
+                ProviderTypes.Lichess    or ProviderTypes.Mastodon or ProviderTypes.Meetup        or
+                ProviderTypes.Nextcloud  or ProviderTypes.Patreon  or ProviderTypes.Reddit        or
+                ProviderTypes.Smartsheet or ProviderTypes.Spotify  or ProviderTypes.SubscribeStar or
+                ProviderTypes.Todoist    or ProviderTypes.Twitter  or ProviderTypes.Zoom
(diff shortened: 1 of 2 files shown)


gehongyan:
`authorization_code` and `refresh_token` authentication flows work.

image
image

---

Docs:

Gitee OAuth: <https://gitee.com/api/v8/oauth_doc#/>
Get user info: <https://gitee.com/api/v5/swagger#/getV5User>
Get user emails: <https://gitee.com/api/v5/swagger#/getV5Emails>
Allowed scopes: (*Translated by Google*)
image


---

**Question 1:**

In Gitee OAuth, the email address of an authorized user is provided through a separate API. In the ticket creation process of `aspnet-contrib` (See [here](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/4685e27bac85dc83f81a29a08ffc373bdde9c863/src/AspNet.Security.OAuth.Gitee/GiteeAuthenticationHandler.cs#L49-L61)), another HTTP request is made to retrieve it. How can this be achieved in `OpenIddict`?

**Question 2:**

Gitee supports the `password` authentication flow. However, it requires that the scope in the request body must match the values just like those used in the `authorization_code` flow, such as `user_info emails`. It also disallows unlisted values like `offline_access`. During testing, I found that the scope in the request for the `password` mode is fixed as `offline_access`, and it cannot be configured via the `Provider.Environment.Scope` XML node or by using `options.AddScopes()`. This might be a non-standard process, but how can we configure or write code to meet Gitee's requirements?

kevinchalet:
Thanks for your PR!

> **Question 1:**
> 
> In Gitee OAuth, the email address of an authorized user is provided through a separate API. In the ticket creation process of `aspnet-contrib` (See [here](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/4685e27bac85dc83f81a29a08ffc373bdde9c863/src/AspNet.Security.OAuth.Gitee/GiteeAuthenticationHandler.cs#L49-L61)), another HTTP request is made to retrieve it. How can this be achieved in `OpenIddict`?

OpenIddict (deliberately) only supports a single userinfo request and retrieving additional information (e.g the email or postal addresses of users) is something that must be done in your own code (for instance using the official SDK/REST API package provided by the vendor if one is available, or using a `HttpClient`).

> **Question 2:**
> 
> Gitee supports the `password` authentication flow. However, it requires that the scope in the request body must match the values just like those used in the `authorization_code` flow, such as `user_info emails`. It also disallows unlisted values like `offline_access`. During testing, I found that the scope in the request for the `password` mode is fixed as `offline_access`, and it cannot be configured via the `Provider.Environment.Scope` XML node or by using `options.AddScopes()`. This might be a non-standard process, but how can we configure or write code to meet Gitee's requirements?

That scope is set here: https://github.com/openiddict/openiddict-core/blob/7cf4c28e40d34d98cd8d9377db92f71ef480d832/sandbox/OpenIddict.Sandbox.Console.Client/InteractiveService.cs#L240-L247

You should be able to remove/update that line to make the provider work (note that if you remove that line, no scope will be sent: the scopes attached to `OpenIddictClientRegistration.Scopes` or defined in the XML only apply to user-interactive/browser-based flows like the authorization code flow).

gehongyan:
Thanks for such a quick response!

---

> ... retrieving additional information is something that must be done in your own code ...

Got it. So I will ignore the emails.

> You should be able to remove/update that line to make the provider work ...

Oh, thank you. I didn't notice that the `scopes` in the console sandbox are set from the user code. I'm not very familiar with OpenIddict. To make Gitee's password mode work in the sandbox code, should I add a prompt asking users if they want to input custom scope values to override the default ones? Or should I just ensure it works correctly in my local tests without changing the sandbox code?

kevinchalet:
> Got it. So I will ignore the emails.

👍🏻 

> Oh, thank you. I didn't notice that the `scopes` in the console sandbox are set from the user code. I'm not very familiar with OpenIddict.

No worries: the OpenIddict client and its web providers support a lot more features than what we had in the aspnet-contrib providers, so there's definitely a learning curve 😄 

> Or should I just ensure it works correctly in my local tests without changing the sandbox code?

That's the easiest option 👍🏻

gehongyan:
Now, `password` mode and refreshing work.

image

kevinchalet:
Awesome! (thanks for taking the time to post these screenshots, it's very useful to see what's returned 👍🏻)

gehongyan:
`password` mode and its refreshing now work.

image

Reposted the screenshot with hiding potential personal info leeks.

kevinchalet:
Merged. Thanks for your contribution! 👏🏻

kevinchalet:
Note: I updated https://github.com/openiddict/openiddict-core/issues/1801 to indicate that Gitee and Kook are now supported by OpenIddict 😃

PR #2098, opened 2024-06-18
Add Weibo to the list of supported providers

gehongyan:
This pull request would like to add Weibo to the list of supported providers, which is also supported by [AspNet.Security.OAuth.Providers](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/dev) - [Weibo](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/docs/weibo.md).


gehongyan:
### Docs

The Chinese and English versions of some documents are not completely identical, but the differences are minor. Therefore, I have included both versions here.

- Authorization mechanism: [Chinese](https://open.weibo.com/wiki/%E6%8E%88%E6%9D%83%E6%9C%BA%E5%88%B6) (_Chinese only_)
- API that requests the code: [Chinese](https://open.weibo.com/wiki/Oauth2/authorize), [English](https://open.weibo.com/wiki/Oauth2/authorize/en)
- API that gets the access token: [Chinese](https://open.weibo.com/wiki/Oauth2/access_token), [English](https://open.weibo.com/wiki/Oauth2/access_token/en)
- API that gets the token info: [Chinese](https://open.weibo.com/wiki/Oauth2/get_token_info) (_Chinese only_)
- API that revokes the token: [Chinese](https://open.weibo.com/wiki/Oauth2/revokeoauth2) (_Chinese only_)

### Hint

The [single English version of OAuth docs](https://open.weibo.com/wiki/OAuth/en) seem to be OAuth 1.0. 

The `refresh_token` grant type is only supported in their SDK. Web applications should be reauthorized instead.

image

---

### Useful Screenshots

Authorization page

image

Authorization result

Weibo returns lots of fields from the user info request.

image

image

Token revocation

image

Authorization denials

image

---

### What I Did

The scopes should be joined with the comma `,` instead of the space, so I adjusted the `OpenIddictClientWebIntegrationHandlers.FormatNonStandardScopeParameter`.

The redirect URI supports additional settings, which are appended in `OpenIddictClientWebIntegrationHandlers.AttachAdditionalChallengeParameters`.

The user info request requires the `access_token` as a urencoded parameter instead of the header, so I adjusted it in `OpenIddictClientWebIntegrationHandlers.Userinfo.AttachAccessTokenParameter`.

The user info request still requires the `uid` parameter, so I appended it in `OpenIddictClientWebIntegrationHandlers.AttachAdditionalUserinfoRequestParameters`.

The user identifier is mapped from `id` in `OpenIddictClientWebIntegrationHandlers.MapCustomWebServicesFederationClaims`.

---

### Questions

Weibo provided an API for getting token info, which returns:

image

The negative number value of `expire_in` represents that the token has expired. Is what this API does similar to token introspection? If so, is there a way we can benefit from this API?

kevinchalet:
Thanks for your PR! Looks great!


> Weibo provided an API for getting token info, which returns:
> 
> image
> 
> The negative number value of `expire_in` represents that the token has expired. Is what this API does similar to token introspection? If so, is there a way we can benefit from this API?

In general, OAuth 2.0 introspection implementations return a JSON response containing only `active: false` when the access token expired (instead of returning a real response with a negative or zero `expires_in` node).

We could probably use that endpoint for introspection by adding some mapping, but the claims returned by that endpoint seem quite limited, so not sure it's worth the extra work.

gehongyan:
> We could probably use that endpoint for introspection by adding some mapping, but the claims returned by that endpoint seem quite limited, so not sure it's worth the extra work.

I am not sure where to map the introspection. I did not find any existing mappers in the WebIntegration project. I also didn't find any providers configured with the introspection endpoint in the XML statically. If you prefer to ignore it in case of lack of benefits, then let's just ignore it. If there will be an introspection mapper, then let's implement the Weibo introspection. Which do you prefer at present? 😃

kevinchalet:
> I did not find any existing mappers in the WebIntegration project. I also didn't find any providers configured with the introspection endpoint in the XML statically.

Yeah, introspection is not widely supported 😭 

> If you prefer to ignore it in case of lack of benefits, then let's just ignore it.

Let's ignore it for now and see if there's a demand for that later 👍🏻

gehongyan:
> Let's ignore it for now and see if there's a demand for that later 👍🏻

Sounds good! 👍

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1898,6 +1898,33 @@
...
+      <Configuration AuthorizationEndpoint="https://api.weibo.com/oauth2/authorize"
+                     RevocationEndpoint="https://api.weibo.com/oauth2/revokeoauth2"
+                     TokenEndpoint="https://api.weibo.com/oauth2/access_token"
+                     UserinfoEndpoint="https://api.weibo.com/2/users/show.json" />
+    </Environment>
+
+    <Setting PropertyName="Display" ParameterName="display" Type="String" Required="false"
+             Description="The value used as the 'display' parameter (can be set to 'default', 'mobile', 'wap', 'client', or 'apponweibo' to adjust the authorization page display style)" />
+
+    <Setting PropertyName="ForceLogin" ParameterName="forceLogin" Type="String" Required="false"
```
kevinchalet: Hum, representing a boolean as a string works, but we should probably update the generator and the XML schema to support boolean settings. Do you want to give it a try? 😃

gehongyan: Sounds good!

gehongyan: We are resolving the same issue we should have resolved at Huawei Question 4 😄

kevinchalet: Haha yeah! (tho' it wasn't necessary for Huawei since we didn't use the userinfo endpoint 😄)

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
```diff
@@ -786,6 +788,7 @@ public sealed partial class {{ provider.name }}
                                 "String" => "string",
                                 "StringHashSet" => "HashSet<string>",
                                 "Uri" => "Uri",
+                                "Boolean" => "bool",
```
kevinchalet: Can you please move it so the alphabetical order is respected?

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
```diff
@@ -1368,6 +1371,7 @@ public sealed class {{ provider.name }}
...
```
kevinchalet: Same.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xsd
```diff
@@ -507,6 +507,7 @@
                         <xs:enumeration value="SigningKey" />
                         <xs:enumeration value="String" />
                         <xs:enumeration value="Uri" />
+                        <xs:enumeration value="Boolean" />
```
kevinchalet: Here too, please.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1729,7 +1729,12 @@ public ValueTask HandleAsync(ProcessChallengeContext context)
                 var settings = context.Registration.GetWeiboSettings();
 
                 context.Request["display"] = settings.Display;
-                context.Request["forcelogin"] = settings.ForceLogin;
+                context.Request["forcelogin"] = settings.ForceLogin switch
```
kevinchalet: Did you add the switch because Weibo didn't like the `True` value that is sent by OpenIddict when using a `bool` parameter?

gehongyan: I didn't notice that, but after I tested their behaviors, it seems that they don't recognize `True`.

image

Only when I set it to `true` will the force login works.

image

gehongyan: Will it be better if we use

```cs
context.Request["forcelogin"] = settings.ForceLogin?.ToString().ToLowerInvariant();
```

kevinchalet: > Will it be better if we use

Nah. I wanted to change `OpenIddictParameter` to use the lowercased `true` and `false` values for `bool` conversions, it's a good opportunity to do it 😄 

I'll open a separate a PR and merge it before yours so you'll be able to use `context.Request["forcelogin"] = settings.ForceLogin` directly.

gehongyan: > I wanted to change OpenIddictParameter to use the lowercased true and false values for bool conversions, it's a good opportunity to do it 😄

Will this change be a breaking change possibly breaking other providers which need an exact `True` value? We need to check existing usages.

kevinchalet: It's technically a potentially-breaking behavior change, but there's only one place where we're using it: for FaceIT, and it should work just fine with a lowercase value.

I'll add a note in the 5.7.0 release notes to mention that.

gehongyan:
It seems that adding a `Boolean` needs me to check the entire source generator logic. For example, the latest commit added support for setting the default value set to a bool setting. I am not sure whether there should be more fixtures needed. If more are found, let's create new pull requests to fix them.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
gehongyan: Nice work!

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
```diff
@@ -699,10 +699,12 @@ public sealed partial class {{ provider.name }}
         {{~ end ~}}
         public {{ provider.name }} Set{{ setting.property_name }}({{ setting.clr_type }} {{ setting.parameter_name }})
         {
+            {{~ if setting.clr_type != 'bool' ~}}
```
kevinchalet: Instead of making the `if (...)` block conditional, could you please copy the entire method and put it in its own `{{~ else if setting.clr_type == 'bool' ~}}` / `{{~ end ~}}`?

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
```diff
@@ -906,6 +909,11 @@ public sealed partial class OpenIddictClientWebIntegrationConfiguration
             {
                 settings.{{ setting.property_name }} = new Uri(""{{ setting.default_value }}"", UriKind.RelativeOrAbsolute);
             }
+            {{~ else if setting.type == 'Boolean' ~}}
+            if (!settings.{{ setting.property_name }}.HasValue && bool.TryParse(""{{ setting.default_value }}"", out var value))
+            {
+                settings.{{ setting.property_name }} = value;
+            }
```
kevinchalet: Using `TryParse` could potentially hide a bug if we somehow set a non-true/false value. Using `Parse()` here would immediately throw an exception.

```suggestion
            {{~ else if setting.type == 'Boolean' ~}}
            if (!settings.{{ setting.property_name }}.HasValue)
            {
                settings.{{ setting.property_name }} = bool.Parse(""{{ setting.default_value }}"");
            }
```

kevinchalet:
> It seems that adding a Boolean needs me to check the entire source generator logic. For example, the latest commit added support for setting the default value set to a bool setting. I am not sure whether there should be more fixtures needed. If more are found, let's create new pull requests to fix them.

I took a look and it seems you covered all the cases, except one in this part (responsible for throw an exception if a setting wasn't set):

https://github.com/openiddict/openiddict-core/blob/17fa2cac528f7b1eb78f41e94ee311371c6dd301/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs#L950-L968

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
gehongyan: Hopes I have understood what you meant. I will try it tomorrow. 🌃

gehongyan:
Copy that. Thanks for your hint. I will implememt it. 🍻

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
kevinchalet: Actually, I'm stupid... there's a much better approach 🤣 

```suggestion
            {{~ else if setting.type == 'Boolean' ~}}
            if (settings.{{ setting.property_name }} is null)
            {
                settings.{{ setting.property_name }} = setting.default_value;
            }
````

File: gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs (hunk 6 above)
Comment (kevinchalet) [reply gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs]: Thanks! Have a great night 🌔

**Comment by gehongyan** [comment]
> I took a look and it seems you covered all the cases, except one in this part (responsible for throw an exception if a setting wasn't set):

It seems that the current generator will also generate proper assertions for bool properties.

```cs
// Generated

if (settings.ForceLogin is null)
{
    throw new InvalidOperationException(SR.FormatID0332(nameof(settings.ForceLogin), Providers.Weibo));
}
```

Do we need some additional checks for `Boolean` just like what was done in the `Uri` condition? Should we still split the bool condition with a different if block such as:

```diff
Index: gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs b/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
--- a/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs	(revision ec411581de26f474c631a8e986183e91b3d87dda)
+++ b/gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs	(date 1718761954304)
@@ -970,6 +970,8 @@
             {{~ if setting.required ~}}
             {{~ if setting.type == 'String' ~}}
             if (string.IsNullOrEmpty(settings.{{ setting.property_name }}))
+            {{~ else if setting.type == 'Boolean' ~}}
+            if (!settings.{{ setting.property_name }}.HasValue)
             {{~ else ~}}
             if (settings.{{ setting.property_name }} is null)
             {{~ end ~}}
             {{~ end ~}}
             {
                 throw new InvalidOperationException(SR.FormatID0332(nameof(settings.{{ setting.property_name }}), Providers.{{ provider.name }}));
             }
             {{~ end ~}}

```

kevinchalet:
> Do we need some additional checks for `Boolean` just like what was done in the `Uri` condition? Should we still split the bool condition with a different if block such as:

Ah, my bad! Yeah, the existing condition works fine for booleans so no need for a specialized check 👍🏻

gen/OpenIddict.Client.WebIntegration.Generators/OpenIddictClientWebIntegrationGenerator.cs
```diff
@@ -688,6 +688,19 @@ public sealed partial class {{ provider.name }}
...
+        /// </summary>
+        /// <param name=""{{ setting.parameter_name }}"">{{ setting.description | string.capitalize }}.</param>
+        /// <returns>The <see cref=""OpenIddictClientWebIntegrationBuilder.{{ provider.name }}""/> instance.</returns>
+        {{~ if setting.obsolete ~}}
+        [Obsolete(""This option is no longer supported and will be removed in a future version."")]
+        {{~ end ~}}
+        public {{ provider.name }} Set{{ setting.property_name }}(bool {{ setting.parameter_name }})
+        {
+            return Set(registration => registration.Get{{ provider.name }}Settings().{{ setting.property_name }} = {{ setting.parameter_name }});
+        }
```
kevinchalet:
```suggestion
        {{~ else if setting.clr_type == 'bool' ~}}
        /// <summary>
        /// Configures {{ setting.description }}.
        /// </summary>
        /// <param name=""{{ setting.parameter_name }}"">{{ setting.description | string.capitalize }}.</param>
        /// <returns>The <see cref=""OpenIddictClientWebIntegrationBuilder.{{ provider.name }}""/> instance.</returns>
        {{~ if setting.obsolete ~}}
        [Obsolete(""This option is no longer supported and will be removed in a future version."")]
        {{~ end ~}}
        public {{ provider.name }} Set{{ setting.property_name }}(bool {{ setting.parameter_name }})
            => Set(registration => registration.Get{{ provider.name }}Settings().{{ setting.property_name }} = {{ setting.parameter_name }});
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1938,6 +1938,33 @@
...
+      <Configuration AuthorizationEndpoint="https://api.weibo.com/oauth2/authorize"
+                     RevocationEndpoint="https://api.weibo.com/oauth2/revokeoauth2"
+                     TokenEndpoint="https://api.weibo.com/oauth2/access_token"
+                     UserinfoEndpoint="https://api.weibo.com/2/users/show.json" />
+    </Environment>
+
+    <Setting PropertyName="Display" ParameterName="display" Type="String" Required="false"
+             Description="The value used as the 'display' parameter (can be set to 'default', 'mobile', 'wap', 'client', or 'apponweibo' to adjust the authorization page display style)" />
+
+    <Setting PropertyName="ForceLogin" ParameterName="forceLogin" Type="Boolean" Required="false"
```
```suggestion
    <Setting PropertyName="ForceLogin" ParameterName="force" Type="Boolean" Required="false"
```

kevinchalet:
Looks great! If you don't have anything else to add, let's merge it 😃

gehongyan:
Thanks for your help! Nothing more is planned at present. ❤️

kevinchalet:
Merged. Thanks a lot for your contribution! ❤️

kevinchalet:
Hey 👋🏻 

FYI, OpenIddict 5.7.0 just shipped and includes the 5 providers you contributed recently: https://github.com/openiddict/openiddict-core/releases/tag/5.7.0 🎉 

Cheers!

gehongyan:
Cheers! 🍻

PR #2136, opened 2024-07-08
Add OpenIddict.AmazonDynamoDB to third party list

ganhammar:
Hey!

Thanks for this great package!

I have written an Amazon DynamoDB adapter for OpenIddict and wondered if you would be interested to add it to the list of third party projects? It is fairly battle tested at this point.
I added it to the end of the list, but, it seems like the list might be alphabetically sorted, should I fixed that?

I also have a [blog post](https://www.ganhammar.se/posts/openiddict-on-aws-serverless-flexible-oauth2-oidc-provider) where I walkthrough how to use it that could be added to the list of blog posts, if that would be of interest? 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Instead, **developers are encouraged to execute the conformance tests against th
 
 **Looking for additional resources to help you get started with OpenIddict?** Don't miss these interesting blog posts:
 
+- **[OpenIddict on AWS Serverless: Flexible OAuth2/OIDC Provider](https://www.ganhammar.se/posts/openiddict-on-aws-serverless-flexible-oauth2-oidc-provider)** by [Anton Ganhammar](https://github.com/ganhammar)
 - **[OpenIddict 5.0 general availability](https://kevinchalet.com/2023/12/18/openiddict-5-0-general-availability/)** by [Kévin Chalet](https://github.com/kevinchalet)
 - **[Introducing native applications, per-client token lifetimes and client assertions support in OpenIddict 5.0 preview1](https://kevinchalet.com/2023/10/20/introducing-native-applications-per-client-token-lifetimes-and-client-assertions-support-in-openiddict-5-0-preview1/)** by [Kévin Chalet](https://github.com/kevinchalet)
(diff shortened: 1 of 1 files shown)


kevinchalet:
Hey Anton!

> Thanks for this great package!

Thanks! ❤️ 

> I have written an Amazon DynamoDB adapter for OpenIddict and wondered if you would be interested to add it to the list of third party projects? It is fairly battle tested at this point.

Awesome! 👏🏻 

> I added it to the end of the list, but, it seems like the list might be alphabetically sorted, should I fixed that?

Yes please 👍🏻 

> I also have a [blog post](https://www.ganhammar.se/posts/openiddict-on-aws-serverless-flexible-oauth2-oidc-provider) where I walkthrough how to use it that could be added to the list of blog posts, if that would be of interest? 🙂

Sure, please add it to the list (at the top since it's a very recent post).

I see you opted for `Results.BadRequest()` to return an error to the caller: it works but it's not the recommended option as it doesn't allow re-entering the OpenIddict pipeline to process that error (which means that you can't customize it and that a 400 response instead of a 401 response - more appropriate for `invalid_client` - will be returned).

Instead, you may want to use `Results.Challenge()` or `Results.Forbidden()` (both do the same thing in OpenIddict):

```csharp
return Results.Challenge(
    authenticationSchemes: [OpenIddictServerAspNetCoreDefaults.AuthenticationScheme],
    properties: new AuthenticationProperties(new Dictionary<string, string>
    {
        [OpenIddictServerAspNetCoreConstants.Properties.Error] = Errors.InvalidClient,
        [OpenIddictServerAspNetCoreConstants.Properties.ErrorDescription] = "The specified hardcoded identity is invalid."
    }));
```

ganhammar:
Thanks for the quick response!

I have updated the sorting of the third party list, added the blog post to the list, and updated the post (and reference project) according to your feedback! Thanks! 👍

kevinchalet:
Great! While we're at it, could you also please update the docs to add your DynamoDB package? https://github.com/openiddict/openiddict-documentation/blob/dev/integrations/index.md

PR #2174, opened 2024-09-03
Rename the device, cryptography, logout and verification endpoints and the associated events/event handlers

kevinchalet:
This PR materializes something I wanted to do for a long time: renaming some of the endpoints to be more specific or match the standard names (e.g the `device authorization endpoint` was initially named `device endpoint` in the first drafts, which is the name currently used by OpenIddict).

Here's the list of endpoints impacted by this change:
  - `Cryptography endpoint` -> `JSON Web Key Set endpoint`.
  - `Device endpoint` -> `Device authorization endpoint`.
  - `Logout endpoint` -> `End-session endpoint`.
  - `Userinfo endpoint` -> `UserInfo endpoint`.
  - `Verification endpoint` -> `End-user verification endpoint`.

All the constants, builder methods, events and event handlers have been updated to use the new names.


kevinchalet:
Note: since it's a massive change, it will need to be mentioned in the 5.x -> 6.x migration guide.

PR #2190, opened 2024-09-26
[Draft] Thread-safe alteration of server credentials during runtime

patrikwlund:
## What we want to achieve
Without intensive changes to the library design, allow for changing signing/encryption keys during runtime. These changes will allow us to safely build key rotation around this server configuration object. We can run our own background service that updates the config when it needs to.

## Changes
- [API Breaking] Make `EncryptionCredentials` and `SigningCredentials` read-only, but settable
- [Behavior Breaking?] New `CreateTokenValidationParameters` method that resolves _current_ credentials, rather than using credentials set during initial startup

## Draft limitations, design feedback
If these suggested changes are relevant to include in the library, I also think we should make the credential validation logic publicly/statically available. So that, in our case, the background service can validate and sort the keys before they are set in the config object. I'll gladly do that as well in this PR if you want it.

## Further considerations
- Should the library perhaps change to a provider-based design, where keys are resolved via some `ISigningKeyProvider` or such?

### DIFF
--- a/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.Authentication.cs
+++ b/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.Authentication.cs
@@ -113,7 +113,7 @@ public static partial class OpenIddictServerAspNetCoreHandlers
                     return;
                 }
 
-                var parameters = context.Options.TokenValidationParameters.Clone();
+                var parameters = context.Options.CreateTokenValidationParameters();
                 parameters.ValidIssuer ??= (context.Options.Issuer ?? context.BaseUri)?.AbsoluteUri;
                 parameters.ValidAudience ??= parameters.ValidIssuer;
                 parameters.ValidTypes = [JsonWebTokenTypes.Private.AuthorizationRequest];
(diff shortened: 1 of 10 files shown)


kevinchalet:
Hey,

Thanks for your PR.

> These changes will allow us to safely build key rotation around this server configuration object. We can run our own background service that updates the config when it needs to.

It's not really how things are supposed to work when using `Microsoft.Extensions.Options` (which is the library OpenIddict uses for all its stacks): instead of trying to mutate the options instance after it's been created, you're expected to reload it entirely and discard the old instance using an option change token source and custom option initializers.

Some resources that you may find useful:
  - https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.ioptionschangetokensource-1?view=net-8.0
  - https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.iconfigureoptions-1?view=net-8.0
  - https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-8.0

Since it's a way more flexible approach - that allows changing all the options and not just the credentials - it's not entirely clear what benefit your approach would have compared to the standard option.

Cheers.

patrikwlund:
> a way more flexible approach - that allows changing all the options and not just the credentials

That does seem better, and the library seems to consistently use `IOptionsMonitor`. I will attempt that angle instead. Thanks!

PR #2194, opened 2024-10-04
Added Avalonia sample app

gentledepp:
I would like to contribute an Avalonia sample app.

It comes with a detailed readme.md which describes
- the approach taken 
- and the errors that happened when developing the sample

**Note:** This is a draft because

1. I want to know if you are interested at all
2. Several things might still need improvement:

  - [ ] unified build props 
        
    I tried to add all dependencies of the sample app to the Directory.Packages.props but this seems to not work due to a bug in the Avalonia tooling:
    image
    Would it be ok to keep it as it is? 😅

  - [ ] The `WebProviders` code generation does not work in Avalonia?
        
 In the Maui sample app, the `MainPage` method

```csharp
    private async void OnLocalLoginWithGitHubButtonClicked(object sender, EventArgs e)
    => await LogInAsync("Local", new()
    {
        [Parameters.IdentityProvider] = Providers.GitHub
    });
```
   passes the constant  `Providers.Github` which is not available in the `MainViewModel` of the Avalonia sample project.
   Did I do something wrong?
         

   - [ ] Obviously I need to rebase on dev (but I initially had problems getting .net 9 preview to work)

# One minor question
And regarding the Desktop (Windows) authentication I do have  question
The login process opens a browser tab in the currently active default browser.
This may be somewhere in the background and invisible to users - potentially causing confusion
Also, when the login process was successful, the browser window is not closed.

Since this is possible on iOS/Android: Can't OpenIddict open a browser as a dialog (on top of the current app) and close the tab when it redirects back to the app? 
Is this due to a shortcoming in the Windows API or is it a safety measure?


kevinchalet:
Hey,

Thanks for your PR.

> I want to know if you are interested at all.

Well, I'm not an Avalonia user myself but it seems the project has a lot of users, so maybe it's not a bad idea. That said, the cost for maintaining these additional sandbox projects must be minimal and given I'm not familiar with Avalonia, it's not clear to me whether maintaining that in the long term will be expensive or not.

> Would it be ok to keep it as it is? 😅

Arf, no: being able to centralize dependencies using NuGet CPVM is essential and dramatically simplifies maintenance (and greatly reduces the risk of having an inconsistent dependencies graph at the solution level).

> Did I do something wrong?

Weird. Did you try to restart VS?

>  Is this due to a shortcoming in the Windows API or is it a safety measure?

It's deliberate:
  - Web views that are not managed by the OS are strongly discouraged by OAuth 2.0 best practices.
  - Having real SSO support is not practical when each app uses its own isolated web view.

PR #2200, opened 2024-10-07
Added MAUI android sample

gentledepp:
I added the android part of the OpenIddict.Sandbox.Maui.Client

What may be missing is 
 - [ ] add a readme to maui describing how to get the sample to run on iOS and Android (i.e. use dev tunnel and do not use developmentcertificates)
 Shall I add that too?

Also, I added a progress dialog with a "cancel" button, as it was confusing to see the buttons to just be disabled.
In case anything goes wrong on the server side you could not cancel the login process, but had to restart the app. 😐

It is rough, but functional. Check it out:

mt-erber_deeplink

### DIFF
--- a/sandbox/OpenIddict.Sandbox.Maui.Client/App.xaml.cs
+++ b/sandbox/OpenIddict.Sandbox.Maui.Client/App.xaml.cs
@@ -1,5 +1,4 @@
-﻿#if IOS || MACCATALYST || WINDOWS
-namespace OpenIddict.Sandbox.Maui.Client;
+﻿namespace OpenIddict.Sandbox.Maui.Client;
 
 public partial class App : Application
 {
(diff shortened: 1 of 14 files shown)


kevinchalet:
Note: not adding Android to that sample was deliberate: the Android tooling is bugged and prevents the project from building correctly on my machine (probably caused by the diacritics in my Windows user folder name). I'm of course not against merging this PR, but the underlying issue will first need to be fixed by the Android folks.

gentledepp:
No problem. I am not in a hurry with this.
Merge it as soon or late as you want.

I used this as an exercise

kevinchalet:
I just gave it another try and the bug is still there, unfortunately. Closing, as I don't think we'll be able to add an Android TFM to that sample in the near future.

Thanks for your PR!

PR #2204, opened 2024-10-10
Allow use of MongoDB 2.28

SimonBartha:
If a project references MongoDB in version 2.28 then openiddict insists on sticking to version 2.20 which results in an unsolvable nuget conflict.

image

Minimal example to reproduce the issue is here:
https://github.com/SimonBartha/OpenIddictNugetExample

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -242,8 +242,8 @@
     <PackageVersion Include="Microsoft.IdentityModel.Protocols"                               Version="7.6.3"  />
     <PackageVersion Include="Microsoft.IdentityModel.Tokens"                                  Version="7.6.3"  />
     <PackageVersion Include="Microsoft.Net.Http.Headers"                                      Version="2.1.14" />
-    <PackageVersion Include="MongoDB.Bson"                                                    Version="2.20.0" />
-    <PackageVersion Include="MongoDB.Driver"                                                  Version="2.20.0" />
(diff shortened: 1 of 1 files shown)


kevinchalet:
Thanks for your PR.

> openiddict insists on sticking to version 2.20 which results in an unsolvable nuget conflict.

Actually, it doesn't (the dependency is not range-constrained by OpenIddict, that only requires 2.20 as the "minimal" package version). The problem is that MongoDB decided to start strong naming their assemblies and introduced that change in a minor version (even if it was a breaking change since it changes the identity of an assembly). That unwise decision is obviously causing a lot of pain in the ecosystem: https://www.mongodb.com/community/forums/t/net-c-driver-strong-naming/291649

Updating OpenIddict to target MongoDB 2.28 is certainly an option, but it's also a breaking change for OpenIddict users that use libraries built using the non-strong-named MongoDB assemblies so it's not clear whether doing that in OpenIddict 6.0 (shipping in December) is the right thing to do.

SimonBartha:
Thank you for the quick reply and the explanation. I tried everything and didnt think of strong naming. Would it be possible to include the bump to 2.28 in a pre release of 6.0. This is our last blocker in updating to .net8.

kevinchalet:
> Would it be possible to include the bump to 2.28 in a pre release of 6.0.

I considered multiple options but wasn't very happy with any of them 😄

We'll directly move to MongoDB 3.0 (that was just released)... it's likely the least terrible option: https://github.com/openiddict/openiddict-core/pull/2206.

PR #2217, opened 2024-11-02
Support for Zendesk

mozts2005:
I have added Support for Zendesk. I am happy to help in testing validation as I have an account with Zendesk that is used for Open-source project testing.

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -407,7 +407,8 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                         throw new InvalidOperationException(SR.FormatID0334("d/results/0"))),
 
                     // Fitbit, Todoist, and Zendesk return a nested "user" object.
-                    ProviderTypes.Fitbit or ProviderTypes.Todoist or ProviderTypes.Zendesk => new(context.Response["user"]?.GetNamedParameters() ??
+                    ProviderTypes.Fitbit or ProviderTypes.Todoist or ProviderTypes.Zendesk
+                        => new(context.Response["user"]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("user"))),
 
                     // Harvest returns a nested "user" object and a collection of "accounts".
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -406,8 +406,8 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                     ProviderTypes.ExactOnline => new(context.Response["d"]?["results"]?[0]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("d/results/0"))),
 
-                    // Fitbit and Todoist return a nested "user" object.
-                    ProviderTypes.Fitbit or ProviderTypes.Todoist => new(context.Response["user"]?.GetNamedParameters() ??
+                    // Fitbit, Todoist, and Zendesk return a nested "user" object.
+                    ProviderTypes.Fitbit or ProviderTypes.Todoist or ProviderTypes.Zendesk => new(context.Response["user"]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("user"))),
 
                     // Harvest returns a nested "user" object and a collection of "accounts".
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -406,7 +406,7 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                     ProviderTypes.ExactOnline => new(context.Response["d"]?["results"]?[0]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("d/results/0"))),
 
-                    // Fitbit, Todoist, and Zendesk return a nested "user" object.
+                    // Fitbit, Todoist and Zendesk return a nested "user" object.
                     ProviderTypes.Fitbit or ProviderTypes.Todoist or ProviderTypes.Zendesk
                         => new(context.Response["user"]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("user"))),
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
@@ -2153,6 +2153,23 @@
     <Environment Issuer="https://api.login.yahoo.com/" />
   </Provider>
 
+  <Provider Name="Zendesk" Id="5577b117-1f3d-464f-b987-aabf7c0ab3f7" Documentation="https://support.zendesk.com/hc/en-us/articles/4408845965210-Using-OAuth-authentication-with-your-application">
+
+    <!--Note: Zendesk is a multitenant provider that relies on subdomains to identify instances.
+      As such, the following URLs all include a {settings.Tenant} placeholder that will be dynamically
+      replaced by OpenIddict at runtime by the tenant configured in the Zendesk settings.-->
+
+    <Environment Issuer="https://{settings.Tenant}.zendesk.com/">
+      <Configuration AuthorizationEndpoint="https://{settings.Tenant}.zendesk.com/oauth/authorizations/new"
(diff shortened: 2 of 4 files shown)


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2153,6 +2153,27 @@
...
+    <!--Note: Zendesk is a multitenant provider that relies on subdomains to identify instances.
+      As such, the following URLs all include a {settings.Tenant} placeholder that will be dynamically
+      replaced by OpenIddict at runtime by the tenant configured in the Zendesk settings.-->
+
+    <Environment Issuer="https://{settings.Tenant}.zendesk.com/">
+      <Configuration AuthorizationEndpoint="https://{settings.Tenant}.zendesk.com/oauth/authorizations/new"
+                     TokenEndpoint="https://{settings.Tenant}.zendesk.com/oauth/tokens"
+                     UserInfoEndpoint="https://{settings.Tenant}.zendesk.com/api/v2/users/me">
+        <CodeChallengeMethod Value="S256" />
+        <GrantType Value="authorization_code" />
```
```suggestion
```

When `authorization_code` is the only grant type supported, adding it explicitly isn't required as it's always added by the provider generator when you don't explicitly set a list of grant types.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2153,6 +2153,27 @@
...
+
+      </Configuration>
+    </Environment>
+    <Setting PropertyName="Tenant" ParameterName="tenant" Type="String" Required="true"
+             Description="Gets or sets the tenant used to identify the Zendesk instance." />
+    <Setting PropertyName="ClientType" ParameterName="kind" Type="String" Required="false"
+         DefaultValue="confidential" Description="Client Type: public or confidential. Public OAuth clients are applications that run in environments where credentials cannot be securely stored, such as mobile and web apps. These clients are required to use PKCE. Confidential OAuth clients run on secure servers where their credentials can be kept secure. These clients can use PKCE, client secret, or both." />
```
```suggestion
```

This setting shouldn't be needed (and since you didn't write custom code to do something based on its value, it really has no effect whatsoever) 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2153,6 +2153,27 @@
...
+      <Configuration AuthorizationEndpoint="https://{settings.Tenant}.zendesk.com/oauth/authorizations/new"
+                     TokenEndpoint="https://{settings.Tenant}.zendesk.com/oauth/tokens"
+                     UserInfoEndpoint="https://{settings.Tenant}.zendesk.com/api/v2/users/me">
+        <CodeChallengeMethod Value="S256" />
+        <GrantType Value="authorization_code" />
+
+      </Configuration>
+    </Environment>
+    <Setting PropertyName="Tenant" ParameterName="tenant" Type="String" Required="true"
+             Description="Gets or sets the tenant used to identify the Zendesk instance." />
```
```suggestion
             Description="Gets or sets the tenant used to identify the Zendesk instance" />
```

(the . is automagically added by the generator)

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2153,6 +2153,27 @@
     <Environment Issuer="https://api.login.yahoo.com/" />
   </Provider>
 
+  <Provider Name="Zendesk" Id="5577b117-1f3d-464f-b987-aabf7c0ab3f7" Documentation="https://developer.zendesk.com/api-reference/introduction/security-and-auth/#oauth-access-token">
```
```suggestion
  <Provider Name="Zendesk" Id="5577b117-1f3d-464f-b987-aabf7c0ab3f7" Documentation="https://support.zendesk.com/hc/en-us/articles/4408845965210-Using-OAuth-authentication-with-your-application">
```

Let's use https://support.zendesk.com/hc/en-us/articles/4408845965210-Using-OAuth-authentication-with-your-application instead of https://developer.zendesk.com/api-reference/introduction/security-and-auth/#oauth-access-token as the last one doesn't contain anything really useful.

kevinchalet (commented):
Thanks for your PR! ❤️ 

> I am happy to help in testing validation as I have an account with Zendesk that is used for Open-source project testing.

Yes please! (I don't have a Zendesk account and won't be able to test it myself)

kevinchalet:
Hey @mozts2005,

If you have any question or need help to complete this PR, please let me know 😃 

Cheers.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
mozts2005: Should the value not be automatically added to the code request as it has both a property name and a parameter name? if not where should I update to include this value as this is required or the client secret is required. I would like to also add validation for configuration.

kevinchalet: > Should the value not be automatically added to the code request as it has both a property name and a parameter name?

Nope (note: the "parameter name" here refers to the name of the parameter used in the C# builder method, not an OIDC request parameter).

> if not where should I update to include this value as this is required or the client secret is required. I would like to also add validation for configuration.

You don't need that parameter: when no client secret or signing credentials is attached to a client registration, OpenIddict detects it when negotiating the client authentication method and sends the token request without a secret attached.

There's no need for special validation logic for Zendesk, it behaves like any other standard provider in this regard.

mozts2005:
I made the changes you requested. 
I also added support for revocation of the token. 
Please let me know if you would like to handle the revocation authentication differently. I did it this way as the same type of override is done for the Trovo userinfo endpoint authentication. https://github.com/openiddict/openiddict-core/blob/ff752efe09b319ee930dd181a77edd34a69a0f48/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs#L200
But I was thinking that if you wanted this could be added as a new EndpointAuthMethod if you prefer. 


Next, I added some mappings for to the UserInfo based on Zendesk endpoint results but if you would like I can remove them or add more.

kevinchalet:
> Please let me know if you would like to handle the revocation authentication differently.

No that's fine 👍🏻

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1362,6 +1362,9 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
 
                 // Shopify returns the email address as a custom "associated_user/email" node in token responses:
                 ProviderTypes.Shopify => (string?) context.TokenResponse?["associated_user"]?["email"],
+                
+                // Zendesk returns the emails address as a custom "user/email" node:
+                ProviderTypes.Zendesk => (string?) context.UserInfoResponse?["user"]?["email"],
```
kevinchalet: Looks like the userinfo details are wrapped in a `user` node? If so, we'll need to unwrap the userinfo response so all the claims can be correctly mapped: https://documentation.openiddict.com/guides/contributing-a-new-web-provider#unwrap-userinfo-responses-if-necessary

Once done, you'll need to remove the `?["user"]` part of these new lines.

kevinchalet:
Thanks. Could you please attach a screenshot of the returned claims (visible in the output generated by the console client sandbox)? Feel free to blur out personal/sensitive details.

Were you able to test everything? Revocation works fine?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
mozts2005: updated

mozts2005:
> Thanks. Could you please attach a screenshot of the returned claims (visible in the output generated by the console client sandbox)? Feel free to blur out personal/sensitive details.
> 
> Were you able to test everything? Revocation works fine?

yes, revocation works and here is the screenshot.
ZendeskAfterUnwrap

kevinchalet:
I pushed a commit that removes the `token` parameter from the revocation request (since Zendesk doesn't expect it and requires using bearer authentication instead). Could you please give it a try to ensure it still works fine with this change?

mozts2005:
> I pushed a commit that removes the `token` parameter from the revocation request (since Zendesk doesn't expect it and requires using bearer authentication instead). Could you please give it a try to ensure it still works fine with this change?


all changes made look good to me after testing.
image

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
```diff
@@ -406,8 +406,9 @@ public ValueTask HandleAsync(ExtractUserInfoResponseContext context)
                     ProviderTypes.ExactOnline => new(context.Response["d"]?["results"]?[0]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("d/results/0"))),
 
-                    // Fitbit and Todoist return a nested "user" object.
-                    ProviderTypes.Fitbit or ProviderTypes.Todoist => new(context.Response["user"]?.GetNamedParameters() ??
+                    // Fitbit, Todoist, and Zendesk return a nested "user" object.
```
kevinchalet:
```suggestion
                    // Fitbit, Todoist and Zendesk return a nested "user" object.
```

kevinchalet:
Merged. Thanks a lot for your contribution, @mozts2005! 👏🏻

PR #2227, opened 2024-11-15
Update OIC sponsor logo

karlrsk:
URL has changed for the OIC logo in the sponsor section. We have updated the readme to reflect a permanent logo.


kevinchalet (approved):
Thanks @karlrsk. Could you please update https://github.com/openiddict/openiddict-documentation/pull/141 to use the same image URL?

karlrsk:
I don't have access to his repo/PR - this one is the one we want to use and the other can be discarded.

PR #2233, opened 2024-11-28
Add Eve Online to the list of supported providers

kalaveijo:
Added Eve Online provider and implemented OpenID inclusion (since well known config does not have it automatically). Tested implementation using Sandbox.Console.Client and it seemed to work after the authentication dance was over. Was able to refresh and revoke token.

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
@@ -600,6 +600,27 @@
   </Provider>
 
   <!--
+                              ▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
+                              ██ ▄▄▄█ ██ █ ▄▄▄████ ▄▄▄ ██ ▀██ ██ ████▄ ▄██ ▀██ ██ ▄▄▄██
+                              ██ ▄▄▄█ ██ █ ▄▄▄████ ███ ██ █ █ ██ █████ ███ █ █ ██ ▄▄▄██
+                              ██ ▀▀▀█▄▀▀▄█ ▀▀▀████ ▀▀▀ ██ ██▄ ██ ▀▀ █▀ ▀██ ██▄ ██ ▀▀▀██
+                              ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+
+  <Provider Name="EveOnline" Id="e4614207-4f73-47c2-86c9-31f59d4c94f9"
+            Documentation="https://developers.eveonline.com/">
+    <Environment Issuer="https://login.eveonline.com/">
+      <Configuration AuthorizationEndpoint="https://login.eveonline.com/v2/oauth/authorize"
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR!

> Added Eve Online provider and implemented OpenID inclusion (since well known config does not have it automatically).

Interesting. Do they return an identity token when you add the `openid` scope?

kalaveijo:
Hello! Yes they do return an identity token when the openid scope is present. You can see token payload (courtesy of jwt.io) in attached image (censored some bits from the payload).

image

kevinchalet:
Hey 👋🏻

> Yes they do return an identity token when the openid scope is present.

Very interesting! It's strange they are not mentioning OpenID Connect support anywhere since it's a great way to improve their existing non-standard implementation (they use their JWT access tokens as a way to share user claims with the client, which is a terrible thing to do 🤣).

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -599,6 +599,27 @@
...
+  <!--
+                              ▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
+                              ██ ▄▄▄█ ██ █ ▄▄▄████ ▄▄▄ ██ ▀██ ██ ████▄ ▄██ ▀██ ██ ▄▄▄██
+                              ██ ▄▄▄█ ██ █ ▄▄▄████ ███ ██ █ █ ██ █████ ███ █ █ ██ ▄▄▄██
+                              ██ ▀▀▀█▄▀▀▄█ ▀▀▀████ ▀▀▀ ██ ██▄ ██ ▀▀ █▀ ▀██ ██▄ ██ ▀▀▀██
+                              ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+
+  <Provider Name="EveOnline" Id="e4614207-4f73-47c2-86c9-31f59d4c94f9"
+            Documentation="https://developers.eveonline.com/">
```
```suggestion
            Documentation="https://docs.esi.evetech.net/docs/sso/sso_authorization_flow.html">
```

The link you used seems very generic. What about using https://docs.esi.evetech.net/docs/sso/sso_authorization_flow.html instead?

kevinchalet:
BTW, last time that provider was mentioned, the company was still forcing developers to have a paid account to be able to test their OAuth service. Is it still true?

kalaveijo:
> BTW, last time that provider was mentioned, the company was still forcing developers to have a paid account to be able to test their OAuth service. Is it still true?

Unfortunately yes. While its possible to create free accounts, adding applications to https://developers.eveonline.com/ requires that account has had some kind of monetary movement (buy gametime or buy premium currency using real money)

> Very interesting! It's strange they are not mentioning OpenID Connect support anywhere since it's a great way to improve their existing non-standard implementation (they use their JWT access tokens as a way to share user claims with the client, which is a terrible thing to do 🤣).

CCP (developers of Eve Online) are known for their 'interesting' API support. Some endpoints have been broken for several years now. That's why its not very surprising that the well known config is slightly misleading.

kevinchalet:
> Unfortunately yes. While its possible to create free accounts, adding applications to https://developers.eveonline.com/ requires that account has had some kind of monetary movement (buy gametime or buy premium currency using real money)

Arf, that's unfortunate. I will count on you to keep an "eye" on that provider to ensure it still works fine over time, then (pun intended 🤣)

PR merged. Congrats for your contribution to OpenIddict! ❤️

kalaveijo:
Thank you! I plan on using OpenIddict on my personal Eve Online tools, so I'll fix it if it breaks :)

kevinchalet:
Awesome, thanks!

Note: I just updated https://documentation.openiddict.com/integrations/web-providers to mention that EVE Online is now supported 😃

Issue #1647, opened 2023-01-17
How to change validation type in runtime?


MDzyga:
### Confirm you've already contributed to this project or that you sponsor it

- [X] I confirm I'm a sponsor or a contributor

### Version

4.x

### Question

Hi @kevinchalet :)
I updated openIdDict to version 4 and I'm trying to use ChangeToken as you said. Unfortunately it doesn't work, change token is executed but OpenIdDict doesn't switch between Introspection and Direct. I tried with ChangeToken registered outside AddValidation, the same efect.

```
services
	.AddOpenIddict()
	.AddCore(options =>
	{
		options.UseEntityFrameworkCore()
			.UseDbContext<ApplicationDbContext>()
			.ReplaceDefaultEntities<Client, Authorization, Scope, Token, int>();
	})
	.AddServer(options =>
	{
		options
			.SetTokenEndpointUris("connect/token")
			.SetRevocationEndpointUris("connect/revocation")
			.SetIntrospectionEndpointUris("connect/introspect");

		options.AddEventHandler<OpenIddictServerEvents.ProcessSignInContext>(builder =>
		{
			builder.UseInlineHandler(context =>
			{
				if (context.Transaction.Request != null)
				{
					context.Response["scope"] = string.Join(" ", context.Transaction.Request.GetScopes().OrderBy(z => z));
				}

				return default;
			});
		});

		options.AddEventHandler<OpenIddictServerEvents.ProcessChallengeContext>(builder =>
		{
			builder.UseInlineHandler(context =>
			{
				var authPropertiesKey = typeof(AuthenticationProperties).ToString();
				if (context.Transaction.Properties.ContainsKey(authPropertiesKey))
				{
					var props = context.Transaction.Properties[authPropertiesKey] as AuthenticationProperties;
					context.Response["message"] = props.Items["message"];
				}

				return default;
			});
		});

		options.AllowClientCredentialsFlow();
		options.AllowPasswordFlow();
		options.AllowRefreshTokenFlow();

		options.UseReferenceRefreshTokens();

		options.RegisterScopes(
			OpenIddictConstants.Permissions.Scopes.Profile.Replace(OpenIddictConstants.Permissions.Prefixes.Scope, string.Empty),
			OpenIddictConstants.Permissions.Scopes.Roles.Replace(OpenIddictConstants.Permissions.Prefixes.Scope, string.Empty),
			CustomScopes.APPLICATION,
			CustomScopes.IDENTITY_SETUP,
[...]

PR #2234, opened 2024-12-11
Feature/thread safe runtime sign cred modification

jhbritton-RSK:
Added abstraction for SigingCredentials list in OpenIddICT server options to allow thread-safe modification of the singing credentials at runtime.


kevinchalet:
Thanks for your PR, @jhbritton-RSK.

This has been suggested before, but it's not really the "OpenIddict way" to implement that.

See https://github.com/openiddict/openiddict-core/pull/2190 and https://github.com/openiddict/openiddict-core/issues/1647 for a more "natural" approach using `Microsoft.Extensions.Options`.

jhbritton-RSK:
Ah, thank you for the input; I may have created this prematurely on the main repo anyway. Was gonna keep it on our fork until we were ready. Will take a look.

PR #2238, opened 2024-12-20
Fix where clause of `OpenIddictEntityFrameworkCoreAuthorizationStore#FindByApplicationIdAsync

simon-wacker:
Before that fix, when OppenIddict was configured with a custom application id type like `Guid`, the where-clause sent to the database was literally `where False`, because the type of `application.Id` was incompatible with the type of the erroneously used variable `identifier`.


kevinchalet:
Thanks for your PR!

Nice catch, I'll make sure it ships as part of the next 6.x release (likely a patch version).

PR #2246, opened 2025-01-29
Add GitCode to the list of supported providers

gehongyan:
This pull request would like to add GitCode to the list of supported providers.


gehongyan:
### Docs

GitCode offers documentation in Chinese and English for the first and second editions. These versions pertain solely to the documentation and not to the API versions.

- Authorization mechanism in [doc v1](https://docs.gitcode.com/en/docs/oauth/)
- Token API in [doc v2](https://docs.gitcode.com/en/docs/apis/post-oauth-token-grant-type-authorization-code-code-code-client-id-client-id-client-secret-client-secret)

---

### Hint

- It appears that GitCode does not support setting the scopes through query strings; the configuration of scopes (App permissions) is instead managed within the OAuth app management on the official web.

---

### Useful Screenshots

*Texts in these screenshots are all auto-translated by Google Chrome*

- GitCode OAuth App Editing

image

- Requesting Authentication

image

- Authenticated (ASP.NET)

image

- Authenticated (Console)

image

- Refreshing

image

kevinchalet:
Merged. Thanks for this great contribution, @gehongyan! ❤️ 🎉

PR #2249, opened 2025-02-10
Update the VK ID provider to use a custom display name

kevinchalet:
/cc @t1moH1ch sorry, I missed that during the review 😃


t1moH1ch (approved):
👍

PR #2254, opened 2025-02-21
Fix userinfo response unwrapping for discord provider

egans146:
This fix unwrap the discord userinfo response allowing OpenIddict mapping to the corresponding claims

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -413,7 +413,7 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                         throw new InvalidOperationException(SR.FormatID0334("d/results/0"))),
 
                     // These providers return a nested "user" object.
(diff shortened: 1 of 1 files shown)


kevinchalet:
Thanks for your PR, @egans146!

It's something I also noticed but since it's technically a breaking change, I wasn't sure if it was the right thing to do.

Since you decided to send a PR, I guess the existing logic (i.e returning the bot+user claims unflatten) is not ideal. On the other hand, if we do that, the bot claims will no longer be returned. A middle ground could be to flatten the user claims and still include a bot claims node as we do for some of the other providers. What do you think?

egans146:
Thank you for taking the time to respond.

I think your proposal would be a good middle ground. I may have initially rushed and failed to consider the other nodes that would no longer be returned.

You think the « bot » node nested in « application » should be included or the whole application node ?

kevinchalet:
After re-reading the docs, it seems there's also another option we can consider: Discord offers a dedicated `/users/@me` endpoint that seems to return more nodes than the `/oauth2/@me` API we're using: https://discord.com/developers/docs/resources/user#get-current-user

Maybe we should just change the userinfo endpoint address to use that endpoint?

kevinchalet:
> Maybe we should just change the userinfo endpoint address to use that endpoint?

Done: https://github.com/openiddict/openiddict-core/pull/2254. The change will land in the next 7.0 preview 😃 

Thanks for bringing that to my attention, @egans146! ❤️

PR #2256, opened 2025-02-25
Add Alibaba Cloud (Aliyun) to the list of supported providers

gehongyan:
This pull request would like to add Alibaba Cloud (Aliyun) to the list of supported providers.


gehongyan:
## Docs

Alibaba Cloud provides services to developers in mainland China under the brand name Aliyun. Documentation is provided on both Alibaba Cloud and Aliyun. OpenID is provided in the RAM (Resource Access Management) product.

* **OAuth Management**: [Alibaba Cloud](https://www.alibabacloud.com/help/en/ram/user-guide/oauth-management/), [Aliyun](https://help.aliyun.com/zh/ram/user-guide/oauth-management/)
* **Obtain user information by using OIDC**: [Alibaba Cloud](https://www.alibabacloud.com/help/en/ram/user-guide/obtain-user-information-through-oidc), [Aliyun](https://help.aliyun.com/zh/ram/user-guide/obtain-user-information-through-oidc)

## Hints

* Alibaba Cloud and Aliyun support discovery. See [Alibaba Cloud](https://oauth.alibabacloud.com/.well-known/openid-configuration), [Aliyun](https://oauth.aliyun.com/.well-known/openid-configuration).
* `expires_in` is formatted as a string. See [Alibaba Cloud](https://www.alibabacloud.com/help/en/ram/user-guide/access-alibaba-cloud-apis-from-a-web-application#section-04), [Aliyun](https://help.aliyun.com/zh/ram/user-guide/access-alibaba-cloud-apis-from-a-web-application#section-04).
* Token revocation requires a `client_id` parameter. See [Alibaba Cloud](https://www.alibabacloud.com/help/en/ram/user-guide/access-alibaba-cloud-apis-from-a-web-application#section-05), [Aliyun](https://help.aliyun.com/zh/ram/user-guide/access-alibaba-cloud-apis-from-a-web-application#section-05).

> [!NOTE]
> **Question 1:** The token revocation does not work yet in the console sample when refreshing tokens. It seems that only refresh tokens can be revoked. How do we handle this issue?
image

## Useful Screenshots

* Aliyun RAM console.

image

* Login page.

image

* Token refresh works in the ASP.NET sample. (A pretty long token)

image

* Extracted claims.

image

kevinchalet:
Thanks for your PR, @gehongyan!

Woooo, compared to their Alipay OAuth 2.0 implementation, it's really night and day! (because let's be honest, Alipay is the worst/most non-standard implementation I've ever seen in my entire life! 🤣)

This new Alibaba Cloud service doesn't cover Alipay, right?

> **Question 1:** The token revocation does not work yet in the console sample when refreshing tokens. It seems that only refresh tokens can be revoked. How do we handle this issue?
> image

It's rare, but it happens and it's something the specification allows. There's nothing we can do on our side, but as long as you're able to confirm it works fine by temporarily tweaking the console sample to send the refresh token instead of the access token, that's fine 👍🏻

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
```diff
@@ -68,6 +69,40 @@ public ValueTask HandleAsync(PrepareRevocationRequestContext context)
...
+            {
+                if (context is null)
+                {
+                    throw new ArgumentNullException(nameof(context));
+                }
+
+                // Alibaba Cloud requires attaching the "client_id" parameter to revocation requests.
+                if (context.Registration.ProviderType is ProviderTypes.AlibabaCloud)
+                {
+                    context.Request.ClientId = context.Registration.ClientId;
```
kevinchalet: Note: OpenIddict is expected to take care of sending the client credentials automatically (client_id/client_secret/client_assertion). Were you seeing an error without that custom handler?

gehongyan:
> This new Alibaba Cloud service doesn't cover Alipay, right?

Yes, Alibaba Cloud and Alipay are different products.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
gehongyan: In my recollection, the first time I invoked the token revocation, it returned an error saying the request was missing the `client_id` parameter. So, I immediately tried adding this handler.  After that, the error message changed to indicating that only `refresh_token` could be revoked.

Strangely, after removing the handler, I haven't been able to reproduce the original error.  Furthermore, after adjusting the console sample, the token revocation request succeeded.  Therefore, given the current situation, I'll remove it for now.

image

gehongyan:
Alibaba Cloud categorizes OAuth applications into WebApp and NativeApp. There appear to be subtle differences in their `refresh_token` return policies. WebApp should specify `access_type` as `offline` in the token request parameters to obtain a `refresh_token`, but NativeApp always returns `refresh_token`.

image
image
image

 This seems similar to Huawei's implementation. Using the Native application type directly can bypass this, but if a developer specifies WebApp, we should also support this scenario. I'll implement this later.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
kevinchalet: Maybe the first time you tried, your client was confidential and was assigned a client secret?

When OpenIddict doesn't find an explicit `revocation_endpoint_auth_methods_supported` node in the server configuration, it uses `client_secret_basic` - i.e the default method - if both a client identifier and client secret are expected to be sent. In that case, the client credentials are not sent as regular OAuth 2.0 parameters but are sent using the HTTP `WWW-Authenticate` header. If no client secret is configured, the `client_id` is sent alone as a regular OAuth 2.0 parameter.

Can you please check the type of client you configured with Alibaba's portal?

kevinchalet:
> Using the Native application type directly can bypass this, but if a developer specifies WebApp, we should also support this scenario. I'll implement this later.

It's indeed fairly common: Google does that too... and if you don't specify `access_type=offline` the very first time the user approves the authorization demand, you'll never get a refresh token, even if future authorization requests include `access_type=offline` (which is a TERRIBLE design!).

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
gehongyan: > Can you please check the type of client you configured with Alibaba's portal?

I tested both WebApp and NativeApp, and both can be revoked successfully.

gehongyan:
I've implemented the `access_type` and `prompt` parameters in the token request.

The `access_type` parameter successfully allows requesting a `refresh_token` in a WebApp, and I was able to revoke it successfully.

Regarding the `prompt` parameter, the documentation mentions that the `admin_consent` value should force the server to display an authorization page before returning to the client. The authorization page does appear, but clicking the authorization button on the page results in an error. I encountered the issue in both Alibaba Cloud and Aliyun. I have submitted a support ticket to Aliyun regarding this issue and am currently awaiting their response.

image

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
kevinchalet: > I tested both WebApp and NativeApp, and both can be revoked successfully.

After changing to web app, you updated your OpenIddict registration to call `SetClientSecret(...)`, right?

Anyway, I took a look at the docs and they don't mention basic authentication support.
Just in case, we should probably tweak the returned configuration response to add `client_secret_post` to both `token_endpoint_auth_methods_supported` and `revocation_endpoint_auth_methods_supported` so that client credentials are always sent as part of the request form.

Can you please give it a try?

https://github.com/openiddict/openiddict-core/blob/7517252174333f1cc963cc9cb2cfd55de4aba443/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Discovery.cs#L262-L361

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1862,6 +1862,18 @@ public ValueTask HandleAsync(ProcessChallengeContext context)
...
+            // By default, Alibaba Cloud doesn't return a refresh token for native applications but allows sending an
+            // "access_type" parameter to retrieve one (but it is only returned during the first authorization dance).
+            // The documentation also indicates the "prompt" parameter is supported but not required,
+            // which can be set to "admin_consent" to force the display of the authorization page
+            if (context.Registration.ProviderType is ProviderTypes.AlibabaCloud)
+            {
+                var settings = context.Registration.GetAlibabaCloudSettings();
+
+                context.Request["access_type"] = settings.AccessType;
+                context.Request["prompt"] = settings.Prompt;
```
kevinchalet:
```suggestion
                context.Request.Prompt = settings.Prompt;
```

You can directly use the `Prompt` property since it's a standard OIDC parameter exposed by the `OpenIddictRequest` primitive 😃

kevinchalet: While you're changing that, let's also change line 1862 to use the same pattern:

```csharp
context.Request.Resources = [settings.Resource];
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
gehongyan: > After changing to web app, you updated your OpenIddict registration to call SetClientSecret(...), right?

Actually, I added registrations for both application types in my test code to conduct these tests.

<details>
<summary>Test code</summary>

```csharp
.AddAlibabaCloud(options =>
    options
        .SetRegion("CN")
        .SetProviderName("AliyunWeb")
        .SetProviderDisplayName("Aliyun Web")
        .SetClientId("******************")
        .SetClientSecret("***********************************************")
        .SetAccessType("offline")
        // .SetPrompt("admin_consent")
        .SetRedirectUri("callback/login/aliyun-web")
        .AddScopes("aliuid", "profile"))
// Aliyun NativeApp
.AddAlibabaCloud(options =>
    options
        .SetRegion("CN")
        .SetProviderName("AliyunNative")
        .SetProviderDisplayName("Aliyun Native")
        .SetClientId("******************")
        .SetClientSecret("***********************************************")
        // .SetPrompt("admin_consent")
        .SetRedirectUri("callback/login/aliyun-native")
        .AddScopes("aliuid", "profile"))
// Alibaba Cloud WebApp
.AddAlibabaCloud(options =>
    options
        .SetProviderName("AlibabaWeb")
        .SetProviderDisplayName("Alibaba Web")
        .SetClientId("******************")
        .SetClientSecret("***********************************************")
        .SetRedirectUri("callback/login/alibabacloud-web")
        .SetAccessType("offline")
        // .SetPrompt("admin_consent")
        .AddScopes("aliuid", "profile"))
// Alibaba Cloud NativeApp
.AddAlibabaCloud(options =>
    options
        .SetProviderName("AlibabaNative")
        .SetProviderDisplayName("Alibaba Native")
        .SetClientId("******************")
        .SetClientSecret("***********************************************")
        // .SetPrompt("admin_consent")
        .SetRedirectUri("callback/login/alibabacloud-native")
        .AddScopes("aliuid", "profile"))
```

</details>

> Just in case, we should probably tweak the returned configuration response.

Sounds good. Completed. All of the registrations above still work well.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
gehongyan: Completed.

I also changed `"display"` to the `Display` property in Huawei and Weibo providers. I hope that I haven't made things worse.

kevinchalet: Excellent, thanks!

> I also changed `"display"` to the `Display` property in Huawei and Weibo providers. I hope that I haven't made things worse.

Great initiative 👍🏻

gehongyan:
```diff
- context.Request["resource"] = settings.Resource;
+ context.Request.Resources = [settings.Resource];
```

The type of `"resource"` is changed from a `string?` to a `string?[]?` likely resulting in a breaking change.

gehongyan:
Response from Aliyun support:

> `admin_consent` is only used for cross-account authorization. Therefore, when the client ID and the logged-in account belong to the same account, even though passing `prompt=admin_consent` will display the authorization page, it won't be possible to grant consent to an application within the same account.  An error message, indicating failure to install, will be displayed. This behavior is by design.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: > The type of "resource" is changed from a `string?` to a `string?[]?` likely resulting in a breaking change.

How that property is represented internally by the `OpenIddictRequest` primitive doesn't affect how it's eventually represented as an OAuth 2.0 parameter: for authorization request parameters, multi-valued parameters are always represented as multiple parameters (which was explicitly forbidden in the initial OAuth 2.0 specification, but was allowed in subsequent specifications, including the specification that defines the standard version of the `resource` parameter). Here, since there's just one value, the representation in the authorization request doesn't actually change at all.

That said, I had not realized that setting was optional in the AD provider, so it's possible that an empty parameter value may be set by using the property. Eventually, there shouldn't be any difference in the actual authorization request as OpenIddict excludes parameters with empty values when generating the URI (e.g multiple `resource` parameters with `string.Empty` or `null` as the value), but you're right it's technically a behavior change that is observable from custom event handlers, so I reverted it (I just used the appropriate constant since it's a standard parameter) 👍🏻

kevinchalet:
> Response from Aliyun support:
> 
> > `admin_consent` is only used for cross-account authorization. Therefore, when the client ID and the logged-in account belong to the same account, even though passing `prompt=admin_consent` will display the authorization page, it won't be possible to grant consent to an application within the same account.  An error message, indicating failure to install, will be displayed. This behavior is by design.

Interesting 👍🏻 (their response time is also top notch!)

If you don't plan on making additional changes, I'll merge your PR later today 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
gehongyan: When I made the change, I did have doubts about whether it was a breaking change, which is why I wrote that deleted reply. Later, however, I realized it was about the internal handling of parameters, so I deleted it.

I truly couldn't think of or understand if it would have any other impacts. I appreciate that you considered and pointed it out, which is truly admirable. It seems that raising reasonable doubts is never a bad idea after all. 👍

gehongyan:
> If you don't plan on making additional changes, I'll merge your PR later today 😃

Sure! Big thanks for your support! ❤️

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: > I appreciate that you considered and pointed it out, which is truly admirable. It seems that raising reasonable doubts is never a bad idea after all. 👍

🙏🏻 

(using the dedicated property actually looked nicer so I pushed a commit to bring it back but guard the call with a `string.IsNullOrEmpty()` check 😄)

kevinchalet:
Merged. Thanks a lot for this great new contribution! 🎉 🙏🏻

PR #2271, opened 2025-03-16
Update `OpenIddictParameter` to offer `ImmutableArray<string>` conversions instead of `string[]` and clone `JsonNode` objects to guarantee immutability

kevinchalet:
This PR introduces the following changes:

  - It replaces all the `OpenIddictMessage`/`OpenIddictParameter` constructors and static operators using `string?[]?` objects by equivalents using `ImmutableArray<string?>` or `ImmutableArray<string?>?`.
  - The `OpenIddictRequest.Audiences` and `OpenIddictRequest.Resources` properties have been updated to use `ImmutableArray<string?>?` instead of `string?[]?`, which will fully users from trying to do unsupported things like `context.Request.Audiences[2] = "overridden audience"` (which may or may not currently work depending on the actual CLR type of the parameter value).
  - For similar reasons, `JsonNode` instances are now cloned by `OpenIddictParameter`'s constructor and cloned by the `JsonNode?` conversion operator.
  - It removes the `OpenIddictParameter.Value` property and replaces it by a new `OpenIddictParameter.GetRawValue()` to discourage using this very-low-level API.
  - The `GetDestinations()`/`SetDestinations()` now use `ImmutableDictionary<string, ImmutableArray<string>>` instead of `ImmutableDictionary<string, string[]>` for consistency with the previous changes.
  - The `OpenIddictParameter` structure was updated to use recently-introduced APIs like `JsonNode.DeepEquals()`, `JsonElement.DeepEquals()` or `JsonElement.GetPropertyCount()`.
  - New `StringValues` conversion operators have been added to the `OpenIddictParameter` primitive.
  - The `OpenIddictRequest.Claims` and `OpenIddictRequest.Registration` properties are now of type `JsonObject` instead of `JsonElement`, which should make these properties easier to use.


kevinchalet:
Note: since this PR introduces a bunch of breaking changes, it will need to be properly documented in the 6.x -> 7.x migration guide.

PR #2282, opened 2025-03-24
Use `Uri.TryCreate()` to build the base and request URIs used by OpenIddict

kevinchalet:
See https://github.com/microsoft/fhir-server/pull/4873#discussion_r2010905623.


kevinchalet:
Note: it will be backported to 6.2.0 later this week.

PR #2284, opened 2025-03-26
Adds Linear, Miro, and Webflow web providers

jerriep:
This adds the web providers for Linear, Miro, and Webflow.

## Linear

image

## Miro

image

## Webflow

image


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
```diff
@@ -433,10 +433,18 @@ ProviderTypes.Patreon or ProviderTypes.Pipedrive or ProviderTypes.Twitter
...
+                    ProviderTypes.Linear => new(context.Response["data"]?["viewer"]?.GetNamedParameters() ??
+                        throw new InvalidOperationException(SR.FormatID0334("data/viewer"))),
+
                     // Meetup returns a nested "self" object that is itself nested in a GraphQL "data" node.
                     ProviderTypes.Meetup => new(context.Response["data"]?["self"]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("data/self"))),
 
+                    // Miro returns a nested "user" object.
+                    ProviderTypes.Miro => new(context.Response["user"]?.GetNamedParameters() ??
+                        throw new InvalidOperationException(SR.FormatID0334("user"))),
```
jerriep: The response from Miro contains multiple nested objects, for example

```json
{
  "type": "oAuthToken",
  "team": {
    "type": "team",
    "name": "Doc sample team",
    "id": "3074457362577833142"
  },
  "createdBy": {
    "type": "user",
    "name": "John Smith",
    "id": "3074457362577955221"
  },
  "scopes": [
    "boards:write",
    "boards:read"
  ],
  "organization": {
    "type": "organization",
    "name": "Doc sample team",
    "id": "3074457363786315310"
  },
  "user": {
    "type": "user",
    "name": "John Smith",
    "id": "3974457362577957221"
  }
}
```

This code extracts the `user` object and allows us to extract the properties from this object. However, it is completely valid for a user of this library to also want to extract properties from the `team` and `organization` as claims. In `AspNet.Security.OAuth.Providers` we handled this with multiple calls to `ClaimActions.MapJsonSubKey` (see https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/fc444407375ccc6ad2ceb8b024e296f0167c8390/src/AspNet.Security.OAuth.Miro/MiroAuthenticationOptions.cs#L23-L28). I am not sure what the correct way is to handle this in the OpenIddict Web Providers project.

Please advise

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1147,6 +1147,15 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
...
                 context.UserInfoRequest["fields"] = string.Join(",", settings.Fields);
             }
+            
+            // Linear's userinfo endpoint is a GraphQL implementation that requires
+            // sending a proper "query" parameter containing the requested user details.
+            else if (context.Registration.ProviderType is ProviderTypes.Linear)
+            {
+                var settings = context.Registration.GetLinearSettings();
+
+                context.UserInfoRequest["query"] = $"query {{ viewer {{ {string.Join(" ", settings.UserFields)} }} }}";
```
jerriep, line 1157: This works fine for single fields, but the user may want to also extract nested objects. In this case, we can trick the provider by supplying the nested object with its fields as a single field, for example:

```cs
options.SetClientId("f94053c5e6e1428c7c17293c6d116dc2")
   .SetClientSecret("192e45d283d9f8c7f36cf499a8b093a6")
   .AddUserFields("id", "name", "organization { id name }")
   .SetPrompt("consent")
   .SetRedirectUri("callback/login/linear");
```

However, I think it would be better to specify nested properties using dot notation as follows:

```cs
options.SetClientId("f94053c5e6e1428c7c17293c6d116dc2")
   .SetClientSecret("192e45d283d9f8c7f36cf499a8b093a6")
   .AddUserFields("id", "name", "organization.id", "organization.name")
   .SetPrompt("consent")
   .SetRedirectUri("callback/login/linear");
```

This makes the intent clearer. I already created a little helper class that can do that as per the following code:

```cs
public static class GraphQlQueryFieldBuilder
{
    public static string Build(string[] fields)
    {
        var root = new NestedField();

        // Build the tree structure
        foreach (var key in fields)
        {
            var parts = key.Split('.');
            var current = root;

            foreach (var part in parts)
            {
                current.Fields ??= new Dictionary<string, NestedField>(StringComparer.Ordinal);

                if (!current.Fields.TryGetValue(part, out var next))
                {
                    next = new NestedField();
                    current.Fields[part] = next;
                }

                current = next;
            }
        }

        return BuildFieldsFromTree(root);
    }
    
    private static string BuildFieldsFromTree(NestedField node)
    {
        if (node.Fields is null || node.Fields.Count == 0)
            return string.Empty;

        var fields = string.Join(" ",
            node.Fields.Select(kvp =>
                kvp.Value.Fields is null || kvp.Value.Fields.Count == 0
                    ? kvp.Key
                    : $"{kvp.Key} {{ {BuildFieldsFromTree(kvp.Value)} }}"));

        return fields;
    }

    private class NestedField
    {
        public Dictionary<string, NestedField>? Fields { get; set; }
    }
}
```

So my questions are:
1. Should I go this route to allow the user to use dot notation
2. Where should I place this helper file in the project? Should I just add it as an internal class to `OpenIddict.Client.WebIntegration`

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
```diff
@@ -433,10 +433,18 @@ ProviderTypes.Patreon or ProviderTypes.Pipedrive or ProviderTypes.Twitter
                         => new(context.Response["data"]?.GetNamedParameters() ??
                         throw new InvalidOperationException(SR.FormatID0334("data"))),
 
+                    // Linear returns a nested "viewer" object that is itself nested in a GraphQL "data" node.
+                    ProviderTypes.Linear => new(context.Response["data"]?["viewer"]?.GetNamedParameters() ??
```
jerriep, line 437: Here we have the potential for a similar issue as for Miro when we query a nested object (as discussed elsewhere Linear). When we do that, the nested object comes back as a JSON string, e.g.

image

So, the same issue as below for Miro where we used to resolve this with `ClaimActions.MapJsonSubKey` in the OAuth providers. What is the correct way to handle this here?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1069,6 +1069,28 @@
     </Environment>
   </Provider>
 
+  <!--
+  LINEAR
```
jerriep: How do I create the text graphic for the names?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: Let's keep the logic simple/stupid by opting for option 1. Users who need to customize the `query` parameter beyond the simple fields logic we support will always be able to add an event handler that sets the `query` parameter to any wanted value.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
kevinchalet: You can use https://textfancy.com/text-art/ with the font `Small Negative` (make sure you add a space after the provider name so it doesn't look weird). I'll update the docs to add a link.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
kevinchalet: In general, OpenIddict tries to preserve the userinfo response as-is so that the resulting `ClaimsPrincipal` is as close as possible to the userinfo representation (unlike the aspnet-contrib providers, we don't filter anything and all the claims are made available to the application).

Here (and it's luckily quite rare), the returned representation is indeed suboptimal. There are three approaches to work around that:
  - Do not flatten anything: in that case everything will be stored as JSON claims. It works but it's not very convenient and users may want a different approach: https://github.com/openiddict/openiddict-core/pull/2254.

  - Flatten the `user` node while still including the other nodes (that will be represented as JSON claims):

```csharp
ProviderTypes.Micro => new(context.Response["user"]?.GetNamedParameters() ??
    throw new InvalidOperationException(SR.FormatID0334("user")))
{
    ["organization"] = context.Response["organization"],
    ["team"] = context.Response["team"]
}
```

  - Rewrite the entire response to use a completely flattened structure (which would end up generating claims identical to what you'd get with the claim mapping rules in the aspnet-contrib provider):

```csharp
ProviderTypes.Micro => new()
{
    ["user_id"] = context.Response["user"]?["id"],
    ["user_name"] = context.Response["user"]?["name"],
    ["organization_id"] = context.Response["organization"]?["id"],
    ["organization_name"] = context.Response["organization"]?["name"],
    ["team_id"] = context.Response["team"]?["id"],
    ["team_name"] = context.Response["team"]?["name"]
}
```

We already use approach 2 for Basecamp and Harvest, so it's probably the most logical one.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
kevinchalet: We can leave it as-is (users will just have to use `JsonObject.Parse(result.Principal.GetClaim("organization"))` to get the organization details) or we can artificially add `organization_id` and `organization_name` nodes (while still returning the raw `organization` JSON node) if you prefer:

```csharp
ProviderTypes.Linear => new(context.Response["data"]?["viewer"]?.GetNamedParameters() ??
    throw new InvalidOperationException(SR.FormatID0334("user")))
{
    ["organization_id"] = context.Response["data"]?["viewer"]?["organization"]?["id"],
    ["organization_name"] = context.Response["data"]?["viewer"]?["organization"]?["name"]
}
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1069,6 +1069,29 @@
...
+                     TokenEndpoint="https://api.linear.app/oauth/token"
+                     RevocationEndpoint="https://api.linear.app/oauth/revoke"
+                     UserInfoEndpoint="https://api.linear.app/graphql" />
+    </Environment>
+
+    <Setting PropertyName="UserFields" ParameterName="fields" Collection="true" Type="String"
+             Description="The list of user fields to expand from the GraphQL endpoint (by default, only basic fields are requested)">
+      <Item Value="id"    Default="true" Required="false" />
+      <Item Value="name"  Default="true" Required="false" />
+      <Item Value="email" Default="true" Required="false" />
```
kevinchalet: Let's sort the fields alphabetically for consistency with the other providers that use the same pattern 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
```diff
@@ -56,14 +111,26 @@ public ValueTask HandleAsync(PrepareRevocationRequestContext context)
...
                 {
                     context.Request.AccessToken = context.Token;
                     context.Request.Token = null;
                     context.Request.TokenTypeHint = null;
                 }
+                
+                // Miro uses a JSON payload that expects the "accessToken", "clientId", and "clientSecret" properties
+                else if (context.Registration.ProviderType is ProviderTypes.Miro)
+                {
+                    context.Request["accessToken"] = context.Token;
```
kevinchalet, line 138: Since you're going to send the access token in the `Authorization` header, there's no real point mapping it here. And if you don't nullify `context.Request.Token` here, you can reuse the same branch as the `Zendesk` provider in the `AttachBearerAccessToken` handler 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
```diff
@@ -56,14 +111,26 @@ public ValueTask HandleAsync(PrepareRevocationRequestContext context)
...
+                    context.Request["clientId"] = context.Request.ClientId;
```
kevinchalet, line 139: Interesting, I don't see these parameters mentioned in https://developers.linear.app/docs/oauth/authentication#id-6.-revoke-an-access-token. Are we sure their non-standard revocation endpoint supports client authentication?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
```diff
@@ -148,6 +215,16 @@ public ValueTask HandleAsync(PrepareRevocationRequestContext context)
                     context.Request.Token = null;
                 }
 
+                // Miro requires using bearer authentication with the token that is going to be revoked.
+                // In the case of Miro, we renamed the Token to accessToken, so we cannot use the Token property.
+                else if (context.Registration.ProviderType is ProviderTypes.Miro)
+                {
+                    if (context.Request.TryGetParameter("accessToken", out var parameter) && parameter.GetRawValue() is string accessToken)
```
kevinchalet: You'll be able to remove that branch so it's not really important, but FYI, using `GetRawValue()` is discouraged as there's no absolute guarantee a parameter will be backed by a CLR `string` value: here, we use `context.Request["accessToken"] = context.Token` earlier, so it may indeed be backed by a `string`, but a custom event handler registered by the user could change that to say, a `JsonValue` representing a `string`:

```csharp
context.Request["accessToken"] = JsonValue.Create(context.Token);
```

It's better to always use the conversion operators so that the `OpenIddictParameter` structure gives you the correct target type independently of its backing CLR type:

```csharp
var token = (string?) context.Request["access_token"];
if (!string.IsNullOrEmpty(token))
{
    request.Headers.Authorization = new AuthenticationHeaderValue(Schemes.Bearer, token);
}
```

... or if you prefer the minimally-indented version... 😄 

```csharp
else if (context.Registration.ProviderType is ProviderTypes.Miro &&
    (string?) context.Request["access_token"] is { Length: > 0 } token)
{
    request.Headers.Authorization = new AuthenticationHeaderValue(Schemes.Bearer, token);
}
```

kevinchalet (commented):
Looks great, thanks for your PR, @jerriep! ❤️

kevinchalet:
Note: if we can merge this PR this week, I'll make sure it's included in the 6.2.0 release 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
kevinchalet: > I'll update the docs to add a link.

Done: https://documentation.openiddict.com/guides/contributing-a-new-web-provider#add-the-ascii-art-header.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
jerriep: OK, I added the `team` and `organization` as JSON claims. See updated screenshot in the PR description

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
jerriep: I kept is as-is. The default GraphQL query will not request those organization fields and there is no way for us to know which fields the user will query. Therefore, we cannot assume these fields will be present in the JSON response. Keeping it as-is will keep it in line with the approach we took for Miro (and Basecamp and Harvest)

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
jerriep: Miro requires both the `accessToken` property in the JSON payload and the Bearer token. If I send just the Bearer token without the `accessToken` property, I get an error:

image

Likewise, if I omit the Bearer token and send just the `accessToken` property, it fails as well:

image

I need to supply both to get a successful response:

image

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
jerriep: Since both the `accessToken` property in the JSON payload, as well as the Bearer token are required, I fixed this code as per your suggestion.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
jerriep: Yeah, Linear is an interesting one. Before I updated the parameters according to their docs, I tested the revocation and it worked:

image

So, in the initial PR, I did not make the code changes according to their docs. But, it is clearly risky doing this as it is not documented to work this way, and they can easily change it in the future so it does not work correctly anymore. 

I have now updated the code to conform to their docs:

image

jerriep:
Hey @kevinchalet, I believe I addressed all the points you raised. I also picked up that the Miro and Webflow nameidentifier claim was not resolving correctly, so I fixed that as well. I did a final test of each provider doing the authorization and revocation and everything works as expected.

Please review again

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
```diff
@@ -56,15 +111,37 @@ public ValueTask HandleAsync(PrepareRevocationRequestContext context)
...
                 }
 
+                // Linear requires only the access_token and no other parameters
+                else if (context.Registration.ProviderType is ProviderTypes.Linear)
+                {
+                    context.Request.AccessToken = context.Token;
+                    context.Request.Token = null;
+                    context.Request.TokenTypeHint = null;
+                    context.Request.ClientId = null;
+                    context.Request.ClientSecret = null;
```
kevinchalet: It works just fine, but there's actually a cleaner way to solve this problem: letting OpenIddict know that the revocation endpoint doesn't support client authentication so that it doesn't even attach the credentials to the request.

For that, you'll just need to add `<RevocationEndpointAuthMethod Value="none" />` under the provider configuration (if you get a XML schema warning telling you the `none` value isn't allowed, you can rebase your branch, I tweaked the schema yesterday to allow all supported values).

```xml
<Configuration AuthorizationEndpoint="https://linear.app/oauth/authorize"
               TokenEndpoint="https://api.linear.app/oauth/token"
               RevocationEndpoint="https://api.linear.app/oauth/revoke"
               UserInfoEndpoint="https://api.linear.app/graphql">
  <RevocationEndpointAuthMethod Value="none" />
</Configuration>
```

If you do that, you should be able to merge that `else if` branch with the previous one.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
kevinchalet: Ah yeah, I thought that branch was for the Linear provider... that uses a single `access_token` parameter 🤣

kevinchalet:
> Please review again

Looks great. Just a minor improvement remaining and it should be good to merge 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Revocation.cs
jerriep: I added the `<RevocationEndpointAuthMethod Value="none" />` and merged the logic branch with the previous one. That caused the `client_secret` not to be sent, but it still sent the `client_id`.

image

So, I decided to keep the Linear logic in its own branch and clear the `client_id` by explicitly setting `ClientId` to `null`.

image

jerriep:
I did another final run through the authorization and revocation flow for all three providers and everything still works. This PR should be good to go.

kevinchalet:
Merged, thanks for your contribution! ❤️ 

(I'll backport your PR to 6.2.0 and release it later today)

PR #2292, opened 2025-04-08
Update ID0051 description

karollitka:
InvalidOperationException message thrown when EndSession is not configured properly suggests to use EnableLogoutEndpointPassthrough() method, however that method does not exist. Updated suggestion to use EnableEndSessionEndpointPassthrough() as I believe to be the correct one.


kevinchalet:
Nice catch!

Merged, thanks for your PR 👍🏻

PR #2312, opened 2025-05-07
Add Genesys as Web Provider

MikeAlhayek:
- Add Genesys as a new Web provider https://developer.genesys.cloud/
- The formatting change was done automatically by VS.


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2472,4 +2472,41 @@
...
+                                                ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+
+  <Provider
+    Name="Genesys"
+    DisplayName="Genesys Cloud"
+    Id="2e34b9d0-5c1f-4e23-9ac9-b4aefde213d2"
+    Documentation="https://developer.genesys.cloud/">
+
+    <Environment Issuer="{settings.Issuer}">
```
MikeAlhayek: alternatively we can use this if Regular expressions are supported. If Regex is supported here, let me know if you like me to make this change.

```
{settings.Issuer.Scheme}://{System.Text.RegularExpressions.Regex.Replace(settings.Issuer.Host, @"^(login\.|api\.)", "")}
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Discovery.cs
```diff
@@ -258,6 +258,15 @@ public ValueTask HandleAsync(HandleConfigurationResponseContext context)
                     context.Configuration.ScopesSupported.Add(Scopes.OpenId);
                 }
 
+                // Genesys supports oAuth but doesn't format the "openid" scope using the standard casing.
+                // To ensure Genesys is not treated as an OAuth 2.0-only provider, the invalid "OpenId" scope is
+                // removed from the list and the "openid" value is added to indicate OpenID Connect is supported.
+                else if (context.Registration.ProviderType is ProviderTypes.Genesys)
```
kevinchalet: Why do you need that? The provider uses a static configuration, so a configuration request will never be sent. It's basically dead code 😃

kevinchalet: Please revert all the changes made to that file (formatting and logic changes).

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1147,7 +1147,7 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
 
                 context.UserInfoRequest["fields"] = string.Join(",", settings.Fields);
             }
-            
+
```
kevinchalet, line 1150: This formatting change is fine but please revert everything else.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1412,28 +1412,31 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
                 // Yandex returns the email address as a custom "default_email" node:
                 ProviderTypes.Yandex => (string?) context.UserInfoResponse?["default_email"],
 
+                // Genesys returns the email address as a custom "default_email" node:
+                ProviderTypes.Genesys => context.MergedPrincipal.GetClaim("email"),
```
kevinchalet: Multiple issues here:
  - The code doesn't match the associated comment.
  - You're not supposed to use the merged principal when looking for claims, that should be retrieved from userinfo or token responses.
  - Assuming the email addresses is returned as an `email` node in the userinfo response, you can remove that because that's the standard claim so custom logic isn't needed.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1494,13 +1497,16 @@ ProviderTypes.Yandex  or ProviderTypes.Zoom
                 // Zoho returns the username as a custom "Display_Name" node:
                 ProviderTypes.Zoho => (string?) context.UserInfoResponse?["Display_Name"],
 
+                // Genesys returns the username as a custom "Display_Name" node:
+                ProviderTypes.Genesys => context.MergedPrincipal.GetClaim("name"),
```
kevinchalet: Same remark here, it's not valid and doesn't match the code comment. If the standard `name` claim is used, remove that line.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1585,9 +1591,12 @@ ProviderTypes.Webflow       or ProviderTypes.Weibo      or ProviderTypes.Yandex
                 // WordPress returns the user identifier as a custom "ID" node:
                 ProviderTypes.WordPress => (string?) context.UserInfoResponse?["ID"],
 
-                // WordPress returns the user identifier as a custom "ZUID" node:
+                // Zoho returns the user identifier as a custom "ZUID" node:
```
kevinchalet, line 1588: 👍🏻

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1585,9 +1591,12 @@ ProviderTypes.Webflow       or ProviderTypes.Weibo      or ProviderTypes.Yandex
...
                 ProviderTypes.Zoho => (string?) context.UserInfoResponse?["ZUID"],
 
+                // Genesys returns the user identifier as a custom "ZUID" node:
+                ProviderTypes.Genesys => context.MergedPrincipal.GetClaim("id"),
```
kevinchalet: Assuming the user identifier is returned as an `id` userinfo node, remove that and add the provider type to the switch case line 1522.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1937,7 +1946,7 @@ public ValueTask HandleAsync(ProcessChallengeContext context)
             else if (context.Registration.ProviderType is ProviderTypes.Linear)
             {
                 var settings = context.Registration.GetLinearSettings();
-                
+
```
kevinchalet, line 1940: This formatting change is fine to keep too, but revert all the other ones.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2472,4 +2472,41 @@
     </Environment>
   </Provider>
 
+  <!--
+                                                ▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
```
kevinchalet: You used the wrong font name and forgot the trailing space: https://documentation.openiddict.com/guides/contributing-a-new-web-provider#add-the-ascii-art-header

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2472,4 +2472,41 @@
...
+                                                ██░▄▄░█░▄▄█░▄▄▀█░▄▄█░▄▄█░██░█░▄▄
+                                                ██░█▀▀█░▄▄█░██░█░▄▄█▄▄▀█░▀▀░█▄▄▀
+                                                ██░▀▀▄█▄▄▄█▄██▄█▄▄▄█▄▄▄█▀▀▀▄█▄▄▄
+                                                ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+
+  <Provider
+    Name="Genesys"
```
kevinchalet: No need for so many line breaks 😄

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2472,4 +2472,41 @@
...
+    DisplayName="Genesys Cloud"
+    Id="2e34b9d0-5c1f-4e23-9ac9-b4aefde213d2"
+    Documentation="https://developer.genesys.cloud/">
```
kevinchalet: A more specific documentation page - ideally dedicated to OAuth 2.0 - would be better.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -2472,4 +2472,41 @@
...
+      <Configuration AuthorizationEndpoint="{settings.Issuer.Scheme}://login.{settings.Issuer.Host}/oauth/authorize"
+                     TokenEndpoint="{settings.Issuer.Scheme}://login.{settings.Issuer.Host}/oauth/token"
+                     UserInfoEndpoint="{settings.Issuer.Scheme}://api.{settings.Issuer.Host}/api/v2/users/me">
+        <CodeChallengeMethod Value="plain" />
+        <CodeChallengeMethod Value="S256" />
+
+        <GrantType Value="client_credentials" />
+        <GrantType Value="authorization_code" />
+        <GrantType Value="refresh_token" />
+        <GrantType Value="implicit" />
```
kevinchalet: Let's remove this one: Genesys only seems to support the insecure OAuth 2.0-only implicit flow flavor and not the OIDC version (which is safer).

kevinchalet (commented):
Thanks for your PR!

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
MikeAlhayek: I think I used the correct one `Small Negative`. But noticed that we should use all caps "which is not part of the manual".

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
MikeAlhayek: this is what the server supports:

image

MikeAlhayek:
@kevinchalet I made the requested changes.

kevinchalet:
Merged: https://github.com/openiddict/openiddict-core/commit/0c58cba5153f4b6e0a24b65545d5c34a848d5409.

Thanks for your contribution! ❤️

PR #2344, opened 2025-06-09
Add Bungie.net to the list of supported providers

kevinchalet:
https://github.com/Bungie-net/api/wiki/OAuth-Documentation


kevinchalet:
Note: this PR will be backported to the 6.x branch.

PR #2355, opened 2025-06-21
Add read-replica support for EF Core stores

tumugin:
This change introduces support for using read-replicas when Entity Framework Core is used as the ORM.

In our application, we balance database load by directing write operations to a primary database and read operations to a read-replica. This is achieved by using separate DbContext instances for each type of operation. To apply this same load balancing strategy when using OpenIddict, we wanted to enable similar functionality within this library.

To ensure backward compatibility, the existing `UseDbContext<TContext>()` method remains unchanged. For enabling read/write splitting, a new overload has been introduced: `UseDbContext<TWriteContext, TReadContext>()`. This allows for the explicit registration of different DbContext types for write and read operations.

### DIFF
--- a/src/OpenIddict.EntityFrameworkCore/IOpenIddictEntityFrameworkCoreContext.cs
+++ b/src/OpenIddict.EntityFrameworkCore/IOpenIddictEntityFrameworkCoreContext.cs
@@ -15,11 +15,20 @@ namespace OpenIddict.EntityFrameworkCore;
 public interface IOpenIddictEntityFrameworkCoreContext
 {
     /// <summary>
-    /// Gets the <see cref="DbContext"/>.
+    /// Gets the <see cref="DbContext"/> for write operation.
     /// </summary>
     /// <returns>
     /// A <see cref="ValueTask{TResult}"/> that can be used to monitor the
     /// asynchronous operation, whose result returns the <see cref="DbContext"/>.
     /// </returns>
-    ValueTask<DbContext> GetDbContextAsync(CancellationToken cancellationToken);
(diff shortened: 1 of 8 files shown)


kevinchalet:
Thanks for your PR!

Honestly, this is a fairly unusual pattern in the ASP.NET Core world (most notably, ASP.NET Core Identity doesn't support that).

How do the two `DbContext`s coordinate in your application? Let's say I'm adding a new client application entry via the "write context", will it be immediately reflected in the "read context" in the same service scope (typically, the same ASP.NET Core request)?

tumugin:
Thank you for your prompt reply.

Depending on the database settings, replication delays may occur and not be reflected immediately.

Indeed, it may be necessary to use the DBContext for writing instead of the DBContext for reading after write processing.

kevinchalet:
Hum, I'm afraid the potential for confusion and subtle bugs is fairly high with such a feature (e.g we do our best in the stores to disable query tracking, but if someone overrides some of the `Find*Async()` methods and doesn't use `AsNoTracking()`, they'll hit an error telling them the object is already tracked by another `DbContext` when trying to use the "write context" to update or delete an entity previously retrieved via the "read context").

For such scenarios, it's likely better to use custom stores so you can have much more control over how things are handled.

Closing for now, but happy to reconsider if there are additional demands.

Thanks for your PR!

tumugin:
> For such scenarios, it's likely better to use custom stores so you can have much more control over how things are handled.

Thank you very much, I will try to get this implemented in my own code. I didn't have a good understanding of my code when I first submitted this PR, but I will try to do that first because I have found that if we need this kind of implementation, it is something we should implement in our own code with a custom store.

PR #2359, opened 2025-07-01
add IdentitySuite link and description.

spin973:
I added the link and description among the turnkey projects because it seems more appropriate, but if you think it is better to move it, choose where it is better to place it. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ can be found in the [dedicated repository](https://github.com/openiddict/openidd
 **Developers looking for a simple and turnkey solution are strongly encouraged to evaluate these popular options**:
   - [Volo.OpenIddict.Pro](https://abp.io/modules/Volo.OpenIddict.Pro), which is based on OpenIddict, supports all the common OAuth 2.0/OpenID Connect flows and offers a powerful applications/scopes management GUI.
   - [OrchardCore.OpenId](https://docs.orchardcore.net/en/latest/docs/reference/modules/OpenId/), which is also based on OpenIddict, comes with sensible defaults and offers a built-in management GUI to easily register OpenID client applications.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,7 @@ can be found in the [dedicated repository](https://github.com/openiddict/openidd
 **Developers looking for a simple and turnkey solution are strongly encouraged to evaluate these popular options**:
   - [Volo.OpenIddict.Pro](https://abp.io/modules/Volo.OpenIddict.Pro), which is based on OpenIddict, supports all the common OAuth 2.0/OpenID Connect flows and offers a powerful applications/scopes management GUI.
   - [OrchardCore.OpenId](https://docs.orchardcore.net/en/latest/docs/reference/modules/OpenId/), which is also based on OpenIddict, comes with sensible defaults and offers a built-in management GUI to easily register OpenID client applications.
+  - [IdentitySuite](https://identitysuite.net), OpenIddict-based solution supporting common OAuth 2.0/OpenID Connect flows. Manage users, applications, UI themes, and email templates through a unified Blazor interface.
```
kevinchalet, line 38: Note: this list currently only includes options I've personally tested and that I personally recommend for the general public. IdentitySuite is still at the early stage and I haven't had a chance to give it a try yet, so it's likely a bit early 😃 

Let's add it to this page instead: https://documentation.openiddict.com/integrations/

kevinchalet (commented):
Thanks for your PR!

README.md
spin973: I totally agree with you, I created a PR in the documentation project to insert a link where you suggested. As for the certification I will follow the OpenId Connect certification process, but I was wondering if there is also a way to get a "certification" from you, for now I will send you a license key in case you want to try the application. Thank you.

kevinchalet: Great, thanks! ❤️

kevinchalet:
Replaced by https://github.com/openiddict/openiddict-documentation/pull/167.

PR #2360, opened 2025-07-04
Optimize token revocation process in EntityFramework stores

syntaximus:
I think it's possible to create a scenario where someone intentionally uses the same token twice to trigger the revocation process, leading to a large number of updates  and excessive resource usage.

This pull request optimizes the token revocation process by filtering out already revoked tokens, reducing memory usage and unnecessary updates. The current implementation loads all tokens into memory and manually updates each one, which can lead to excessive data being loaded and a high number of unnecessary update operations. By filtering out already revoked tokens, we reduce memory usage and improve performance.

### DIFF
--- a/src/OpenIddict.EntityFramework/Stores/OpenIddictEntityFrameworkTokenStore.cs
+++ b/src/OpenIddict.EntityFramework/Stores/OpenIddictEntityFrameworkTokenStore.cs
@@ -751,7 +751,7 @@ public class OpenIddictEntityFrameworkTokenStore<
         foreach (var token in await (from token in context.Set<TToken>()
                                                           .Include(token => token.Application)
                                                           .Include(token => token.Authorization)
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR!

That's indeed a good idea. Can you just split the `where` conditions you've added so they appear on their own lines? Once it's done, I'll merge this PR and make sure it's part of the 7.0 release that's about to ship.

Cheers.

syntaximus:
> Thanks for your PR!
> 
> That's indeed a good idea. Can you just split the `where` conditions you've added so they appear on their own lines? Once it's done, I'll merge this PR and make sure it's part of the 7.0 release that's about to ship.
> 
> Cheers.

Thank you for your feedback! I've made the changes as per your suggestions and updated the pull request.

kevinchalet:

PR #2362, opened 2025-07-07
Add NetSuite Provider

antonypetras:
This PR adds support for Oracle NetSuite as a provider.

## Quirks
- `expires_in` returned in non-standard string format
- `id_token` not always returned
- `at_hash` not constructed correctly
- introspection response contains `scope` as an array of strings

After much poking around I decided a new handler for the introspection response was required. Hopefully that is correct :sunglasses:

## Other fun notes for future readers
- `refresh_token` grant only returns an additional refresh token when the client is configured as public in NetSuite and is valid for 3 hours only
- revoke requires that `token` is a refresh token and will not work otherwise

[Docs](https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/chapter_160077062690.html)

openiddict

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Exchange.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Exchange.cs
@@ -360,8 +360,8 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                     context.Response.RefreshToken = null;
                 }
 
-                // Note: Alibaba Cloud, Exact Online, and NetSuite return a non-standard "expires_in"
-                // parameter formatted as a string instead of a numeric type.
+                // Note: Alibaba Cloud, Exact Online, and NetSuite return a non-standard
+                // "expires_in" parameter formatted as a string instead of a numeric type.
                 if (context.Registration.ProviderType is ProviderTypes.AlibabaCloud or ProviderTypes.ExactOnline or ProviderTypes.NetSuite &&
                     long.TryParse((string?) context.Response[Parameters.ExpiresIn],
                         NumberStyles.Integer, CultureInfo.InvariantCulture, out long value))
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Exchange.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Exchange.cs
@@ -360,9 +360,9 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                     context.Response.RefreshToken = null;
                 }
 
-                // Note: Alibaba Cloud and Exact Online returns a non-standard "expires_in"
+                // Note: Alibaba Cloud, Exact Online, and NetSuite return a non-standard "expires_in"
                 // parameter formatted as a string instead of a numeric type.
-                if (context.Registration.ProviderType is ProviderTypes.AlibabaCloud or ProviderTypes.ExactOnline &&
+                if (context.Registration.ProviderType is ProviderTypes.AlibabaCloud or ProviderTypes.ExactOnline or ProviderTypes.NetSuite &&
                     long.TryParse((string?) context.Response[Parameters.ExpiresIn],
                         NumberStyles.Integer, CultureInfo.InvariantCulture, out long value))
                 {
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Introspection.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Introspection.cs
@@ -1,0 +1,70 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/openiddict/openiddict-core for more information concerning
+ * the license and the contributors participating to this project.
+ */
+
+using System.Collections.Immutable;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using static OpenIddict.Client.WebIntegration.OpenIddictClientWebIntegrationConstants;
+
+namespace OpenIddict.Client.WebIntegration;
+
+public static partial class OpenIddictClientWebIntegrationHandlers
+{
+    public static class Introspection
+    {
+        public static ImmutableArray<OpenIddictClientHandlerDescriptor> DefaultHandlers { get; } =
+        [
+            /*
+             * Token response extraction:
+             */
+            MapNonStandardResponseParameters.Descriptor
+        ];
+
+        /// <summary>
+        /// Contains the logic responsible for mapping non-standard response parameters
+        /// to their standard equivalent for the providers that require it.
+        /// </summary>
+        public sealed class MapNonStandardResponseParameters : IOpenIddictClientHandler<ExtractIntrospectionResponseContext>
+        {
+            /// <summary>
+            /// Gets the default descriptor definition assigned to this handler.
+            /// </summary>
+            public static OpenIddictClientHandlerDescriptor Descriptor { get; }
+                = OpenIddictClientHandlerDescriptor.CreateBuilder<ExtractIntrospectionResponseContext>()
+                    .UseSingletonHandler<MapNonStandardResponseParameters>()
+                    .SetOrder(int.MaxValue - 50_000)
+                    .SetType(OpenIddictClientHandlerType.BuiltIn)
+                    .Build();
+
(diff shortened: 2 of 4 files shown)


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Introspection.cs
```diff
@@ -0,0 +1,72 @@
...
+
+                // Note: NetSuite returns the "scope" parameter as a non-standard string array,
+                // so this is converted to a space-separated string.
+                if (context.Registration.ProviderType is ProviderTypes.NetSuite &&
+                    (JsonElement?)context.Response[Parameters.Scope] is { ValueKind: JsonValueKind.Array } scopeArray)
+                {
+                    context.Response.Scope = string.Join(
+                        " ",
+                        scopeArray.EnumerateArray()
+                            .Select(val => val.GetString()?.ToLowerInvariant())
```
kevinchalet: Do they really return scopes not cased exactly as during the authorization dance? 😕

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -826,6 +827,11 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
                 // While PayPal claims the OpenID Connect flavor of the code flow is supported,
                 // their implementation doesn't return an id_token from the token endpoint.
                 ProviderTypes.PayPal => (false, false, false),
+                
+                // NetSuite does not return an id_token when using the refresh_token grant type.
+                // Additionally, the at_hash inside their id_token is not a valid hash of the
```
kevinchalet, line 832: lol, I've seen wild things... but that's a first 😄

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Exchange.cs
```diff
@@ -360,9 +360,9 @@ public ValueTask HandleAsync(ExtractTokenResponseContext context)
                     context.Response.RefreshToken = null;
                 }
 
-                // Note: Alibaba Cloud and Exact Online returns a non-standard "expires_in"
+                // Note: Alibaba Cloud, Exact Online, and NetSuite return a non-standard "expires_in"
                 // parameter formatted as a string instead of a numeric type.
```
kevinchalet:
```suggestion
                // Note: Alibaba Cloud, Exact Online, and NetSuite return a non-standard
                // "expires_in" parameter formatted as a string instead of a numeric type.
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1584,7 +1584,25 @@
...
+  -->
+
+  <Provider Name="NetSuite" Id="df795175-5032-419c-baa9-47c5a325e76e"
+            Documentation="https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/chapter_160077062690.html">
+    <Environment Issuer="https://system.netsuite.com"
+                 ConfigurationEndpoint="https://{settings.AccountId}.suitetalk.api.netsuite.com/.well-known/openid-configuration" />
 
+    <Setting PropertyName="AccountId" ParameterName="accountId" Type="String" Required="true"
+             Description="Your NetSuite account ID (e.g. 123456, or 123456-sb1 for a sandbox)"
+    />
```
```suggestion
    <Setting PropertyName="AccountId" ParameterName="identifier" Type="String" Required="true"
             Description="Your NetSuite account ID (e.g. 123456, or 123456-sb1 for a sandbox)" />
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1584,7 +1584,25 @@
...
+                                      ▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
+                                      ██ ▀██ ██ ▄▄▄█▄▄ ▄▄██ ▄▄▄ ██ ██ █▄ ▄█▄▄ ▄▄██ ▄▄▄██
+                                      ██ █ █ ██ ▄▄▄███ ████▄▄▄▀▀██ ██ ██ ████ ████ ▄▄▄██
+                                      ██ ██▄ ██ ▀▀▀███ ████ ▀▀▀ ██▄▀▀▄█▀ ▀███ ████ ▀▀▀██
+                                      ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+
+  <Provider Name="NetSuite" Id="df795175-5032-419c-baa9-47c5a325e76e"
+            Documentation="https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/chapter_160077062690.html">
+    <Environment Issuer="https://system.netsuite.com"
```
```suggestion
    <Environment Issuer="https://system.netsuite.com/"
```

FYI, we generally add a trailing slash in "pathless" issuers, as recommended by the URI spec (see https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.3). It's not a problem for token validation, as OpenIddict will allow both `https://system.netsuite.com` and `https://system.netsuite.com/` for the valid issuer.

(well... except for the Contentful provider, but I'll fix that in a separate PR 😄)

kevinchalet (commented):
Thanks for your great PR! ❤️

If we can merge it this morning, it will ship as part of the OpenIddict 7.0 release 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Introspection.cs
antonypetras: Truly... any case accepted in /authorize, always lowercase in token payloads, but all caps in introspect :unamused:. Even when not specifying any particular scope during /authorize, introspection will return ["OPENID"]. I didn't feel good about a surprise transformation but I figured that not finding a match for the openid scope constant would be even more surprising.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
antonypetras: This is just the tip of the NetSuite iceberg. Don't get me started :weary:

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Introspection.cs
kevinchalet: Good Lord...

Just in case they'd also return non-string items in that parameter in the future, I added a `OpenIddictHelpers.ValidateArrayElements(element, JsonValueKind.String)` check. I also updated the code to use a `HashSet<string>` in case they'd also return duplicate values 🤣 

Can you please give it a try before I merge it?

kevinchalet:
> After much poking around I decided a new handler for the introspection response was required. Hopefully that is correct 😎

No, that's perfect! 👍🏻

> `refresh_token` grant only returns an additional refresh token when the client is configured as public in NetSuite and is valid for 3 hours only

Out of curiosity, does that imply that for confidential clients, they don't use refresh tokens at all and issue access tokens that never expire?

antonypetras:
Thanks for making it so easy to contribute with such good docs.

> Out of curiosity, does that imply that for confidential clients, they don't use refresh tokens at all and issue access tokens that never expire?

All access tokens are valid for 60 minutes.
- For confidential clients your original refresh token is valid for 7 days. Once you use the refresh token once, you get another access token that is valid for 60 minutes (no refresh token) and then you're done.
- For public clients your refresh tokens are only valid for 3 hours, but you can cycle through refresh tokens forever

NetSuite is one of those systems that logs you out of the UI constantly and I guess they want to pass that on downstream.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Introspection.cs
antonypetras: Sounds smart. Tested and working!

see 😃 
openiddict2

kevinchalet:
> Thanks for making it so easy to contribute with such good docs.

❤️ 

> Once you use the refresh token once, you get another access token that is valid for 60 minutes (no refresh token) and then you're done.

🔫 ... yeah, that's really stupid 😄

Merged, thanks a lot for your PR! And congrats for your first contribution to OpenIddict! 🎉

PR #2367, opened 2025-07-24
Reduce the maximum length of string primary keys in the EF 6 and EF Core stores

### LINKED ISSUE #2366: Reduce the maximum length of `string` primary keys in the EF 6 and EF Core stores

- [x] I confirm I'm a sponsor or a contributor

### Version

7.0.0

### Describe the bug

OpenIddict 7.0 increased the maximum length of the `OpenIddictTokens.Type` column from 50 to 150, which caused a regression with MySQL/MariaDB, as the length of the index is now `(255 + 50 + 400 + 50) * 4 = 3420` (which exceeds the maximum value supported by MySQL/MariaDB when using `utf8mb4`).

To fix this issue, we'll have to reduce the length of the `Id` columns from 255 to 100 in the next major version (by default, OpenIddict uses `Guids-as-strings` identifiers like ASP.NET Core Identity, so 100 should be more than enough).

### To reproduce

https://matrix.to/#/!DYBDMbtjENXaKIMiCN:gitter.im/$Rglm1ZpKWhfMdemas-x3AlcBgdL4sFIS48GV3gsqQq0?via=gitter.im&via=matrix.org

### Exceptions (if any)

```shell
MySql.Data.MySqlClient.MySqlException (0x80004005): Specified key was too long; max key length is 3072 bytes
```

kevinchalet:
Fixes https://github.com/openiddict/openiddict-core/issues/2366.

### DIFF
--- a/src/OpenIddict.EntityFramework/Configurations/OpenIddictEntityFrameworkApplicationConfiguration.cs
+++ b/src/OpenIddict.EntityFramework/Configurations/OpenIddictEntityFrameworkApplicationConfiguration.cs
@@ -9,6 +9,7 @@ using System.ComponentModel.DataAnnotations.Schema;
 using System.Data.Entity.Infrastructure.Annotations;
 using System.Data.Entity.ModelConfiguration;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq.Expressions;
 using OpenIddict.EntityFramework.Models;
 
 namespace OpenIddict.EntityFramework;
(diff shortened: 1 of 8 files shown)


uss-185:
There's a typo in the title `lengh` (this was also in the title of the reported bug https://github.com/openiddict/openiddict-core/issues/2366)

kevinchalet:
Nice catch @uss-185. Thanks!

PR #2403, opened 2025-12-08
Add osu! to the list of supported providers

gehongyan:
This pull request would like to add osu! to the list of supported providers.


gehongyan:
## Docs

* osu!: <https://github.com/ppy/osu>
* OAuth: <https://osu.ppy.sh/docs/index.html#authentication>
* Get Owner Data: <https://osu.ppy.sh/docs/index.html#get-own-data>

## Hints

- The Client Credentials Grant requires the `scope` parameter with the value `public`.
  
> [!NOTE]
> **Question 1:** Is it proper to force setting it in the `AttachAdditionalTokenRequestParameters`?

- The user info endpoint supports an optional `mode` parameter. osu! has four game modes: osu!standard, osu!taiko, osu!catch, and osu!mania (in the commonly used order). In osu! API, `catch` is called `fruit` because this mode lets players catch fruits, and the game mode is called [Ruleset](https://osu.ppy.sh/docs/index.html#ruleset).

> [!NOTE]
> **Question 2:** Which naming do you prefer for parameter and constant names: GameMode, Mode, or Ruleset? For the osu!catch constant name, do you prefer `Catch` (more familiar) or `Fruits` (consistent with the API)? For ordering, do you prefer the familiar/custom order or switching to alphabetical order?

## Screenshots

- OAuth App Managing

image

- Authentication Page

image

- Denying Result

image

- Allowing Result (very long, with some flattened arrays)

image
image

- Refreshing

image

- Client Credentials Granting Result

image

kevinchalet:
Hey @gehongyan! Thanks for this new PR! 👏🏻

> **Question 1:** Is it proper to force setting it in the `AttachAdditionalTokenRequestParameters`?

That definitely works, tho' we could also add a new dedicated `OverrideScopes` event handler in `OpenIddictClientWebIntegrationHandlers.cs` - just before the `AttachAdditionalTokenRequestParameters` handler, with an order of `AttachTokenRequestParameters.Descriptor.Order - 500` - that mutate `context.Scopes` before it is attached to the token request.

That said, I see in the docs that the `public` scopes seems to be also supported for the code flow. Maybe we should instead just declare it `Default="true" Required="false"` in `OpenIddictClientWebIntegrationProviders.xml` so that it's always added if the user doesn't add any other scope, independently of the flow? (client credentials or code flow).

> **Question 2:** Which naming do you prefer for parameter and constant names: GameMode, Mode, or Ruleset? For the osu!catch constant name, do you prefer `Catch` (more familiar) or `Fruits` (consistent with the API)? For ordering, do you prefer the familiar/custom order or switching to alphabetical order?

Re-using the official names is generally the most reasonable approach, but if find the names currently used in your PR are clearer, I don't mind keeping them as-is 😃

Note: you'll probably want to update `MapCustomWebServicesFederationClaims` to map the user ID/name/email to their WS-Federation equivalent for consistency with the other providers.

Thanks!

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1747,6 +1747,35 @@
...
+                                                   ██ ▀▀▀ ██ ▀▀▀ ██▄▀▀▄██▀██
+                                                   ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+
+  <Provider Name="Osu" DisplayName="osu!" Id="924114e6-6334-4124-bbf9-46cd5458da25"
+            Documentation="https://osu.ppy.sh/docs/index.html#authentication">
+    <Environment Issuer="https://osu.ppy.sh/">
+      <Configuration AuthorizationEndpoint="https://osu.ppy.sh/oauth/authorize"
+                     TokenEndpoint="https://osu.ppy.sh/oauth/token"
+                     UserInfoEndpoint="{(string.IsNullOrEmpty(settings.GameMode) ? 'https://osu.ppy.sh/api/v2/me' : $'https://osu.ppy.sh/api/v2/me/{settings.GameMode}')}">
```
kevinchalet: Based on your experience with osu!, do you think it would be worth making the game mode (and thus the userinfo response, by extension) per-authentication demand so you can dynamically control it or is using a global setting that applies to all users enough?

gehongyan:
Hi, thanks for helping.

> Maybe we should instead just declare it Default="true" Required="false" in OpenIddictClientWebIntegrationProviders.xml so that it's always added if the user doesn't add any other scope, independently of the flow?

I’ve tested the behavior when the `public` scope is added by default.

As shown in the screenshot below, the permission `Read public data on your behalf` is requested during the code flow. This allows apps to access endpoints requiring the `public` scope (indicated by the green tag in the documentation, such as [Get Beatmap Pack](https://osu.ppy.sh/docs/index.html#get-beatmap-pack)).

image

However, I couldn't find a clean way to remove this scope if the client application specifically doesn't want to request it. Given this, should we still add it by default, or is there a specific method I missed that allows for removing default scopes?

Also, it seems that the Client Credentials flow does not automatically pick up the `public` scope even if it is configured as a default.

In the meantime, assuming the `public` scope shouldn't be forced via configuration, I’ve gone ahead and implemented the `OverrideScope` handler as you suggested. However, setting the order to `AttachTokenRequestParameters.Descriptor.Order - 500` causes `OverrideScope` to execute before `AttachTokenRequestParameters`, where `context.TokenRequest` is initialized. Before this handler, `context.TokenRequest` is still null. Do you have any further suggestions? Would `AttachTokenRequestParameters.Descriptor.Order + 250` be appropriate here?

> Re-using the official names is generally the most reasonable approach, but if find the names currently used in your PR are clearer, I don't mind keeping them as-is.

I furtherly checked how osu!web and SDKs name it.

- [ppy/osu-web](https://github.com/ppy/osu-web/blob/ecf53d79ba3610410f538e8e3fdbccc7cc6179fd/app/Enums/Ruleset.php#L10-L15): `Ruleset::catch`
- [Liam-DeVoe/ossapi](https://github.com/Liam-DeVoe/ossapi/blob/46a843f51a9161c0cfd14fa037e76229123d3106/ossapi/enums.py#L20-L24): `GameMode.CATCH`
- [NiceAesth/aiosu](https://github.com/NiceAesth/aiosu/blob/b0032c0ddcd3fedc8f6b9ef10b3cb5aac682b096/aiosu/models/gamemode.py#L35-L39): `Gamemode.CTB`
- [Sheppsu/osu.py](https://github.com/Sheppsu/osu.py/blob/e50385494050df7f0193f44ab45f0eaaa0a912aa/osu/enums.py#L521-L542): `GameModeStr.CATCH`
- [MaxOhn/rosu-mods](https://github.com/MaxOhn/rosu-mods/blob/6c5634488a076b35a1a0e8e9de6e7fefc7b05985/src/mode.rs#L3-L33): `GameMode::Catch`
- [L-Mario564/osu.js](https://github.com/L-Mario564/osu.js/blob/a76dc179a486aa210f9cdf94338568d19168fac3/src/types/index.ts#L64): `'fruits'` in `GameMode` string union types
- [TTTaevas/osu-api-v2-js](https://github.com/TTTaevas/osu-api-v2-js/blob/d3ad9e486be54c6d38d5a4408606e3ad073f12f6/lib/index.ts#L23-L29): `Ruleset.fruits`

Given that most SDKs use the names `GameMode` and `Catch`, I decided to stick with the more familiar `Catch`. Since it's not a Flags enum, using the singular form (instead of `GameModes`) seems more reasonable and consistent with common usage.

> you'll probably want to update MapCustomWebServicesFederationClaims to map the user ID/name/email to their WS-Federation equivalent for consistency with the other providers.

The ID is mapped from `id`, and the name is mapped from `username` now. The email is not provided.

image

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
gehongyan: Actually, the API for retrieving the currently authorized user's information returns a significant amount of business data unrelated to the authorization process itself, such as pp (Performance Points) and global rank, which change frequently. If the business logic requires the data specified by gamemode, it can easily be retrieved by sending one or more separate requests as needed.

Based on my analysis of the features supported by osu! applications I've encountered, developers likely don't need to focus too much on the data state at the exact moment of authorization. Instead, they would typically fetch the player's latest data in real-time when subsequently conducting business logic. Caching them may result in outdated data.

That said, I agree that being able to dynamically control what parameters are passed during user info API requests, rather than setting a default value for the entire application, offers much better flexibility and future-proofing.

kevinchalet: > That said, I agree that being able to dynamically control what parameters are passed during user info API requests, rather than setting a default value for the entire application, offers much better flexibility and future-proofing.

Roger that. In this case, let's update the `OverrideUserInfoEndpoint` handler to check if the developer added a special `.game_mode` property in the authentication properties and determine the userinfo endpoint dynamically. If it's not present, you can use `https://osu.ppy.sh/api/v2/me` as the fallback URI.

You can take a look at what the Zoho provider does for some inspiration:

https://github.com/openiddict/openiddict-core/blob/cefc4dcba9ab53b8cf68d65bdd340ec2ed93a459/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs#L1090-L1105

If you do that, you'll be able to remove the userinfo URI from the .xml doc.

kevinchalet:
> However, I couldn't find a clean way to remove this scope if the client application specifically doesn't want to request it. Given this, should we still add it by default, or is there a specific method I missed that allows for removing default scopes?

The idea behind default scopes is that they are not requested if the user explicitly added any other value (they exist because some services require at least one scope to be defined, so default scopes can be considered "fallback values").

> Also, it seems that the Client Credentials flow does not automatically pick up the `public` scope even if it is configured as a default.

Ah yeah, good point. You're right, the scopes configured there are only used for user-interactive flows.

> In the meantime, assuming the `public` scope shouldn't be forced via configuration, I’ve gone ahead and implemented the `OverrideScope` handler as you suggested. However, setting the order to `AttachTokenRequestParameters.Descriptor.Order - 500` causes `OverrideScope` to execute before `AttachTokenRequestParameters`, where `context.TokenRequest` is initialized. Before this handler, `context.TokenRequest` is still null. Do you have any further suggestions? Would `AttachTokenRequestParameters.Descriptor.Order + 250` be appropriate here?

Instead of setting `context.TokenRequest.Scope` directly, consider just adding the value to `context.Scopes`: OpenIddict will later attach that collection to the token request without requiring any additional code.

> Given that most SDKs use the names `GameMode` and `Catch`, I decided to stick with the more familiar `Catch`. Since it's not a Flags enum, using the singular form (instead of `GameModes`) seems more reasonable and consistent with common usage.

👍🏻 

Thanks!

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
gehongyan: I updated the code and tested specifying it via

```csharp
var response = await _service.AuthenticateInteractivelyAsync(new()
{
    CancellationToken = stoppingToken,
    Nonce = result.Nonce,
    Properties = new Dictionary<string, string?>
    {
        [OpenIddictClientWebIntegrationConstants.Osu.Properties.GameMode] = OpenIddictClientWebIntegrationConstants.Osu.GameMode.Standard
    }
});
```

It does work. Thanks! :smile:

gehongyan:
> The idea behind default scopes is that they are not requested if the user explicitly added any other value.

I agree that. But the osu! API documentation states that [identify](https://osu.ppy.sh/docs/index.html#scope-identify) is the default scope for the [Authorization Code Grant](https://osu.ppy.sh/docs/index.html#authorization-code-grant) and always implicitly provided. The [Client Credentials Grant](https://osu.ppy.sh/docs/index.html#client-credentials-grant) does not currently have any default scopes.

Furthermore, the scope parameter is optional because when no scope is specified, osu!web only requests identify permissions (Identify you and read your public profile), and the code flow still functions correctly without any explicit scopes. In this case, do we still need to set a default value for Scope in the XML?

image

> Instead of setting `context.TokenRequest.Scope` directly, consider just adding the value to `context.Scopes`: OpenIddict will later attach that collection to the token request without requiring any additional code.

Thank you for the tip; `AttachTokenRequestParameters.Descriptor.Order - 500` works correctly now.

However, I suddenly discovered another issue. Some scopes seemed to be only specified in the Client Credentials mode, including `delegate`, `forum.write_manage` and `group_permissions` (can be confirmed [here](https://github.com/ppy/osu-web/blob/c2c9404146a934bd985d40a11a5a96fda7261338/app/Models/OAuth/Token.php#L24). How should the developer's user code specify these scopes for Client Credentials authentication if we might add them in the `OverrideScopes` handler? It seems that specifying Scopes via Properties is not quite appropriate.

kevinchalet:
> In this case, do we still need to set a default value for Scope in the XML?

If the `scope` parameter is really optional for the code flow, let's not add that to the XML file. The event handler will take care of the client credentials scenario.

> How should the developer's user code specify these scopes for Client Credentials authentication if we might add them in the `OverrideScopes` handler? It seems that specifying Scopes via Properties is not quite appropriate.

Scopes registered via `OpenIddictClientRegistration.Scopes` (or via `options.UseWebProviders().Add[provider name]().AddScopes(...)` for a web provider) are only use for interactive flows like the code flow or the device authorization flow. For other flows like password or the client credentials grant, developers are expected to attach the scopes to the `Scopes` property:

```csharp
// Ask OpenIddict to authenticate the client application using the client credentials grant.
await _service.AuthenticateWithClientCredentialsAsync(new()
{
    CancellationToken = stoppingToken,
    ProviderName = provider,
    Scopes = ["scope1", "scope2"]
});
```

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -621,6 +622,44 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // osu! requires that the scope must be "public" for client credentials
+            // grant, as other scopes have no meaningful effect.
+            if (context.GrantType is GrantTypes.ClientCredentials &&
+                context.Registration.ProviderType is ProviderTypes.Osu)
+            {
+                context.Scopes.Add("public");
+                // TODO: how does user code add custom scopes here?
```
kevinchalet: As mentioned in my other reply, scopes are added by the developers via `ClientCredentialsAuthenticationRequest.Scopes`. When doing so, these scopes are automatically attached to `ProcessAuthenticationContext.Scopes` by OpenIddict.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -621,6 +622,44 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // osu! requires that the scope must be "public" for client credentials
+            // grant, as other scopes have no meaningful effect.
+            if (context.GrantType is GrantTypes.ClientCredentials &&
+                context.Registration.ProviderType is ProviderTypes.Osu)
+            {
+                context.Scopes.Add("public");
```
kevinchalet, line 634: Let's try something: try requesting one of the `delegate`, `forum.write_manage` and `group_permissions` scopes without adding `public`: if it works, it means `public` is not strictly required but that osu requires at least one scope to be set (what OpenIddict calls "default scope").

If it works without `public` being present, we may want to only add it if the `context.Scopes` collection is empty?

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
gehongyan: Got it. Thanks!

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
gehongyan: I don't have a Chat Bot account, so I cannot request `delegate` or other scopes that require delegation; my personal account receives the error 'Delegation with Client Credentials is only available to chat bots.' when requesting such permissions.

However, the [documentation](https://osu.ppy.sh/docs/index.html#client-credentials-delegation) mentions: 'When using delegation, scopes that support delegation cannot be used together with scopes that do not support delegation. Delegation is only available to [Chat Bot](https://osu.ppy.sh/wiki/en/Bot_account)s.'

This indicates that when a developer requests delegation scopes, appending `public` is incorrect, because the [Scopes](https://osu.ppy.sh/docs/index.html#scopes) section points out that `public` is not marked as `Can Delegate`.

The source code of osu!web also confirms this:

https://github.com/ppy/osu-web/blob/c2c9404146a934bd985d40a11a5a96fda7261338/app/Models/OAuth/Token.php#L27
https://github.com/ppy/osu-web/blob/c2c9404146a934bd985d40a11a5a96fda7261338/app/Models/OAuth/Token.php#L236-L245

In a Client Credentials request, if the scopes contain elements from `SCOPES_REQUIRE_DELEGATION`, all elements must be included in the `SCOPES_REQUIRE_DELEGATION`.

Therefore, I agree with your suggestion to add `public` to the `context.Scopes` only when it is empty.

gehongyan: I need to create an issue on the osu!web later regarding the documentation.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1747,6 +1747,34 @@
...
+        <GrantType Value="client_credentials" />
+        <GrantType Value="refresh_token" />
+      </Configuration>
+
+    </Environment>
+
+    <Constant Class="GameMode" Value="osu"    Name="Standard" />
+    <Constant Class="GameMode" Value="taiko"  Name="Taiko"    />
+    <Constant Class="GameMode" Value="fruits" Name="Catch"   />
+    <Constant Class="GameMode" Value="mania"  Name="Mania"    />
```
kevinchalet: Let's name the class `GameModes` to match the convention used in the OpenIddict constant classes 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1059,6 +1098,13 @@ ProviderTypes.HubSpot when
                         left : new Uri("https://api.hubapi.com/oauth/v1/access-tokens", UriKind.Absolute),
                         right: new Uri(token, UriKind.Relative)),
 
+                // osu! supports specifying a game mode when querying the userinfo endpoint.
+                ProviderTypes.Osu
+                    => !context.Properties.TryGetValue(Osu.Properties.GameMode, out string? gameMode) ||
+                        string.IsNullOrEmpty(gameMode) ?
+                            new Uri("https://osu.ppy.sh/api/v2/me", UriKind.Absolute) :
+                            new Uri($"https://osu.ppy.sh/api/v2/me/{gameMode}", UriKind.Absolute),
```
kevinchalet: Just to be safe, let's use `OpenIddictHelpers.CreateAbsoluteUri()` here 😃

gehongyan: Thanks, I overlooked that. 👍

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
gehongyan: Okay, the naming has been changed, and I also aligned the code.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
gehongyan: https://github.com/ppy/osu-web/issues/12611

kevinchalet:
Hey @gehongyan,

FYI: OpenIddict just moved from PolySharp to Polyfill, which allows us to use a lot more polyfills thanks to .NET 10's extensions-everything model 😃 

To make sure your PR can benefit from that change, I merged the `dev` branch and I updated the new `OverrideScopes` handler to use `ThrowIfNull()` and `GetValueOrDefault()`.

Cheers 😄

gehongyan:
Nice work! 😄

kevinchalet:
@gehongyan do you think we can merge your PR or are there changes you'd still like to make depending on https://github.com/ppy/osu-web/pull/12631's outcome? 😃

gehongyan:
@kevinchalet
The documentation changes align with the analysis we made based on the code previously, so there shouldn't be anything that needs to be changed at the moment. I think this PR can now be completed. If there are any other issues in the future, I will open new tickets. 🚀

kevinchalet:
Merged! Thanks for this great new contribution (and sorry for the late merge 😅)

kevinchalet:
Hey,

Just a quick message to let you know your provider finally shipped as part of today's release: https://github.com/openiddict/openiddict-core/releases/tag/7.3.0. I'm sorry it took much more time than usual but I wanted to make sure the design of the new mTLS feature was perfectly flawless before shipping it 😃

Thanks again for your contribution! ❤️

gehongyan:
Obviously, the stability of the infrastructure is very important, and it’s well worth spending some extra time to improve it! Thank you for your work! :rocket:

PR #2416, opened 2026-01-31
Add support for Client ID Metadata Document (CIMD)

thoraj:
## Add Client ID Metadata Document (CIMD) support

Implements [draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) support for the OpenIddict server. When enabled, clients can use an HTTPS URL as their `client_id`. If the `client_id` is not found in the application store, the server fetches a JSON metadata document from the URL and synthesizes a virtual application from it.

### Key changes

- **New `OpenIddict.Server.SystemNetHttp` project** — HTTP client infrastructure for fetching and validating CIMD metadata documents, a CIMD-aware application manager that synthesizes virtual applications, and a scoped context for sharing state within a request
- **Server pipeline changes** — `ValidateClientId` flags unknown HTTPS URL client IDs for metadata fetch instead of rejecting them; discovery metadata advertises `client_id_metadata_document_supported`; bypasses added for handlers that assume pre-registered applications
- **Virtual applications** are always public, always require PKCE, and map `redirect_uris`, `grant_types`, `response_types`, `client_name`, and `application_type` from the metadata document
- **Sandbox demonstrator** (`OpenIddict.Sandbox.AspNetCore.CimdServer`) for manual testing with authorization code + PKCE and refresh token flows
- **31 tests** — 23 unit tests for the application manager, context, and handler filter; 8 integration tests for URL validation rules and discovery metadata

CIMD is disabled by default and opt-in via `EnableClientIdMetadataDocumentSupport()`.

For detailed design documentation, architecture decisions, and manual verification steps, see [docs/CIMD.md](docs/CIMD.md).


kevinchalet:
Thanks for your PR, @thoraj!

That's a very nice use of Claude, that seems to have done a pretty decent job for an AI 😄

That said, it's likely way too early to start implementing CIMD in OpenIddict: it's still an extremely early draft and I'm sure there will be massive breaking changes in the future iterations (it's pretty much an equivalent to unauthenticated dynamic client registration so it shares many of the security implications).

To make sure you don't waste too much time designing something based on CIMD before we're closer to an adoption, I'm going to close this PR. That said, we'll definitely come back to it at some point 😃 

Cheers.

thoraj:
@kevinchalet Thanks for looking at the draft and your comments. It is definitively not ready for prime time, and I did not expect anyone to spend time on this now. 

I wanted to see if it could be done, and how it could look (hence the draft). And as you noted, I used Claude Code to explore. The first attempt worked, but was quite messy. It took some reading and coaching to improve and create "take2". Since I'm still very fresh wrt to the code, I'm sure there is still many improvements to be had. 

The reason for doing this PoC was to understand if openiddict could be an alternative to Duende which we're using today. And if the OpenIdDict maintainers is open to CIMD, and other contributions, I think it is a very interesting alternative for us. 

Looking forward to seeing how the CIMD spec and OpenIdDict evolves 🚀

PR #2427, opened 2026-02-22
Feat/update google client settings

StellaAlexis:
Small PR to add the prompt value to the Google Web Integration configuration. Docs of the parameter can be found here
https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
@@ -1129,8 +1129,9 @@
 
     <Setting PropertyName="AccessType" ParameterName="type" Type="String" Required="false"
              Description="The value used as the 'access_type' parameter (can be set to 'offline' to retrieve a refresh token)" />
+
     <Setting PropertyName="Prompt" ParameterName="prompt" Type="String" Required="false"
-             Description="The value used as the 'prompt' parameter can be set to none, consent (triggers a consent prompt on every session) or select_account (allows the user to select which account to use on each session)." />
+             Description="The value used as the 'prompt' parameter: can be set to none, consent (triggers a consent prompt on every session) or select_account (allows the user to select which account to use on each session)" />
   </Provider>
 
   <!--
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR!

FYI, adding the setting the .xml isn't enough (it won't do anything, basically). You also need to attach the prompt value to the request. See how it's done for the other providers:

https://github.com/openiddict/openiddict-core/blob/7b9fd7677f7a13c998ee6d8cda9e4140029d3688/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs#L2030-L2038

StellaAlexis:
Ah thanks for pointing that out! I think I added the required extra change now. Also checked if there was anything else that needed updating or any tests around this part but I couldn't find any.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1129,6 +1129,8 @@
 
     <Setting PropertyName="AccessType" ParameterName="type" Type="String" Required="false"
              Description="The value used as the 'access_type' parameter (can be set to 'offline' to retrieve a refresh token)" />
+    <Setting PropertyName="Prompt" ParameterName="prompt" Type="String" Required="false"
+             Description="The value used as the 'prompt' parameter can be set to none, consent (triggers a consent prompt on every session) or select_account (allows the user to select which account to use on each session)." />
```
```suggestion

    <Setting PropertyName="Prompt" ParameterName="prompt" Type="String" Required="false"
             Description="The value used as the 'prompt' parameter: can be set to none, consent (triggers a consent prompt on every session) or select_account (allows the user to select which account to use on each session)" />
```

(FYI, the final dot doesn't need to be added, as it's included in the generator template)

kevinchalet (approved):
Thanks! ❤️

kevinchalet:

kevinchalet:
Hey,

Just a quick message to let you know the new setting shipped as part of today's release: https://github.com/openiddict/openiddict-core/releases/tag/7.3.0.

Cheers.

PR #2442, opened 2026-03-10
Sign In With Apple - Name is not available in the ClaimsIdentity

anarian:
Sign In With Apple provides the [name as JSON object](https://developer.apple.com/documentation/signinwithapple/incorporating-sign-in-with-apple-into-other-platforms#Handle-the-response). They only provide this on the first sign up with to the app, afterwards they stop providing this unless you delete the app from your iOS settings or Apple account settings. 

image

I wasn't able to test this using the sandbox console client, Sign In With Apple requires HTTPS return URLs, I'm not able to specify a localhost URL. I'm also having trouble building the binaries locally to run with my web application. 

image

Similar to my last PR for Clever, I added the remapping here to handle this. I also added `"name"` to the `Scopes` constant since this was hardcoded as a string in `OverrideResponseMode`, and required with `OpenIddictClientWebIntegrationBuilder.Apple.AddScopes` for Apple to provide the user's name.

### DIFF
--- a/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
+++ b/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
@@ -512,6 +512,18 @@ public static partial class OpenIddictClientWebIntegrationHandlers
                     }
                 }
 
+                // Note: Apple returns a non-standard "name" claim formatted as a JSON object.
+                else if (context.Registration.ProviderType is ProviderTypes.Apple)
+                {
+                    var name = context.Response[Claims.Name];
+                    if (name is not null)
+                    {
+                        context.Response[Claims.Name] = $"{name?["firstName"]} {name?["lastName"]}";
+                        context.Response[Claims.FamilyName] = name?["lastName"];
+                        context.Response[Claims.GivenName] = name?["firstName"];
+                    }
+                }
+
                 return ValueTask.CompletedTask;
             }
         }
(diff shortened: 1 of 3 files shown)


kevinchalet:
Thanks for your PR.

Not extracting the `user` node from the authorization response was actually a deliberate design choice when I added the provider back in March 2022: this non-standard node is not authenticated and can be easily manipulated by a malicious user to inject any arbitrary name or email address. If an app relies on the email address to make any security-related decision, there's a huge potential for abuse (typically, an elevation of privilege). See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/713 for additional information.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.Userinfo.cs
```diff
@@ -512,6 +512,18 @@ public ValueTask HandleAsync(ExtractUserInfoResponseContext context)
                     }
                 }
 
+                // Note: Apple returns a non-standard "name" claim formatted as a JSON object.
+                else if (context.Registration.ProviderType is ProviderTypes.Apple)
+                {
+                    var name = context.Response[Claims.Name];
```
kevinchalet, line 518: Hum, I don't think it does what you want: this handler exclusively deals with parameters present in the userinfo response, not the authorization response. Unless Apple has decided to fix their horrible implementation, the `user` node is part of the authorization response, not the userinfo response.

kevinchalet:
> I wasn't able to test this using the sandbox console client, Sign In With Apple requires HTTPS return URLs, I'm not able to specify a localhost URL.

To work around this stupid limitation, you can use ngrok or Microsoft's Dev Tunnels.

> I'm also having trouble building the binaries locally to run with my web application.

Are you seeing any particular error message?

anarian:
Thanks for having a look at this. Apple is still using the non-standard node for the information, so I decided to use two custom event handlers to handle this instead. 

If you're interested in having a look, I'll push the changes to my remote branch and [leave that solution](https://github.com/myblueprint-spaces/openiddict-core/commit/c80bd3b513b5c5a28db43f57ba7888170d706166) there if anyone else encounters this error in the future.

PR #2449, opened 2026-03-27
MapStandardWebServicesFederationClaims clears existing WebServiceFederation claims

ax0l0tl:
If a principal has a claim with a WebServiceFederation ClaimType that is mapped in MapStandardWebServicesFederationClaims client handler and there is no corresponding short standard claim then the WebServiceFederation claim entry is cleared. 

Proposed fix: Call SetClaim with WebServiceFederation ClaimType only if there actually is a short standard claim entry.

### DIFF
--- a/src/OpenIddict.Client/OpenIddictClientHandlers.cs
+++ b/src/OpenIddict.Client/OpenIddictClientHandlers.cs
@@ -4758,15 +4758,17 @@ public static partial class OpenIddictClientHandlers
 
             void MapClaim(string destinationClaimType, string sourceClaimType, string? alternativeSourceClaimType = null)
             {
+                if (context.MergedPrincipal.HasClaim(destinationClaimType))
+                {
+                    return;
+                }
                 var claim = context.MergedPrincipal.GetClaim(sourceClaimType);
-                if (claim != null)
+                if (claim == null && alternativeSourceClaimType != null)
                 {
-                    context.MergedPrincipal.SetClaim(destinationClaimType, claim, issuer);
+                    claim = context.MergedPrincipal.GetClaim(alternativeSourceClaimType);
                 }
-                else if (alternativeSourceClaimType != null)
-                {
-                    MapClaim(destinationClaimType, alternativeSourceClaimType);
-                }
+
+                context.MergedPrincipal.SetClaim(destinationClaimType, claim, issuer);
             }
         }
     }
--- a/src/OpenIddict.Client/OpenIddictClientHandlers.cs
+++ b/src/OpenIddict.Client/OpenIddictClientHandlers.cs
@@ -4741,22 +4741,33 @@ public static partial class OpenIddictClientHandlers
                          context.Registration.ProviderName ??
                          context.Registration.Issuer.AbsoluteUri;
 
-            context.MergedPrincipal
-                .SetClaim(ClaimTypes.Email,          context.MergedPrincipal.GetClaim(Claims.Email),             issuer)
-                .SetClaim(ClaimTypes.Gender,         context.MergedPrincipal.GetClaim(Claims.Gender),            issuer)
-                .SetClaim(ClaimTypes.GivenName,      context.MergedPrincipal.GetClaim(Claims.GivenName),         issuer)
-                .SetClaim(ClaimTypes.Name,           context.MergedPrincipal.GetClaim(Claims.PreferredUsername) ??
-                                                     context.MergedPrincipal.GetClaim(Claims.Name),              issuer)
-                .SetClaim(ClaimTypes.NameIdentifier, context.MergedPrincipal.GetClaim(Claims.Subject),           issuer)
-                .SetClaim(ClaimTypes.OtherPhone,     context.MergedPrincipal.GetClaim(Claims.PhoneNumber),       issuer)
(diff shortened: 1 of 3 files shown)


kevinchalet:
Thanks for your PR.

> If a principal has a claim with a WebServiceFederation ClaimType that is mapped in MapStandardWebServicesFederationClaims client handler and there is no corresponding short standard claim then the WebServiceFederation claim entry is cleared.

Interesting. The merged principal is typically populated from the userinfo response and identity token claims, meaning a WS-Federation claim was returned via one of these. Is it a theoretical issue or do you actually have a provider that uses WS-Federation claims?

ax0l0tl:
Thanks for the quick response! 

For us it was not a theoretical problem, but we might do non standard things :). Our Service acts as an IdentityServer as well as a OIDC client AND it exposes an API, were a frontend is executing request passing its access token in the auth header. The access tokens are received from arbitrary (configurable) OIDC providers (the same we authenticate against with our OpenIddict client implementation). With our OpenIddict client we want to produce a Principal that that looks exactly like the one we have if a frontend api request is executed. To achieve that, we added custom client handlers. Claims for the OpenIddict merged principal are taken from the AccessToken also. We mapped those claims using the JsonWebTokenHandler.DefaultInboundClaimTypeMap to align them with the claims we see in frontend requests. So we had ws federation claims in our merged principal and they 'vanished' magically.

src/OpenIddict.Client/OpenIddictClientHandlers.cs
```diff
@@ -4741,22 +4741,33 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
...
-            context.MergedPrincipal.SetClaim("http://schemas.microsoft.com/accesscontrolservice/2010/07/claims/identityprovider",
-                context.MergedPrincipal.GetClaim(Claims.Private.ProviderName));
+            MapClaim("http://schemas.microsoft.com/accesscontrolservice/2010/07/claims/identityprovider",
+                Claims.Private.ProviderName);
 
             return ValueTask.CompletedTask;
+
+            void MapClaim(string destinationClaimType, string sourceClaimType, string? alternativeSourceClaimType = null)
+            {
+                var claim = context.MergedPrincipal.GetClaim(sourceClaimType);
```
kevinchalet: Let's opt for a simpler approach: if the WS-Federation claim is already present, immediately return (you can use the `HasClaim()` helper to make that prettier).

kevinchalet:
Definitely an interesting scenario 😃

There's another place where we do WS-Federation claims mapping:

https://github.com/openiddict/openiddict-core/blob/c79fd288e867309e02eca4993faab9b507d15ff9/src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs#L1393-L1655

We should use the same logic there too (even if none of the existing web providers currently return WS-Federation-style claims, at least AFAIK).

src/OpenIddict.Client/OpenIddictClientHandlers.cs
```diff
@@ -4741,22 +4741,33 @@ public ValueTask HandleAsync(ProcessAuthenticationContext context)
...
+                if (claim != null)
+                {
+                    context.MergedPrincipal.SetClaim(destinationClaimType, claim, issuer);
+                }
+                else if (alternativeSourceClaimType != null)
+                {
+                    MapClaim(destinationClaimType, alternativeSourceClaimType);
```
kevinchalet: While we're at it, let's try to avoid the recursive call if possible.

src/OpenIddict.Client/OpenIddictClientHandlers.cs
ax0l0tl: Changed it to leave existing claims untouched.

src/OpenIddict.Client/OpenIddictClientHandlers.cs
ax0l0tl: Done.

ax0l0tl:
Added the HasClaim check to the MapCustomWebServicesFederationClaims handler, too. I would have promoted the local function to an internal helper extension, but I was insecure where to put it.

kevinchalet:
> I would have promoted the local function to an internal helper extension, but I was insecure where to put it.

A local method is fine 👍🏻 

Note: I slightly tweaked the code to explicitly set a claim value type when doing the mapping. Currently, all mapped claims are strings, but it's likely we'll have to map other types (integer, for instance) in the future, so the code will already be in place 😄 

Please give it a final try to ensure it does what you want and I'll merge it.

kevinchalet:
Merged. Thanks for your PR, @ax0l0tl! ❤️

ax0l0tl:
Sorry, wasn't at home for some days. Tested it and it works fine!

kevinchalet:
@ax0l0tl thanks for giving it a try! 👍🏻

kevinchalet:
FYI, your change was included in 7.5.0 (available later today). Thanks again for your contribution! 😃

ax0l0tl:
Thanks a lot for the quick reaction and friendly conversation! I's fun to contribute here 👍

kevinchalet:
My pleasure. Thanks for your kind words, @ax0l0tl! 😃

PR #2485, opened 2026-06-22
Add ID Austria to the list of supported providers

kescherCode:
This pull requests adds support for [ID Austria](https://www.id-austria.gv.at/en), the Austrian governmental electronic identity service, as a web service provider. Both government entities and private entities (as long as they exist as a company) can create registrations for it in the "company service portal" (https://usp.gv.at).

Developer documentation, unfortunately, only exists in German.

I've validated that these changes work by locally adding the provider and the required credentials (client ID and secret) to the sandbox projects, and setting the required callback URLs at the service registration.

Some notes on the implementation:

- There are two environments, the one usually used to test implementations is called "Referenz", "REF" for short, which I've taken the liberty to loosely translate to "reference".
- For some reason, the backchannel identity token contains a "sub" node with some value which changes on every login, which would by default be mapped to the `NameIdentifier` claim by `MapStandardWebServicesFederationClaims`, and is described by the documentation as "transient value that should instead be replaced with the bPK value". Therefore, I added some code to replace the `NameIdentifier` claim below the condition that would usually handle a _missing_ `NameIdentifier` claim in `MapCustomWebServicesFederationClaims`. Feel free to adjust this check or let me know what to adjust about it, I am somewhat unhappy about the place I had to put this workaround into for a consistent (per app registration) user identifier.
- No `userinfo` endpoint exists because the implementation has deliberately been made in a way where issued tokens cannot be directly mapped back to persons. This is why the "bPK" value exists. More has been described, also only in German, at https://www.bundeskanzleramt.gv.at/agenda/digitalisierung/stammzahlenregisterbehoerde/bereichsspezifische-personenkennzeichen/beschreibung.html


src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1515,6 +1515,14 @@ ProviderTypes.Contentful or ProviderTypes.Smartsheet
 
                     // HubSpot returns the username as a custom "user" node:
                     ProviderTypes.HubSpot => (string?) context.UserInfoResponse?["user"],
+                    
+                    // ID Austria returns a "given_name" and "family_name" node.
+                    // Both are optional.
+                    ProviderTypes.IdAustria => $"{context.BackchannelIdentityTokenPrincipal?.GetClaim("given_name")}{
```
kevinchalet: For consistency, let's treat this new provider like the existing Basecamp or Content providers, where we only add the claim if both the first name and last name are present.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
```diff
@@ -1679,6 +1687,19 @@ ProviderTypes.Webflow       or ProviderTypes.Weibo      or ProviderTypes.Yandex
...
+            // ID Austria's documentation specifies the "sub" node to be a transient value, and refers
+            // to using their custom "bPK" for recognizing users across logins instead.
+            if (context.Registration.ProviderType == ProviderTypes.IdAustria)
+            {
+                if (context.BackchannelIdentityTokenPrincipal?.GetClaim("urn:pvpgvat:oidc.bpk") is { } bpk)
+                {
+                    context.MergedPrincipal.RemoveClaims(ClaimTypes.NameIdentifier);
+                    context.MergedPrincipal.AddClaim(ClaimTypes.NameIdentifier, bpk, issuer);
+                }
+            }
```
kevinchalet: Since the `sub` claim is useless for this provider, I'd probably change the logic to always use the `SetClaim()` extension: if no `urn:pvpgvat:oidc.bpk` claim is present, the claim will be automatically removed from the merged principal, which is better than flowing a non-stable user identifier down to the user code.

```suggestion
            if (context.Registration.ProviderType is ProviderTypes.IdAustria)
            {
                context.MergedPrincipal.SetClaim(ClaimTypes.NameIdentifier,
                    context.BackchannelIdentityTokenPrincipal?.GetClaim("urn:pvpgvat:oidc.bpk") ??
                    context.FrontchannelIdentityTokenPrincipal?.GetClaim("urn:pvpgvat:oidc.bpk"));
            }
```

Note: ID Austria doesn't seem to support the implicit or hybrid flows at this time, but let's fall back to `context.FrontchannelIdentityTokenPrincipal` in case they'd implement it in the future.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1224,6 +1224,25 @@
       <Scope Name="oauth" Default="true" Required="true" />
     </Environment>
   </Provider>
+  
+  <!--
+                                        ▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
+                                        █▄ ▄██ ▄▄▀███ ▄▄▀█ ██ █ ▄▄█▄ ▄█ ▄▄▀██▄██ ▄▄▀██
```
kevinchalet: It looks weird. Did you use lower case characters? If so, please use upper case-only characters.

I'll update the documentation to be clearer about that.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
```diff
@@ -1224,6 +1224,25 @@
...
+                                        ██ ███ ██ ███ ▀▀ █ ██ █▄▄▀██ ██ ▀▀▄██ ▄█ ▀▀ ██
+                                        █▀ ▀██ ▀▀ ███ ██ ██▄▄▄█▄▄▄██▄██▄█▄▄█▄▄▄█▄██▄██
+                                        ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
+  -->
+  
+  <Provider Name="IdAustria" Id="3d66449c-ce7b-46cf-b2f2-6c1fb81bb5fd" Documentation="https://www.id-austria.gv.at/de/developer">
+    <Environment Name="Production" Issuer="https://idp.id-austria.gv.at/">
+      <!-- Required in practice, documentation strongly suggests it -->
+      <Scope Name="profile" Default="true" Required="true" />
```
kevinchalet: Curious: what happens when the `profile` scope isn't requested?

kevinchalet (commented):
Thanks for your PRs!

(the fact this provider returns a non-stable `sub` identifier is extremely unusual, but I guess there's a first for everything 🤣)

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
kescherCode: I did use lower-case characters, yeah. I guess I shall use uppercase ones!

kevinchalet:
Please ignore the build errors: they are caused by a vulnerable transitive package referenced by the EF Core SQLite package: https://github.com/dotnet/efcore/issues/38257. I'll take care of that in a separate PR.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
kescherCode: All claims that relate to the person, *including the bPK claim*, are omitted. So, you're basically only left with the `sub` claim that changes on every sign-in and any claims that aren't actually relating to the person (e.g. there is an optional claim, which can be enabled in an app registration, which sends a static identifier along with the token, for applications to verify if they are receiving tokens they are meant to receive)

kescherCode: Given that, if an application developer wouldn't want to know the person signing in, they'd not request the claims in the app registration in the first place. And without the bPK claim, which gets omitted entirely without the profile scope, well, the entire token becomes really, really useless for the use-case of ID Austria.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kescherCode: I will do this, but I must also point out that real people will not always have both names to begin with. Yes, these are "edge cases", but still. See also https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/, some of which I've encountered irl before.

kescherCode:
> Thanks for your PRs!
> 
> (the fact this provider returns a non-stable `sub` identifier is extremely unusual, but I guess there's a first for everything 🤣)

I was honestly surprised myself, given they could just set the value in their custom bPK claim into the sub claim instead. But, they don't.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: Yeah, it's a good point. It's really a best effort: even simply concatenating the first name/last name in this specific order is fundamentally wrong as it doesn't account for RTL languages like Arabic.

Maybe we should update the affected providers to exclusively use the first name or fall back to the last name otherwise instead of trying to concatenate them? In any case, there's no way we'll ever be able to produce a perfect value that covers all cases.

kevinchalet:
> I was honestly surprised myself, given they could just set the value in their custom bPK claim into the sub claim instead. But, they don't.

It's not super clear what they are trying to achieve with that weird approach. If they wanted to prevent tracking across multiple clients, they could have implemented pairwise subject identifiers instead, which is a fully standard thing: https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kescherCode: ~~RTL languages *should* actually be covered by this, if Unicode control characters are set appropriately (and if not, then that's the provider's issue).~~

Nevermind, you're right. Displaying this as RTL text on LTR UI and vice versa could prove to be funny in this regard.

kescherCode: But I agree on the rest. However, I also feel it's out of scope for this specific PR :D

kescherCode:
> > I was honestly surprised myself, given they could just set the value in their custom bPK claim into the sub claim instead. But, they don't.
> 
> It's not super clear what they are trying to achieve with that weird approach. If they wanted to prevent tracking across multiple clients, they could have implemented pairwise subject identifiers instead, which is a fully standard thing: https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes.

Well, they are also a SAML2 provider, and they've kind of defined this specific identifier outside of any other standard. I guess that may be one reason they are not specifically doing that. But I agree that they could've picked *some* standard way of achieving this.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationProviders.xml
kescherCode: Actually, that's funny. If I don't add this scope (either in this place in the XML or manually when registering the provider in the sandbox), I get no useful claims, as I mentioned. But, the documentation says

> the scopes defined for OIDC have no direct influence on the claims delivered

That's... not true, though. There are missing claims if I omit `profile`, as I described above. I specifically tested this in the sandbox.

Whatever. The documentation does still also say:

> For compatibility reasons, however, it is recommended to request the scopes “openid” and “profile”.

kevinchalet: Very interesting.

Is there anything concrete you can do with the access token issued by the server? The OpenIddict client can be officially used as a pure "authorization-only" client - i.e for API-access-only scenarios where zero user information is available - but since you mentioned the whole point of the service is to authenticate users, it's probably not worth supporting that by making the `profile` scope optional anyway 😅

kescherCode: You could, in theory, create an API that is only accessible for ID Austria (and optionally, eIDAS) users, without caring about user details.

kevinchalet: Hum, do their access tokens use a standard format (e.g JWT)? Do they allow flowing a specific token audience?

Is this an officially supported scenario? It's extremely rare in general.

kescherCode: Access tokens do not use a standard format, no, and you can't flow for any audience that isn't the client ID.

My point is: In theory, you could use an authorization code flow without this scope to only check if a user has this method of logging in.
I'm not aware of any existing implementation of an ID Austria login that does that, nor would it make much sense; if someone implements ID Austria into their auth system, they want something akin to "logging in with your passport or your ID card" (or "some information that could be gleaned from that, verified to be true by a government IT system").

kevinchalet: Good. Let's keep the `profile` scope required then.

Please just add a comment indicating why we're doing that, just so we don't forget when updating this provider in the future.

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kevinchalet: > However, I also feel it's out of scope for this specific PR :D

Well, since your PR adds a compound name claim, we need to determine what's the right value to use:
  - If we decide to use the same approach as the other provider, we need to make it consistent.
  - If we decide to use a different approach (e.g using the first name and falling back to the last name), we'll need to update the other provider in a follow-up issue/PR.

Any opinion on that? What would be your preferred option?

kevinchalet:
Note: the build errors should go away if you merge `dev` into your branch or rebase it on top of `dev`.

kevinchalet:
Please don't hesitate if you need help to finish that PR 😃

src/OpenIddict.Client.WebIntegration/OpenIddictClientWebIntegrationHandlers.cs
kescherCode: My opinion is: In this PR, do it as was done before (so, require both first and last names to be set, and only then set the name claim).

Changing this behaviour across providers should be in a separate PR.

kescherCode:
I was busy otherwise, but I've finished up this PR now and I think that everything is covered.

kevinchalet:
Merged (with a few minor modifications): https://github.com/openiddict/openiddict-core/commit/948e3fe59c4ed3bcf1201d74b6f9d352f75789f3.

Thanks for that PR and congrats for your first contribution to OpenIddict 👏🏻

kevinchalet:
FYI, the ID Austria provider was backported to 7.6.0: https://www.nuget.org/packages/OpenIddict.Client.WebIntegration/7.6.0.

Thanks again for your contribution!

kescherCode:
Oh, neat! Thanks for the backport!

PR #2486, opened 2026-06-23
EF: Restore authorization.Tokens entity state after failed deletion

tedchirvasiu:
I was reading the source code to familiarize myself with the codebase and learn how it works and I noticed that in the DeleteAsync method of OpenIddictEntityFrameworkCoreApplicationStore the tokens for authorizations are removed as such:

https://github.com/openiddict/openiddict-core/blob/5b83f60d4a6e0ae2aff67dc3f1b4e37c9c78ea90/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreApplicationStore.cs#L216-L227

but in the catch clause where the entity states get reverted back to Unchanged, authorization.Tokens seem to be left out:

https://github.com/openiddict/openiddict-core/blob/5b83f60d4a6e0ae2aff67dc3f1b4e37c9c78ea90/src/OpenIddict.EntityFrameworkCore/Stores/OpenIddictEntityFrameworkCoreApplicationStore.cs#L249-L252

I was wondering, shouldn't there be a loop to restore their states as well, as suggested in this PR?

I know this is a very very minor and obscure issue for an edge-case that probably nobody ever ran into, but it caught my eye.

P.S.: Thank you for this amazing library


kevinchalet:
Thanks for your PR! 👋🏻

> I was wondering, shouldn't there be a loop to restore their states as well, as suggested in this PR?

Likely a corner case as you said, but it doesn't hurt to do that 👍🏻 

> P.S.: Thank you for this amazing library

Thanks for your kind words! ❤️

kevinchalet:
Note: the build errors should go away if you merge `dev` into your branch or rebase it on top of `dev`.

kevinchalet:

tedchirvasiu:
> Merged. Congrats for your first contribution to OpenIddict! 🎉

Thank you!

PR #2494, opened 2026-07-11
Add Vercel to the list of supported providers

kevinchalet:
Note: this PR also removes the `CreateAbsoluteUri()` helper accepting a `right` parameter represented as a `string` to force callers to specify a `Uri` with a `UriKind` they expect (instead of always assuming `UriKind.RelativeOrAbsolute`).


kevinchalet:
Note: this provider was backported to 7.6.0.

